PR #10, opened 2015-09-29
Fix duplication of fields in partition container editor

apmasell:
When you save a partition container, it will duplicate the contents of certain fields due to duplicate entries in the form.


frankie13:
I have tested this and the form seems intact.  I also saved a SequencerPartitionContainer and it has been persisted correctly.

frankie13:
:+1:

PR #11, opened 2015-09-30
Migrate to Flyway

apmasell:
This remove all traces of Liquibase and migrates everything to Flyway.

I think `sqlstore/src/main/resources/schemas/dump-lims-schema.sh` should be deleted at this point, since it can't generate useful diffs. Also, there's `sqlstore/src/main/resources/schemas/conan-schema-20120601.sql` and I have no idea what that is.


froggleston:
Great! The conan schema is for the analysis-server. This should be kept.

apmasell:
I wanted to, but I can't find any documentation on how to do it. It seems to be possible if Flyway is integrated in the application proper so that it runs during start up, but I assume we want to keep it safely chained to Maven.

frankie13:
on the plus side the tests pass and the app looks good.

tonydebat:
Flyway will pick up a _flyway.properties_ file in the same directory as the _pom.xml_ file and the values in the properties file will override the values in the pom.

This means that developers can add their own _miso_lims/sqlstore/flyway.properties_ file to specify their actual datatbase connection without needed to modify the _pom.xml_ file. We can update the _.gitignore_ file to ensure that _miso_lims/sqlstore/flyway.properties_ is never committed.

    flyway.user=miso
    flyway.password=secret
    flyway.url=jdbc:mysql://localhost:3307/devmisodb

This doesn't get away from duplication though. The db credentials need to be provided twice, once in _flyway.properties_ for development and again in the JNDI context for deployment.

PR #12, opened 2015-10-02
Bugfix: MISO-161 Add emPCR

frankie13:
https://tracker.tgac.ac.uk/browse/MISO-161

This applies to /miso/library/ library dilutions panel.

Small UI change, all platforms will show a `create pools` button. Solid and 454 platforms show an `Add emPCR` button.


froggleston:
Looks good to me!

PR #14, opened 2015-10-16
Delete project files functionality and tests MISO-29

frankie13:
As per Jira MISO-29

I can add "Project Files" to a project, but how can I remove these
files from a project? I miss a "delete" button.

So now there's a delete button.
More importantly there are example tests for controller helper services (and controllers albeit commented out).


apmasell:
Looks sensible. There should probably be a TODO to delete those commented out code blocks once we have other examples.

frankie13:
@apmasell  I'll remember ;)

callunity:
Looks good to me.

PR #26, opened 2015-11-17
Add a “low quality” flag

apmasell:
This is described in GLT-277.  The idea is that a library might be lousy and shouldn't be put on the sequencer again. This change adds that flag and then highlights pools containing such libraries.


frankie13:
Hi @apmasell !  This needs a rebase I'm afraid :spaghetti:

PR #27, opened 2015-11-17
Make allow anything naming scheme allow anything

apmasell:
Ironically, the “allow anything” scheme was rather restrictive.


frankie13:
@apmasell oddly, this one does too :spaghetti:

frankie13:
Looks great @apmasell !  Might I request a test for this? :neckbeard:

apmasell:
Sure. Given there aren't boundaries, what did you have in mind? Also, the neckbeard emoji looks like Harry Potter.

frankie13:
hmm well my main concern here is bringing the test coverage up.
@CoverageIgnore annotations on methods that are too trivial / you don't want to test are a quick win.

A test case for a null / empty string in validateField would exercise the code there.
Also something uber complexto test that method.  I realise it can only return true but it may be changed in the future and the test should fail if it becomes restrictive.  I guess this one's somewhat open to interpretation, it'll provide a good model for the other naming scheme tests too.
I think also the generateNameFor method should be asserted in a test, not sure what depends on that downstream.
Thanks
:christmas_tree:

apmasell:
Now with more unit tests!

PR #30, opened 2015-11-25
Uncontroversial things skimmed from the bootstrap branch

apmasell:


frankie13:
this compiles fine but I noticed the edit project buttons :pencil2: on /miso/projects is now missing!

apmasell:
Pencils are fixed.

PR #32, opened 2015-11-25
Harmonise empty string checks

apmasell:
There shall be one way to do it.

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/decorator/submission/era/EraExperimentDecorator.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/decorator/submission/era/EraExperimentDecorator.java
@@ -71,7 +71,7 @@ public class EraExperimentDecorator extends AbstractSubmittableDecorator<Documen
       Element studyRef = submission.createElementNS(null, "STUDY_REF");
       studyRef.setAttribute("refname", e.getStudy().getAlias());
       studyRef.setAttribute("refcenter", submissionProperties.getProperty("submission.centreName"));
-      if (e.getStudy().getAccession() != null && !isStringEmptyOrNull(e.getStudy().getAccession())) {
+      if (!isStringEmptyOrNull(e.getStudy().getAccession())) {
         studyRef.setAttribute("accession", e.getStudy().getAccession());
       }
       experiment.appendChild(studyRef);
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/decorator/submission/era/EraExperimentDecorator.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/decorator/submission/era/EraExperimentDecorator.java
@@ -23,6 +23,8 @@
 
 package uk.ac.bbsrc.tgac.miso.core.data.decorator.submission.era;
 
+import static uk.ac.bbsrc.tgac.miso.core.util.LimsUtils.isStringEmptyOrNull;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
@@ -69,7 +71,7 @@ public class EraExperimentDecorator extends AbstractSubmittableDecorator<Documen
       Element studyRef = submission.createElementNS(null, "STUDY_REF");
       studyRef.setAttribute("refname", e.getStudy().getAlias());
       studyRef.setAttribute("refcenter", submissionProperties.getProperty("submission.centreName"));
-      if (e.getStudy().getAccession() != null && !"".equals(e.getStudy().getAccession())) {
+      if (e.getStudy().getAccession() != null && !isStringEmptyOrNull(e.getStudy().getAccession())) {
         studyRef.setAttribute("accession", e.getStudy().getAccession());
       }
       experiment.appendChild(studyRef);
@@ -98,7 +100,7 @@ public class EraExperimentDecorator extends AbstractSubmittableDecorator<Documen
             member.setAttribute("member_name", dil.getName());
             member.setAttribute("refcenter", submissionProperties.getProperty("submission.centreName"));
             member.setAttribute("refname", relevantLibrary.getSample().getAlias());
-            if (relevantLibrary.getSample().getAccession() != null && !"".equals(relevantLibrary.getSample().getAccession())) {
+            if (!isStringEmptyOrNull(relevantLibrary.getSample().getAccession())) {
               sampleDescriptor.setAttribute("accession", relevantLibrary.getSample().getAccession());
             }
             pool.appendChild(member);
@@ -121,7 +123,7 @@ public class EraExperimentDecorator extends AbstractSubmittableDecorator<Documen
             relevantLibrary = dil.getLibrary();
             sampleDescriptor.setAttribute("refname", relevantLibrary.getSample().getAlias());
             sampleDescriptor.setAttribute("refcenter", submissionProperties.getProperty("submission.centreName"));
-            if (relevantLibrary.getSample().getAccession() != null && !"".equals(relevantLibrary.getSample().getAccession())) {
+            if (!isStringEmptyOrNull(relevantLibrary.getSample().getAccession())) {
               sampleDescriptor.setAttribute("accession", relevantLibrary.getSample().getAccession());
             }
           }
@@ -133,7 +135,7 @@ public class EraExperimentDecorator extends AbstractSubmittableDecorator<Documen
       Element libraryDescriptor = submission.createElementNS(null, "LIBRARY_DESCRIPTOR");
       Element libraryName = submission.createElementNS(null, "LIBRARY_NAME");
       if (relevantLibrary != null) {
-        if (e.getPool().getAlias() != null && !"".equals(e.getPool().getAlias())) {
+        if (!isStringEmptyOrNull(e.getPool().getAlias())) {
           libraryName.setTextContent(e.getPool().getAlias());
         } else {
           libraryName.setTextContent(e.getPool().getName());
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/event/alerter/EmailAlerterService.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/event/alerter/EmailAlerterService.java
@@ -23,8 +23,6 @@
 
 package uk.ac.bbsrc.tgac.miso.core.event.alerter;
 
-import static uk.ac.bbsrc.tgac.miso.core.util.LimsUtils.isStringEmptyOrNull;
-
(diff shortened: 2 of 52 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/event/alerter/EmailAlerterService.java
```diff
@@ -60,7 +62,7 @@ public void raiseAlert(Alert a) throws AlertingException {
       throw new AlertingException("No SMTP host specified in the mail.properties configuration file. Cannot send email.");
     } else {
       String from = mailProps.getProperty("mail.from");
-      if (from == null || "".equals(from)) {
+      if (!isStringEmptyOrNull(from)) {
```
frankie13: This one is different! so this should be checking to see if the string is null (as per the removed line).
This change would set the from address to `miso@your.miso.server` as long as it wasn't null.
:confounded:

apmasell: Ooops. I'll fix this. Also, do we need the empty check at all, or should this just use the default argument to `getProperty`?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/decorator/submission/era/EraExperimentDecorator.java
```diff
@@ -69,7 +71,7 @@ public void buildSubmission() {
       Element studyRef = submission.createElementNS(null, "STUDY_REF");
       studyRef.setAttribute("refname", e.getStudy().getAlias());
       studyRef.setAttribute("refcenter", submissionProperties.getProperty("submission.centreName"));
-      if (e.getStudy().getAccession() != null && !"".equals(e.getStudy().getAccession())) {
+      if (e.getStudy().getAccession() != null && !isStringEmptyOrNull(e.getStudy().getAccession())) {
```
frankie13: this check could be condensed into 

    if (!isStringEmptyOrNull(e.getStudy().getAccession()))

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/service/forms/MisoFormsService.java
```diff
@@ -106,14 +108,11 @@ public void importSampleDeliveryFormSamples(List<Sample> samples, boolean checkT
...
-    Boolean b = true;
+    boolean b = true;
     for (Sample s : samples) {
-      if (s.getDescription() != null && !"".equals(s.getDescription()) && s.getScientificName() != null
-          && !"".equals(s.getScientificName())) {
-        b = b & true;
-      } else {
+      if (isStringEmptyOrNull(s.getDescription()) || isStringEmptyOrNull(s.getScientificName())) {
         log.warn(s.getIdentificationBarcode() + ": Sample not valid!");
-        b = b & false;
```
frankie13: lol!

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/util/SessionConversationIdTag.java
```diff
@@ -66,12 +68,12 @@ protected int writeTagContent(TagWriter tagWriter) throws JspException {
...
 
     // if no value was found then try to pull value as request parameter.
-    if (conversationId == null || conversationId.trim().length() == 0) {
+    if (isStringBlankOrNull(conversationId)) {
       conversationId = pageContext.getRequest().getParameter(attributeName + "_cId");
     }
 
     // if a conversation Id was found then process it.
-    if (conversationId != null && conversationId.trim().length() > 0) {
+    if (!isStringBlankOrNull(conversationId)) {
```
frankie13: could use an `else` here

apmasell: Wow...I was clearly in Tetris mode when I applied the changes.

apmasell: No, wait, that changes the flow of this code...I'm not sure what it is supposed to do.

notification-consumer-services/src/main/java/uk/ac/bbsrc/tgac/miso/notification/consumer/service/mechanism/IlluminaNotificationMessageConsumerMechanism.java
```diff
@@ -167,8 +169,7 @@ public void setAttemptRunPopulation(boolean attemptRunPopulation) {
 
               if (run.has("startDate")) {
                 try {
-                  if (run.get("startDate") != null && !run.getString("startDate").equals("null")
-                      && !"".equals(run.getString("startDate"))) {
+                  if (!"null".equals(run.getString("startDate")) && !isStringEmptyOrNull(run.getString("startDate"))) {
```
frankie13: is it best to check for the actual null first?  That way the shortcircuit `&&` will exit before calling `run.getString("startDate")` and prevent a possible exception.

apmasell: This one of those fun cases where equals is not commutative because the receiver is special. `"x".equals(null)` is false, while `((String)null).equals("x")` is an exception. I'm happy to reverse it for clarity.

notification-server/src/test/java/uk/ac/bbsrc/tgac/miso/notification/service/IlluminaTransformerTest.java
```diff
@@ -228,7 +230,7 @@ public void testDateCompletedViaCycleTimes() throws JSONException {
 
     assertTrue(run.has(IlluminaTransformer.JSON_COMPLETE_DATE));
     String date = run.getString(IlluminaTransformer.JSON_COMPLETE_DATE);
-    assertTrue(date != null && !date.equals(""));
+    assertTrue(!isStringEmptyOrNull(date));
     assertTrue(!date.equals("null"));
```
frankie13: Out of scope I guess for this request but these tests could check for a valid date.

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/LimsBindingInitializer.java
```diff
@@ -559,7 +561,7 @@ private Group resolveGroup(Object element) throws IllegalArgumentException {
    */
   private User resolveUser(Object element) throws IllegalArgumentException {
     Long id = null;
-    if (element instanceof String && !"".equals(element)) id = NumberUtils.parseNumber((String) element, Long.class).longValue();
+    if (element instanceof String && !isStringEmptyOrNull((String) element)) id = NumberUtils.parseNumber((String) element, Long.class).longValue();
```
frankie13: Again, out of scope but since you touched the code :smile_cat: 
Does our style guide allow for single line if statements?  :neckbeard: :neckbeard:

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/FlexReportingControllerHelperService.java
```diff
@@ -126,7 +128,7 @@ public JSONObject searchProjectsByCreationDateandString(HttpSession session, JSO
     try {
 
       Collection<Project> projects = null;
-      if (searchStr != null && !searchStr.equals("")) {
+      if (searchStr != null && !isStringEmptyOrNull(searchStr)) {
```
frankie13: This could be shortened to: 
    if (isStringEmptyOrNull(searchStr))

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/ImportExportControllerHelperService.java
```diff
@@ -169,7 +171,7 @@ public JSONObject confirmSamplesUpload(HttpSession session, JSONObject json) thr
     for (JSONArray jsonArrayElement : (Iterable<JSONArray>) jsonArray) {
 
       Sample s = null;
-      if (jsonArrayElement.get(3) != null && !"".equals(jsonArrayElement.getString(3))) {
+      if (jsonArrayElement.get(3) != null && !isStringEmptyOrNull(jsonArrayElement.getString(3))) {
```
frankie13: yet again, very out of scope.
Magic ints, should be defined as constants... somewhere.

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/ImportExportControllerHelperService.java
```diff
@@ -570,7 +572,7 @@ public JSONObject librarySelectionTypesString(HttpSession session, JSONObject js
 
   public JSONObject changePlatformName(HttpSession session, JSONObject json) {
     try {
-      if (json.has("platform") && !json.get("platform").equals("")) {
+      if (json.has("platform") && !isStringEmptyOrNull((String) json.get("platform"))) {
```
frankie13: I'm beginning to think maybe `isStringEmptyOrNull` could take care of the cast for you.  What do you reckon?

apmasell: I strongly disagree. How do you convert from Object to String? There are two ways I see:
- just do instanceof String, in which case calling it on a StringBuffer or the like will return false
- call toString on it, in which case nearly every object will return true.

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/RunControllerHelperService.java
```diff
@@ -940,14 +942,14 @@ public JSONObject getPoolByBarcode(HttpSession session, JSONObject json) {
     try {
       Pool<? extends Poolable> p = null;
 
-      if (barcode != null && !"".equals(barcode)) {
+      if (barcode != null && !isStringEmptyOrNull(barcode)) {
```
frankie13: This one could be shortened down.

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/ImportExportControllerHelperService.java
frankie13: fair enough, must get out of python mode... :white_check_mark:

PR #36, opened 2015-12-08
Change Flyway trigger handling and fix a Study update bug

apmasell:
This moves the triggers into a separate file that runs after Flyway migration every time. We can also put stored procedures here or other bits of SQL that are more like code than like schema.

Additionally, this fixes a bug in the Study update trigger where it uses `studyType` instead of `typeId`.

Finally, there is an update to the readme about all this nonsense.


readme.md
```diff
@@ -85,8 +79,10 @@ database from your remote machine:
 
 Then populate the database with the two dumps by running the following commands at a shell prompt:
 
-    mysql -u tgaclims -p -D lims < lims-schema.sql
-    mysql -u tgaclims -p -D lims < miso_type_data.sql
+    cd sqlstore && mvn -P external compile flyway:migrate
+
+The database schema many need to be upgraded in future version. This commany
```
frankie13: :abc: command*

PR #40, opened 2016-01-06
Box

apmasell:
Add scannable storage boxes. These are similar to pools but 1) can contain pools 2) can be reorganised (where as pooling DNA is pretty irreversible).


apmasell:
It's worth noting that b08ee8d is independent of the box work, but got mixed in.

PR #41, opened 2016-01-08
Reduce list pool overhead

apmasell:
This remove a large cell from the “List Pools” page and reduces the number of AJAX calls. It also avoid loading unnecessary content from the database.


frankie13:
This has been put on hold for now.
It seems this functionality wasn't working for TGAC anyway, presumably due to the size of their database.
This makes the change somewhat tricky to test against the production database.

PR #43, opened 2016-01-12
Add scientific name field to library report

frankie13:
Adds the scientific name field to library flex report.
Apparently this makes life easier :fireworks:


callunity:
Agreed, this looks good.

PR #44, opened 2016-01-12
More lastModifier additions

callunity:
Fixes brokenness in current develop branch by adding in the (hopefully) final additions required for changelogs.

* updates ControllerHelperServiceTests that require User
* adds missing last modifiers for run and partition container saves


apmasell:
Looks good to me.

PR #45, opened 2016-01-12
Make the UserAuthMisoRequestManager use delegation

apmasell:
Inheritance was an awkward way to go about this. I've switched to a delegation model and discovered lots of methods weren't covered before. I invented policy to cover those cases and am open to changing it as I was mostly guessing.


apmasell:
The tests look good. I'm a little nervous about `expectMessage`; I'd avoid hard-coded text.

frankie13:
@apmasell I do dig what you're saying here, however we do need to test the exact expected exception is being thrown here.  So the message is a good identifier.  
We could make the class under test and the test refer to an 'all the strings file' but MISO is riddled with this sort of thing and moving everything would be very time consuming.
Any suggestions?

apmasell:
I'm okay with merging as-is. I don't really have any ideas...well, good, easy, fast; pick two. Maybe just leave a TODO and we can figure out a better testing strategy later.

PR #46, opened 2016-01-14
Moving Confluence docs to gh pages

morgantaschuk:
The pages can be seen here: http://oicr-gsi.github.io/miso-lims/

GLT_388 https://jira.oicr.on.ca/browse/GLT-388 

Note there's a step in the docs readme in order to push to gh-pages.


morgantaschuk:
I see that I massively over rebased.... trying again.

PR #50, opened 2016-01-19
Fix for NPEs in bulk SAM & LIB creation

callunity:
[GLT-359] 
* identificationBarcode was never added to the JSON objects being returned to the server, so checking for them was throwing NPEs


callunity:
The interface for bulk creation will be entirely remodelled with Handsontable, so this fix gets rid of NPEs until then.

PR #52, opened 2016-01-21
fixed table widths

djcooke:
DataTables with hidden columns were setting a static table width on page load. Got ugly when the window was resized


callunity:
:+1:

PR #53, opened 2016-01-22
added/fixed field length counters

djcooke:
Field length counters were half-implemented on most create/edit pages, working on a few, and not present at all on a few. Added/fixed them on all pages


frankie13:
:spaghetti:

djcooke:
No more spaghetti

callunity:
:+1:

PR #55, opened 2016-01-25
added/fixed field length counters

djcooke:
Field length counters were half-implemented on most create/edit pages, working on a few, and not present at all on a few. Added/fixed them on all pages


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditGroupController.java
```diff
@@ -80,6 +90,11 @@ public void setSecurityManager(SecurityManager securityManager) {
       throw ex;
     }
   }
+  
+  @ModelAttribute("maxLengths")
+  public Map<String, Integer> maxLengths() throws IOException {
+    return DbUtils.getColumnSizes(interfaceTemplate, "_Group");
```
frankie13: These are kinda gross and should probably be defined in an enum or similar:

```java
public enum Tables {
    Box("Box"),
    Group("_Group"),
    Pool("Pool"),
    KitDescriptor("KitDescriptor");

    private final String name;       

    private Tables(String s) {
        name = s;
    }

    public String toString() {
       return this.name;
    }
}
```

frankie13:
:+1:  this all seems good to me.  Would be nice to get the table names somewhere standardised though.

Thanks!

apmasell:
:+1:

djcooke:
@frankie13 I had followed the existing method of doing this, but I agree completely. I will create a separate ticket to remove hard-coded table names (and probably JDBC calls) from the Controller classes

PR #56, opened 2016-01-26
Glt 212 sequencer ref details

djcooke:
Adds to Sequencer Reference
* Service Records (with file attachments)
* Fields:
  * serial number
  * date commissioned
  * date decommissioned
  * upgraded sequencer reference
* Additional data (derived from above)
  * active
  * last serviced date
* Read-only (non-admin) Sequencer pages
  * List Sequencers
  * View Sequencer
  * View Service Record


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/UserAuthMisoRequestManager.java
```diff
@@ -2317,4 +2337,29 @@ public void deleteBox(Box box) throws IOException {
...
+    else {
+      throw new IOException("User " + getCurrentUser().getFullName() + " cannot write to this Service Record");
+    }
+  }
+
+  @Override
+  public SequencerServiceRecord getSequencerServiceRecordById(long id) throws IOException {
+    return backingManager.getSequencerServiceRecordById(id);
+  }
+
```
callunity: This get and the list calls below don't follow the read check standard for the rest of the request manager ( ex. https://github.com/oicr-gsi/miso-lims/blob/GLT-212_sequencer_ref_details/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/UserAuthMisoRequestManager.java#L703 )

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditServiceRecordController.java
```diff
@@ -0,0 +1,156 @@
...
+  private DataObjectFactory dataObjectFactory;
+  
+  @Autowired
+  private JdbcTemplate jdbcTemplate;
+
+  public void setInterfaceTemplate(JdbcTemplate interfaceTemplate) {
+    this.jdbcTemplate = interfaceTemplate;
+  }
+
+  public void setRequestManager(uk.ac.bbsrc.tgac.miso.core.manager.RequestManager requestManager) {
```
callunity: This may be an obvious answer, but why are you fully qualifying the RequestManager?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/UserAuthMisoRequestManager.java
djcooke: readCheck/writeCheck are for objects types that have more specific permissions to them than "all users" or "admin only." Get/list methods for other readable-by-all object types in this class follow the same pattern as this ( https://github.com/oicr-gsi/miso-lims/blob/GLT-212_sequencer_ref_details/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/UserAuthMisoRequestManager.java#L2180 )

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
```diff
@@ -88,7 +95,7 @@ public Boolean getAvailable() {
 
   @Override
   public void checkAvailability(int timeout) throws IOException {
-    this.available = getIpAddress().isReachable(timeout);
+    this.available = getIpAddress() == null ? false : getIpAddress().isReachable(timeout);
```
callunity: Thoughts on this null check vs. isStringEmptyOrNull ?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
```diff
@@ -2565,4 +2593,24 @@ public void deleteBox(Box box) throws IOException {
       throw new IOException("No boxStore available. Check that it has been declared in the Spring config.");
     }
   }
+
+  @Override
+  public long saveSequencerServiceRecord(SequencerServiceRecord record) throws IOException {
+    return sequencerServiceRecordStore.save(record);
```
callunity: MISO convention for RequestManager has an initial check for if xStore is null (add to this and three methods below)

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/ViewSequencerReferenceController.java
```diff
@@ -0,0 +1,113 @@
...
+import com.eaglegenomics.simlims.core.User;
+import com.eaglegenomics.simlims.core.manager.SecurityManager;
+
+/**
+ * Controller for View (read-only) Sequencer Reference and Service Record pages. Redirects to /stats for 
+ * write access if user is admin
+ */
+@Controller
+@RequestMapping("/sequencer")
+public class ViewSequencerReferenceController {
```
callunity: Shouldn't this be called "EditSequencerReferenceController" for consistency?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditServiceRecordController.java
djcooke: I don't see a reason for it. I can chop it down

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
djcooke: Not a String. getIpAddress() returns InetAddress

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
djcooke: Right. Will do

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/ViewSequencerReferenceController.java
djcooke: This Controller serves the all-users/read-only Sequencer and Service Record pages and maps to /miso/sequencer/. The Controller for editing Sequencer References is actually called StatsController, and maps to /miso/stats/. I would prefer they shared one Controller, and will think about this some more

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
callunity: Gotcha

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/UserAuthMisoRequestManager.java
callunity: Sounds good

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/ViewSequencerReferenceController.java
callunity: I would also prefer that, though the jsp auth checks and will be non-trivial. Feature/fix for a later version?

miso-web/src/main/webapp/pages/editSequencerReference.jsp
```diff
@@ -67,23 +86,292 @@
...
+            Options
+            <span style="float:right" class="ui-icon ui-icon-triangle-1-s"></span>
+          </a>
+          <div id="recordmenu" onmouseover="mcancelclosetime()" onmouseout="mclosetime()">
+            <c:choose>
+              <c:when test="${sequencerReference.dateDecommissioned == null}">
+                <a href='<c:url value="/miso/stats/sequencer/servicerecord/new/${sequencerReference.id}"/> '>Add new Service Record</a>
+              </c:when>
+              <c:otherwise>
+                <a onclick="alert('Error: Cannot add service records to a retired sequencer')">Add new Service Record</a>
```
callunity: What do you think about hiding the "Add new Service Record" button altogether if it's been decommissioned?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/ViewSequencerReferenceController.java
djcooke: Yes, would probably be best left for later

miso-web/src/main/webapp/pages/editSequencerReference.jsp
djcooke: I like having an indication of **why** records can't be added, rather than have the user wonder why the button is gone. I suppose a message could be placed somewhere visible without clicking anything, but I'm not sure I like that better. Thoughts?

callunity: Perhaps change the text to "Cannot add Service Records to a retired Sequencer" and disable the button (and add class="disabled") if it's retired?

djcooke: Sounds good

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/ViewSequencerReferenceController.java
djcooke: Created ticket: https://jira.oicr.on.ca/browse/GLT-405

djcooke:
Fixed all discussed above

callunity:
:+1:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
frankie13: :white_check_mark:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
```diff
@@ -113,14 +170,28 @@ public boolean isDeletable() {
...
-    sb.append(" : ");
-    sb.append(getAvailable());
-    return sb.toString();
+    return "AbstractSequencerReference [id=" + id + ", name=" + name
+        + ", platform=" + (platform == null ? null : platform.getPlatformId()) + ", available=" + available
+        + ", ip=" + (ip == null ? null : ip.toString())
+        + ", serialNumber=" + serialNumber
+        + ", dateCommissioned=" + (dateCommissioned == null ? null : dateCommissioned.toString())
+        + ", dateDecommissioned=" + (dateDecommissioned == null ? null : dateDecommissioned.toString())
+        + ", upgradedSequencerReference=" + (upgradedSequencerReference == null ? null : upgradedSequencerReference.getId()) + "]";
```
frankie13: If you'd like to get the String value of something and want it to be null friendly you could always try this:
```java
    String s = String.valueOf(obj);
```
I believe `String.valueOf(obj)` looks like this under the hood:

```java
public static String valueOf(Object obj) {
    return (obj == null) ? "null" : obj.toString();
}
```

Applies file wide :neckbeard: #notadealbreaker

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
```diff
@@ -113,14 +170,28 @@ public boolean isDeletable() {
...
+  }
+  
+  @Override
+  public boolean isActive() {
+    return dateDecommissioned == null;
```
frankie13: implicit reliance on java setting this to null despite it being uninitialized makes me squirm a bit.

From the Java docs:

```
For all reference types, the default value is null.
```

Of course, unless your variable is declared as a local variable.  Then everything breaks.
I guess what I'm saying is I'd sleep much better if you initialized this before attempting a comparison.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencerReference.java
```diff
@@ -49,4 +53,74 @@
...
+  
+  public void setSerialNumber(String serialNumber);
+  
+  public String getSerialNumber();
+  
+  /**
+   * Sets the date when use of this sequencer began
+   * 
+   * @param date
+   */
```
frankie13: I heartily approve of your docs :100:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencerServiceRecord.java
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * @return the name of the person who performed the service
+   */
+  public String getServicedByName();
+  
+  /**
+   * Sets the service reference number, which may indicate an external service id, a phone number, or other source of 
+   * additional information
+   * 
+   * @param phone
```
frankie13: this says `phone` and the signature says `referenceNumber`

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/UserAuthMisoRequestManager.java
```diff
@@ -1582,6 +1584,18 @@ public Submission getSubmissionById(long submissionId) throws IOException {
     }
     return accessibles;
   }
+  
+  @Override
+  public Collection<Run> listRunsBySequencerId(Long sequencerReferenceId) throws IOException {
+    User user = getCurrentUser();
+    Collection<Run> accessibles = new HashSet<>();
```
frankie13: ooh didn't know you could do this :white_check_mark:

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditServiceRecordController.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  @RequestMapping(value = "/{recordId}", method = RequestMethod.GET)
+  public ModelAndView viewServiceRecord(@PathVariable(value = "recordId") Long recordId, ModelMap model) throws IOException {
+    SequencerServiceRecord sr = requestManager.getSequencerServiceRecordById(recordId);
+    if (sr != null) {
+      model.put("serviceRecord", sr);
+      model.put("serviceRecordFiles", populateServiceRecordFiles(sr));
+    } else {
+      throw new IOException("Cannot retrieve the requested Service Record");
```
frankie13: Is this good use of an exception?
Perhaps just returning an error response here would suffice.  After all - who cares that the wrong id was sent by the client?  Probably the client / user.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditServiceRecordController.java
```diff
@@ -0,0 +1,147 @@
...
+  
+  @RequestMapping(value = "/new/{sequencerReferenceId}", method = RequestMethod.GET)
+  public ModelAndView newServiceRecord(@PathVariable(value = "sequencerReferenceId") Long sequencerReferenceId, ModelMap model) throws IOException {
+    SequencerReference sequencer = requestManager.getSequencerReferenceById(sequencerReferenceId);
+    if (sequencer == null) {
+      throw new IOException("Cannot retrieve the requested Sequencer Reference");
+    }
+    SequencerServiceRecord record = dataObjectFactory.getSequencerServiceRecord();
+    record.setSequencerReference(sequencer);
+    model.put("serviceRecord", record);
```
frankie13: Would be good if these were constants or maybe in an enum?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditServiceRecordController.java
```diff
@@ -0,0 +1,147 @@
...
+  
+  @RequestMapping(method = RequestMethod.POST)
+  public String processSubmit(@ModelAttribute("serviceRecord") SequencerServiceRecord record, ModelMap model, SessionStatus session)
+      throws IOException {
+    try {
+      requestManager.saveSequencerServiceRecord(record);
+      session.setComplete();
+      model.clear();
+      return "redirect:/miso/stats/sequencer/servicerecord/" + record.getId();
+    } catch (IOException ex) {
```
frankie13: I guess I am a little uncomfortable with the use of checked exceptions here.
This exception goes down through the request manager and (I think) tells us whether a user is admin OR `SequencerServiceRecord` that is being sent through is `null`.
Since we don't attempt to recover and exceptions get logged anyways we have really done nothing useful with this exception.
Here are some other ideas...
######Options include:
* Not throwing the exceptions in the first place and just doing the logging.
* Throwing *RuntimeException* instead so you don't have to bother catching and you get the logging for free.
* Attempting some sort of recovery if this exception is caught (forward to `404` / `400` page).

Personally I'm all about checking the inputs in the API before they ever make it to request manager.  I think they call it *sanitizing your inputs*.  That way you don't even try to save if things don't look right, you can just redirect to a friendly "Nope 404 / 400" page.  It's a better user experience I think.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/StatsController.java
```diff
@@ -116,9 +118,26 @@ public ModelAndView view(ModelMap model) throws IOException {
...
+      }
+    }
+    otherSequencers.remove(sameRef);
+    
     if (sr != null) {
       model.put("sequencerReference", sr);
-      String ip = sr.getIpAddress().toString();
+      model.put("sequencerRuns", runs);
+      model.put("sequencerServiceRecords", serviceRecords);
+      model.put("otherSequencerReferences", otherSequencers);
```
frankie13: more magic strings :sparkles: defining these elsewhere will encourage reuse.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/UploadController.java
```diff
@@ -326,6 +328,15 @@ public void uploadLibraryList(MultipartHttpServletRequest request) throws IOExce
       log.debug("UPLOAD FAIL:", e);
     }
   }
+  
+  @RequestMapping(value = "/servicerecord", method = RequestMethod.POST)
+  public void uploadServiceRecordDocument(MultipartHttpServletRequest request) throws IOException {
+    String recordId = request.getParameter("serviceRecordId");
```
frankie13: check this exists before usage?

miso-web/src/main/webapp/scripts/sequencer_reference_ajax.js
```diff
@@ -105,5 +153,117 @@ Sequencer.ui = {
...
+    jQuery("#upgradedSequencerReferenceLink").empty();
+    if (jQuery("#upgradedSequencerReference").val() != "" && jQuery("#upgradedSequencerReference").val() != 0) {
+      jQuery("#upgradedSequencerReferenceLink").append("<a href='/miso/sequencer/" + jQuery("#upgradedSequencerReference").val() + "'>View</a>");
+    }
+  },
+  
+  deleteServiceRecord : function(recordId, successfunc) {
+    if (confirm("Are you sure you really want to delete service record "+recordId+"? This operation is permanent!")) {
+      Fluxion.doAjax(
+        'serviceRecordControllerHelperService',
```
frankie13: :crying_cat_face: :neckbeard:

frankie13:
Hi @djcooke - I added a bunch of notes, will look again when you have responded.  Also, :100: for tests!

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
djcooke: Changing where appropriate

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
djcooke: I will initialize dateDecommissioned to null in the field declaration

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencerServiceRecord.java
djcooke: Renamed the field and missed the doc. Fixing. Thanks

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditServiceRecordController.java
djcooke: Currently, most controllers throw a SecurityException (unchecked) if the object isn't found (requestManager call returns null). I was going for consistency, but used IOException (checked) instead as it makes a bit more sense. The Request Manager throws an IOException if there is a permission issue, so a return of null just means the object doesn't exist (not a security issue). The result is the same: error is logged, user sees a message + stack trace.

I agree that it's not the best user experience, and should be changed in all Controller classes at some point. A 404 page (and no logging since it's a user error) would probably be best. Would you prefer that I do this (or something else) here, or create a separate ticket to address this in all controllers including this one at some point later on?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditServiceRecordController.java
djcooke: adding ModelKeys enums for ServiceRecord, SequencerReference, and Stats controllers

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/StatsController.java
djcooke: adding enum

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/UploadController.java
djcooke: Adding check that throws IOException in case of null/missing ID parameter, or non-existant object (Service Record in this case) in this method and other similar methods in UploadController

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditServiceRecordController.java
djcooke: I was following the pattern of other EditXControllers, but as you say, the catch/log/throw is a bit pointless, since it will already be logged. I'll remove the try/catch block for now.

MISO throws IOExceptions for a lot of different reasons. In the future, it'd be nice to throw more varied exception types, preferably annotated with Spring's ResponseStatus, so more friendly and helpful error pages could be easily displayed.

djcooke:
@frankie13: Latest commit contains all of the changes I mentioned above in response to your comments

frankie13:
:spaghetti:

djcooke:
Rebased and removed outdated pencil edit links

frankie13:
Hi @djcooke when I try to access the `List Sequencers` page I get the following error.  I have performed all the migrations from this branch:

```
Error: Error invoking listSequencersDataTable on SequencerReferenceControllerHelperService: org.springframework.jdbc.BadSqlGrammarException: StatementCallback; bad SQL grammar [SELECT sr.referenceId, sr.name, sr.ipAddress, sr.platformId, sr.available, sr.serialNumber, sr.dateCommissioned, sr.dateDecommissioned, sr.upgradedSequencerReferenceId, sv.lastServiced FROM SequencerReference sr LEFT JOIN (SELECT sequencerReferenceId, MAX(serviceDate) AS 'lastServiced' FROM SequencerServiceRecord GROUP BY sequencerReferenceId) sv ON sv.sequencerReferenceId = sr.referenceId]; nested exception is com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: Table 'lims.SequencerServiceRecord' doesn't exist```

djcooke:
@frankie13 Are you sure none of the migrations failed? Works fine for me, and the SequencerServiceRecord is created in [sqlstore/src/main/resources/db/migration/V0013__sequencer_details.sql](https://github.com/TGAC/miso-lims/pull/56/files#diff-451c096096c1398ac0c14b9f89cfbae4)

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
frankie13: :white_check_mark:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencerReference.java
frankie13: :white_check_mark:

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/UploadController.java
frankie13: :white_check_mark: might think about changing these from throwing exceptions to returning Bad Request objects or similar.

miso-web/src/main/webapp/scripts/sequencer_reference_ajax.js
frankie13: :white_check_mark:

frankie13:
:+1:

PR #57, opened 2016-01-27
Glt 401 column sizes via dao

djcooke:
Controllers now use RequestManager to get table column size limits for the purpose of showing field length counters. Previously, controllers made JDBC calls via DbUtils and had to include the database table name.


callunity:
:spaghetti: JSONArray got fixed in another merge, and the pencils/links are Hyperlinkified.

djcooke:
Removed the outdated pencil links fix and ate all of the spaghetti

callunity:
:+1: I wish we had a 'hungry hungry hippo' emoji to signify clearing out spaghetti!

frankie13:
:+1:

PR #58, opened 2016-01-28
Update broken query in MySQL  5.7.5+.  Query tried to return columns …

przemekk13:
…that are not functionality dependent on GROUP BY columns


apmasell:
:+1: Looks good.

callunity:
:+1:

PR #59, opened 2016-01-28
GLT-407 more autogenerate identification barcodes

callunity:
- hide "Regenerate All Barcodes" option if autogenerateIdentificationBarcodes is turned off in miso.properties
- hide option to update an identificationBarcode until the item has been created.


apmasell:
Looks good other than the missing paranoia check in the controller.

callunity:
The CacheHelperService can't access the MisoPropertyExporter to test whether autogenerateIdentificationBarcodes is set, so I can't add a check in that method. However: 
* regenerateAllBarcodes() only sets barcodes when the item's barcode is an empty string, so that won't overwrite any pre-existing barcodes;
* the method generates the barcodes by calling save() on the item, and the save() method does perform the check for autogenereateIdentificationBarcodes in the DAO.

apmasell:
Good reasons. :-) :+1:

callunity:
Force-pushed to evict sneaky :space_invader:

frankie13:
:flag-ca:

frankie13:
:maple_leaf:

PR #62, opened 2016-02-01
Add more UserAuthMisoRequestManager Tests

przemekk13:


callunity:
Can you squash these commits into one since they're all along the same theme (more RequestManager unit tests)?

apmasell:
:+1: Looks good.

callunity:
:+1: conditional on changing the PR name to something like "Add more UserAuthMisoRequestManager Tests"

PR #68, opened 2016-02-09
Glt 234 sample hierarchy

tonydebat:
Hello Chris,
It's back! It's the sample hierarchy. Try it out with LDAP to make sure that works. (I was able to get the server to start with LDAP selected, but I could not do a full test since I don't have password access to an LDAP server at the OICR.)

Tony


frankie13:
:gb:

frankie13:
Thanks @tonydebat ! This is great, the LDAP authentication is working fine.
Also, noticed the introduction of dtos :+1: 

Please can someone from :maple_leaf: review this too before merging?

apmasell:
I reviewed it previously and don't have any commits in there (other than the one that gets reverted). I'll have another look.

apmasell:
The additional changes look sensible and I'm assuming the revert is exactly what it says on the tin. :+1:

PR #73, opened 2016-02-12
GLT-411 Add Plate valid'n (and fix Pool valid'n)

callunity:


apmasell:
:+1:

miso-web/src/main/webapp/styles/style.css
```diff
@@ -1848,90 +1848,6 @@ button.disabled {
   opacity: .4;
 }
 
- .portlet {
```
callunity: This CSS styling was introduced in an earlier commit by me, and I am now deliberately removing it.

djcooke:
:+1: other than the 404 on creating new pool

callunity:
Fixed the 404 on creating a new pool

PR #77, opened 2016-02-17
get identity by external name and add unit tests to HibernateIdentityDao

przemekk13:


tonydebat:
Hi Przemek, your changes look good. I would like to see a single value returned instead of a list but I realize in order to do this you'll need to make some changes to Identity. Please make externalName unique in both flyway and IdentityImpl and this will allow you to return a single value.
Thanks,
Tony

apmasell:
:+1:

tonydebat:
:+1:

PR #78, opened 2016-02-17
added user authorization checks at service level

callunity:
for OICR sample hierarchy services

(Merge 1)

### DIFF
--- a/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/security/DefaultAuthorizationManager.java
+++ b/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/security/DefaultAuthorizationManager.java
@@ -1,0 +1,134 @@
+package uk.ac.bbsrc.tgac.miso.service.security;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.core.Authentication;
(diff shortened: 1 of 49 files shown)


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/security/DefaultAuthorizationManager.java
```diff
@@ -0,0 +1,134 @@
...
+  @Override
+  public User getCurrentUser() throws IOException {
+    Authentication auth = securityContextHolderStrategy.getContext().getAuthentication();
+    if (auth == null) {
+      return null;
+    }
+    User user = securityManager.getUserByLoginName(auth.getName());
+    if (user == null && auth.isAuthenticated()) {
+      user = new UserImpl();
+      user.setAdmin(true);
```
callunity: Should there be an admin check as well before setting this?

callunity:
:100: for authorizationManager!

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/security/DefaultAuthorizationManager.java
djcooke: There's no way to do an admin check at this point. I'm not really sure of the case that this supports. The method was copied from UserAuthMisoRequestManager

callunity: Yeah, setting a null user to admin is bizarre. But if it's the pre-existing behaviour, I guess we go with that.

callunity:
:+1:

apmasell:
:+1: throws IOException

PR #79, opened 2016-02-17
Add subproject to sample group

callunity:
(Merge 2)


callunity:
(I'll reopen the PR once auth checks are merged in so they don't have to be present here)

djcooke:
:+1: after auth checks are merged

PR #87, opened 2016-02-18
updates for REST Sample*Controllers (OICR Sample Hierarchy)

callunity:
* rename TissueSampleController to SampleTissueController
* change endpoints for SampleAdditionalInfo, SampleAnalyte and SampleTissue Controllers to be more RESTful
* write URLs only if referent is not null


callunity:
:+1: (@przemekk13 wrote the code, I created the PR)

tonydebat:
:+1:

PR #89, opened 2016-02-18
Replace "Tissue Condition" with "Tissue Type" on front end

callunity:
We had briefly discussed renaming GSLE "Tissue Type" to "Tissue Condition" (experimental condition of Primary/Reference/Metastatic/Xenograft Tissue) for increased semantic clarity, but decided that would be more of an annoyance than a benefit to the lab.


apmasell:
:+1: TT makes me think of the crying emoji.

djcooke:
:+1:

PR #92, opened 2016-02-19
GLT-434 improve asset handling

callunity:
Maven Minify plugin concatenates and minifies CSS and JS files during the build cycle. 

In header.jsp: concatenated imports are **style.css** and **header_script.js** . The external header is untouched except for the removal of cache-invalidating timestamps.

These can be replaced with the minified versions which are also generated, **style.min.css** and **header_script.min.js** , but the minification makes it difficult to debug and I haven't yet found a good way of adding a "production" flag to the jsp.

There is also a cleanup commit which adds semicolons, "===" instead of "==", and fixes some whitespace issues.

### DIFF
--- a/miso-external-web/src/main/webapp/externalHeader.jsp
+++ b/miso-external-web/src/main/webapp/externalHeader.jsp
@@ -38,10 +38,10 @@
     <%-- timestamp to force browser to reload javascript --%>
     <jsp:useBean id="timestamp" class="java.util.Date" scope="request"/>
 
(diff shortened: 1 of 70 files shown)


apmasell:
:+1:

djcooke:
:+1:

If you include the Maven argument '-Dnosuffix=true' when building, the ".min" is omitted and the minified scripts replace the non-minified ones, and so are automatically used

Could add a custom Maven parameter (production/dev switch) that controls this to make the purpose more obvious

PR #94, opened 2016-02-19
added LibraryAdditionalInfo model, dto, dao, and ws

djcooke:


apmasell:
Should the `SQLLibraryDAO` be deleted?

djcooke:
No, this does not replace SQLLibraryDAO. This is for the new table LibraryAdditionalInfo, which is similar to SampleAdditionalInfo

apmasell:
Okay. Just checking. :+1:

djcooke:
Closed for now. Will save for after we have Hibernatized everything so that Hibernate can be used for Kits as well

PR #95, opened 2016-02-22
GLT-431 Sample Hierarchy - sample ws save implementation

tonydebat:
Added ability to create new Analyte samples. When creating samples a parent must always be provided in the form of an id or an external name. In the case of an external name a matching parent will be used and if a match is not found a new parent will be created.

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSample.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSample.java
@@ -61,6 +61,8 @@ import uk.ac.bbsrc.tgac.miso.core.data.impl.SampleAnalyteImpl;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.SampleAnalyteNode;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.SampleIdentityNode;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.SampleImpl;
+import uk.ac.bbsrc.tgac.miso.core.data.impl.SampleTissueImpl;
+import uk.ac.bbsrc.tgac.miso.core.data.impl.SampleTissueNode;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.UserImpl;
 import uk.ac.bbsrc.tgac.miso.core.exception.MalformedLibraryException;
 import uk.ac.bbsrc.tgac.miso.core.exception.MalformedSampleException;
@@ -129,6 +131,10 @@ public abstract class AbstractSample extends AbstractBoxable implements Sample {
   @OneToOne(targetEntity = SampleAnalyteImpl.class)
   @JoinColumn(name = "sampleAnalyteId")
   private SampleAnalyte sampleAnalyte;
+
+  @OneToOne(targetEntity = SampleTissueImpl.class)
+  @JoinColumn(name = "sampleTissueId")
+  private SampleTissue sampleTissue;
 
   @OneToOne(targetEntity = IdentityImpl.class)
   @JoinColumn(name = "identityId")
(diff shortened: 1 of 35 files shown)


miso-dto/pom.xml
```diff
@@ -22,7 +22,7 @@
     </dependency>
     <dependency>
       <groupId>uk.ac.bbsrc.tgac.miso</groupId>
-      <artifactId>miso-service</artifactId>
+      <artifactId>sqlstore</artifactId>
```
apmasell: Is this really necessary? The logical flow of these as separable modules is really falling apart.

apmasell:
:+1:  Please change the name of this PR to something more descriptive.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSample.java
```diff
@@ -713,20 +748,29 @@ public Sample build() {
...
-          return new SampleAnalyteNode(this);
+          return SampleImpl.sampleAnalyte(this);
+        } else if (sampleTissue != null) {
+          checkNotNull(sampleTissue, "sample tissue must be provided to create a SampleTissue");
+          checkNotNull(sampleTissue.getSample(), "sample must be provided to create a SampleTissue");
+          checkNotNull(sampleTissue.getLab(), "Lab must be provided to create a SampleTissue");
+          checkNotNull(sampleTissue.getLab().getAlias(), "Lab alias must be provided to create a SampleTissue");
+          checkNotNull(sampleTissue.getLab().getInstitute(), "Institute must be provided to create a SampleTissue");
+          checkNotNull(sampleTissue.getLab().getInstitute().getAlias(), "Institute alias must be provided to create a SampleTissue");
+          return new SampleTissueNode(this);
```
callunity: My understanding of Lab and Institute is that Lab alias, Lab Institute, and Institute Alias are all admin-only for create and edit, and the null checks are enforced by the database. If that's indeed the case, then shouldn't the user be providing only a Lab? (This would also mean that the user can't actually fix any errors for null lab alias, lab institute or lab institute alias)

tonydebat: Yes, I've updated this just to check for Lab.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -48,30 +76,82 @@ public Sample get(Long sampleId) throws IOException {
...
+    Sample sample = Dtos.to(sampleDto);
+    sample.setProject(sqlProjectDAO.get(sampleDto.getProjectId()));
+    if (sampleDto.getParentId() != null) {
+      sample.setParent(sampleDao.getSample(sampleDto.getParentId()));
+    }
+    if (sampleDto.getSampleIdentityDto() != null) {
+      sample.setIdentity(identityService.to(sampleDto.getSampleIdentityDto()));
+    }
+    if (sampleDto.getSampleAnalyte() != null) {
+      sample.setSampleAnalyte(sampleAnalyteService.to(sampleDto.getSampleAnalyte()));
```
callunity: Do we have the structure in place to add a condition for if (sampleDto.getSampleTissue() != null) ?

callunity:
:+1:

miso-dto/pom.xml
tonydebat: I've updated this to depend on core instead. The dtos just need access to the model objects.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
tonydebat: This was missing, so I updated the ws to create SampleTissue as well.

PR #97, opened 2016-02-22
Update Lab-Institute connection & add Lab front end

callunity:
- Labs and Institutes are now editable by admins
- all empty collection return an empty list instead of a 404
- fixed some small typos

### DIFF
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleIdentityController.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleIdentityController.java
@@ -50,10 +50,6 @@ public class SampleIdentityController {
         log.error("Looking for sample categories", e);
       }
     }
-    if (results.isEmpty()) {
-      return new ResponseEntity<>(HttpStatus.NOT_FOUND);
-    } else {
-      return new ResponseEntity<>(results, HttpStatus.OK);
-    }
+    return new ResponseEntity<>(results, HttpStatus.OK);
   }
 }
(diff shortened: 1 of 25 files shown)


apmasell:
:+1:

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleIdentityController.java
```diff
@@ -50,10 +50,6 @@
         log.error("Looking for sample categories", e);
       }
     }
-    if (results.isEmpty()) {
-      return new ResponseEntity<>(HttpStatus.NOT_FOUND);
-    } else {
-      return new ResponseEntity<>(results, HttpStatus.OK);
-    }
+    return new ResponseEntity<>(results, HttpStatus.OK);
```
djcooke: No need to use ResponseEntity. Can just make the method return Set\<SampleCategoryDto\> (doesn't really matter though)

djcooke:
On the admin page, when attempting to delete an Institute that has Labs associated with it, the error message is "Sorry, something went wrong. Please try again. If the issue persists, contact your administrator." Could this be made clearer?

Other than that, :+1:

callunity:
I can add in a check so Institutes can't be deleted if Labs reference them. As far as clearer error messages, I've chosen to present that one because the JSON error details are only somewhat useful to the user -- for instance, the response.detail message for that error is `could not execute statement; SQL [n/a]; constraint [null]; nested exception is org.hibernate.exception.ConstraintViolationException: could not execute statement`, and the message is `Internal Server Error`. Is there another error message that could be more useful?

PR #101, opened 2016-02-24
Remove Volume from SampleAdditionalInfo

callunity:
Volume got added to Sample directly because it interacts with Boxes. This commit removes it from SampleAdditionalInfo as well.

This commit changes V0010__sample_hierarchy.sql .


apmasell:
:+1: Turn down the volume.

djcooke:
Needs to be removed from SampleAdditionalInfoDto as well

callunity:
Fixed.

djcooke:
:+1:

PR #103, opened 2016-02-25
Relocate lab from SampleTissue to SampleAdditionalInfo

apmasell:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSample.java
```diff
@@ -764,9 +764,6 @@ public Sample build() {
           checkArgument(sampleAdditionalInfo.getTissueType() != null,
               "SampleAdditionalInfo.tissueType must be provided to create a Sample.");
           return SampleImpl.sampleAnalyte(this);
-        } else if (sampleTissue != null) {
-          checkArgument(sampleTissue.getLab() != null, "SampleTissue.lab must be provided to create a SampleTissue");
```
callunity: Shouldn't the checkArgument line be the only thing removed?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SampleAdditionalInfoImpl.java
```diff
@@ -275,6 +280,16 @@ public void setPrepKit(KitDescriptor prepKit) {
...
+
+  @Override
+  public void setLab(Lab lab) {
+    this.lab = lab;
+  }
+
+  @Override
   public Long getHibernateKitDescriptorId() {
     return kitDescriptorId;
   }
```
callunity: toString method? (not a requirement)

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/SampleTissueDto.java
```diff
@@ -126,9 +117,9 @@ public void setLabUrl(String labUrl) {
   @Override
   public String toString() {
     return "SampleTissueDto [bioBankId=" + bioBankId + ", cellularity=" + cellularity + ", createdById=" + createdById + ", createdByUrl="
-        + createdByUrl + ", creationDate=" + creationDate + ", id=" + id + ", labId=" + labId + ", labUrl=" + labUrl
-        + ", instituteTissueName=" + instituteTissueName + ", lastUpdated=" + lastUpdated + ", updatedById=" + updatedById
-        + ", updatedByUrl=" + updatedByUrl + ", url=" + url + "]";
+        + createdByUrl + ", creationDate=" + creationDate + ", id=" + id + ", labUrl=" + labUrl + ", instituteTissueName="
+        + instituteTissueName + ", lastUpdated=" + lastUpdated + ", updatedById=" + updatedById + ", updatedByUrl=" + updatedByUrl
+        + ", url=" + url + "]";
   }
```
callunity: labUrl needs to be taken out as well, and added in to SampleAdditionalInfoDto

apmasell:
Fixed. The labUrl isn't set anywhere, BTW.

tonydebat:
Lab is marked as NOT NULL which made sense when it was part of SampleTissue, since it would occur only once in the hierarchy. Now that it's part of SampleAdditionalInfo it will be present in every sample and entering the same value every time will be tiresome. Suggest we make Lab not a required field.

apmasell:
Okay. Heather was in favour of making it required.

callunity:
Tony makes a good point. We could set some internal logic that it must be included if it's at the top of the hierarchy, but it doesn't make sense to have it saved all the way down the hierarchy.

apmasell:
Changed!

callunity:
:+1:

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleAdditionalInfoService.java
```diff
@@ -151,6 +156,7 @@ public SampleAdditionalInfo to(SampleAdditionalInfoDto sampleAdditionalInfoDto)
     authorizationManager.throwIfUnauthenticated();
     checkArgument(sampleAdditionalInfoDto.getSampleClassId() != null,
         "A SampleAdditionalInfo.sampleClassId must be provided to construct SampleAdditionalInfo.");
+    checkArgument(sampleAdditionalInfoDto.getLabId() != null, "A SampleAdditionalInfo.labId must be provided to construct SampleTissue.");
```
tonydebat: Lab can be null, so this check should be removed.

tonydebat:
:+1:

PR #104, opened 2016-02-25
Add missing tube number in DTO marshalling

apmasell:


miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -237,6 +238,7 @@ public static SampleAdditionalInfo to(SampleAdditionalInfoDto from) {
     to.setTubeNumber(from.getTubeNumber());
     to.setVolume(from.getVolume());
     to.setConcentration(from.getConcentration());
+    to.setTubeNumber(from.getTubeNumber());
     return to;
```
callunity: It is already set in this method, three lines up.

apmasell: Fixed.

callunity:
:+1:

djcooke:
:+1:

PR #105, opened 2016-02-26
Expand table if anchor link was clicked

callunity:
Clicking on anchor links for the Institute Defaults page now expands the table and jumps to the table.


djcooke:
:+1:

PR #107, opened 2016-02-29
GLT-459 Update Sample WS to create un-parented samples

tonydebat:
Also updated Sample and SampleDto to write volume. Updated SampleRelationshipValidation to permit simple samples as valid.


callunity:
:+1:

PR #108, opened 2016-02-29
Add reference genome to project and subproject

przemekk13:


sqlstore/src/main/resources/db/migration/V0014__reference_genome.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE `ReferenceGenome` (
+  `referenceGenomeId` INT NOT NULL,
+  `alias` VARCHAR(45) NULL,
+  PRIMARY KEY (`referenceGenomeId`));
+
```
tonydebat: Add auto_increment to referenceGenomeId.
Make alias unique.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultReferenceGenomeService.java
```diff
@@ -0,0 +1,44 @@
...
+
+  @Autowired
+  private ReferenceGenomeDao referenceGenomeDao;
+
+  @Autowired
+  private AuthorizationManager authorizationManager;
+
+  @Override
+  public Collection<ReferenceGenome> listAllReferenceGenomeTypes() throws IOException {
+    authorizationManager.throwIfNonAdmin();
```
tonydebat: Use authorizationManager.throwIfUnauthenticated() instead so that all users can view the references. With the current throwIfNonAdmin() it's not possible for a non-admin user to create or edit a Project.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/ReferenceGenomeImpl.java
```diff
@@ -0,0 +1,46 @@
...
+import javax.persistence.Table;
+
+import uk.ac.bbsrc.tgac.miso.core.data.ReferenceGenome;
+
+@Entity
+@Table(name = "ReferenceGenome")
+public class ReferenceGenomeImpl implements ReferenceGenome {
+
+  @Id
+  @Column(unique = true, nullable = false)
```
tonydebat: To support id auto generation in mysql change this line to:
@GeneratedValue(strategy = GenerationType.IDENTITY)

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SubprojectController.java
```diff
@@ -104,6 +109,12 @@ private static SubprojectDto writeUrls(SubprojectDto subprojectDto, UriComponent
     return subprojectDtos;
   }
 
+  @RequestMapping(value = "/referenceGenomeOptions", method = RequestMethod.GET, produces = { "application/json" })
```
tonydebat: To match other resource names, this should be all lower case. To indicate that it's a collection use a plural. "/referencegenomes"

miso-web/src/main/webapp/pages/editProject.jsp
```diff
@@ -133,6 +133,16 @@
     </td>
   </tr>
   <tr>
+    <td>Reference Genome :*</td>
```
tonydebat: Since this isn't a required field, the asterisk isn't needed.

miso-web/src/main/webapp/scripts/instituteDefaults_ajax.js
```diff
@@ -404,18 +417,33 @@ var Subproject = Subproject || {
     return select.join('');
   },
 
+  createReferenceGenomeSelect: function(idValue, referenceGenomeId) {
+    var select = [];
+    select.push('<select id="'+ idValue +'">');
+    select.push('<option value="-1">--Please select</option>');
```
tonydebat: For subproject I wonder if we need an option to use the parent project reference? Rather than forcing the choice of a new reference.

sqlstore/src/main/resources/db/migration/V0014__reference_genome.sql
```diff
@@ -0,0 +1,25 @@
...
+INSERT INTO `ReferenceGenome` (`referenceGenomeId`, `alias`) VALUES ('11', 'Vaccinia JX-594 (OICR)');
+INSERT INTO `ReferenceGenome` (`referenceGenomeId`, `alias`) VALUES ('12', 'P. falciparum');
+INSERT INTO `ReferenceGenome` (`referenceGenomeId`, `alias`) VALUES ('13', 'R. sphaeroides');
+INSERT INTO `ReferenceGenome` (`referenceGenomeId`, `alias`) VALUES ('14', 'S. aureus');
+INSERT INTO `ReferenceGenome` (`referenceGenomeId`, `alias`) VALUES ('15', 'de novo assemby');
+INSERT INTO `ReferenceGenome` (`referenceGenomeId`, `alias`) VALUES ('16', 'See comments');
+
+ALTER TABLE Project ADD COLUMN `referenceGenomeId` BIGINT (20) DEFAULT -1 after alias;
+
+ALTER TABLE Subproject ADD COLUMN `referenceGenomeId` BIGINT (20) DEFAULT -1 after alias;
```
tonydebat: For both the Project and Subproject tables add Foreign Key constrains to the ReferenceGenome table.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Subproject.java
```diff
@@ -42,4 +42,8 @@
 
   void setPriority(Boolean priority);
 
+  void setReferenceGenomeId(Long referenceGenomeId);
+
+  Long getReferenceGenomeId();
```
tonydebat: Here in Subproject the Long object is used but in Project the primitive long is used. Should be consistent unless there's a specific reason.

przemekk13:
@tonydebat I made the changes we discussed in a separate commit so you can see the differences.

tonydebat:
@przemekk13 Thanks for making all the changes, they all look good.

I notice that both Project and Subproject have a reference to the id (Long) of the GenomeReference. My preference would be for both Project and Subproject to have a reference to the actual GenomeReference object (instead of the id). This would provide more of an object focus and less of a db focus. However, since Project is on the "sharp edge" between hibernate and MISO persistence, and needs to work with both schemes, you can leave this as is. We will potentially modify this in the future when we revisit how persistence is handled.

:+1:

PR #110, opened 2016-03-02
Add Handsontable files

callunity:
This adds all the Handsontable scripts.


callunity:
Some of the items, like datetime i18n, might be unnecessary for our purposes...

tonydebat:
This is handsontable installed as suggested with bower?

callunity:
I believe that this is downloaded from GH as a zip file, but @frankie13 would be the person to confirm that (this is a commit from the intial work he did in the `f_hot` branch)

tonydebat:
:+1: The Handsontable GH page doesn't indicate that i18n is optional -- seems fine to leave it in. Will this js be compressed/concatenated when deployed?

callunity:
It can be compressed/concatenated, but I suspect (and am trying to find evidence) that it's not actually served to the client by default.

apmasell:
:+1: We can change concatenation in the future.

djcooke:
:+1:

PR #112, opened 2016-03-07
Added Illumina Whole Transcriptome LibraryType

djcooke:


apmasell:
This has to go in a separate migration or it will break the TGAC production DB and our MISO-demo.

djcooke:
@apmasell: Moved to new migration script

PR #117, opened 2016-03-10
Add SampleGroup to Institute Defaults page

callunity:
GLT-382


callunity:
I'll make changes such that Group IDs are unique within a project, not just within a subproject.

PR #118, opened 2016-03-10
GLT-314 Use Handsontable for bulk saving samples

callunity:
- uses Handsontable library to bulk create plain samples and detailed samples parented to identities


apmasell:
In the `EditSampleController`, why does the JSON object get put into a JS string?

PR #120, opened 2016-03-11
Added Sample STR Status

djcooke:
I've added this to plain sample, but it occurs to me now that TGAC may not be interested in this field. Can move it to detailed sample if that is the case


callunity:
STR is a 'genetic identity test', we use it to confirm that a tumor sample is from the same person as its matched normal sample, and that they're not related to other donors that we've received. I don't believe that TGAC deals with matched pairs so I'm tempted to push it to detailed sample, but it might be relevant if they have some test that they regularly run to check for sample swaps. I'd ask their lab before adding it to plain sample.

djcooke:
Closing until it's decided whether to move to SampleAdditionalInfo

PR #132, opened 2016-03-17
GLT-490 Added tests for SQLStudyDAO

tonydebat:
Updated hamcrest dependency.

### DIFF
--- a/sqlstore/src/test/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLStudyDAOTest.java
+++ b/sqlstore/src/test/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLStudyDAOTest.java
@@ -1,0 +1,260 @@
+package uk.ac.bbsrc.tgac.miso.sqlstore;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.Matchers.hasEntry;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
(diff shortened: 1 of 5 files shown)


sqlstore/src/test/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLStudyDAOTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+  @Test
+  public void testListBySearchNone() {
+    List<Study> runs = dao.listBySearch("pizza");
+    assertEquals(0, runs.size());
+  }
+
+  @Test
+  public void testListBySearchEmpty() {
+    List<Study> studies = dao.listBySearch("");
```
apmasell: This seems to liberal as a test.

apmasell: I guess: if someone searches for empty string, should they get results at all?

apmasell:
:+1:

frankie13:
@tonydebat I don't think your test is getting run as part of the suite.

There is no change to the **TestSuite file and this test file doesn't end with Suite so it won't get run.

sqlstore/src/test/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLStudyDAOTest.java
tonydebat: The search adds a SQL wildcard %. So searching for something that doesn't exist returns nothing and searching for nothing returns everything.

PR #134, opened 2016-03-17
SQLPlatformDAO tests

callunity:
Note: Since Platforms get added as part of the migration scripts (before the test data is migrated), nextAutoIncrementId needs to be the ID after the last of the added Platforms since neither MySQL nor H2 roll IDs back.


tonydebat:
:+1:

frankie13:
:spaghetti:

PR #145, opened 2016-03-18
allow users to delete their own notes

djcooke:
Allows non-admin users to delete their own notes on Sample, Library, Pool, Run, and ProjectOverview. This seems to have been the original intended behaviour, but was thwarted by copy pasta.


apmasell:
:+1:

frankie13:
:+1:

PR #150, opened 2016-03-21
Improve change logs

apmasell:
Add them to kits and avoid blank log entries


callunity:
:+1:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/kit/KitDescriptor.java
```diff
@@ -61,6 +67,10 @@
   private String partNumber = "";
   private Integer stockLevel = 0;
   private String description = "";
+  @Transient
```
frankie13: :100:

frankie13:
:+1:

PR #151, opened 2016-03-21
rounding QC and Library Dilutions to 2 decimals

przemekk13:


spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/LibraryControllerHelperService.java
```diff
@@ -673,6 +674,12 @@ public JSONObject addLibraryQC(HttpSession session, JSONObject json) {
     return JSONUtils.SimpleJSONError("Cannot add LibraryQC");
   }
 
+  private double round(double value, int numberOfDigitsAfterDecimalPoint) {
```
apmasell: Ick. It saddens me that Java doesn't have a rounding method. Since this is going to be converted to a string, it would be reasonable to use `DecimalFormat`, but meh.

apmasell:
:+1:

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/LibraryControllerHelperService.java
frankie13: It might be useful to make this `public` and `static` and move it to `/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/LimsUtils.java`.

I find the whole spring module thing a little odd...

przemekk13: Refactored to LimsUtils

callunity:
:+1:

PR #154, opened 2016-03-22
GLT-314 Use Handsontable for bulk saving samples

callunity:


miso-web/src/main/webapp/pages/editSample.jsp
```diff
@@ -329,6 +337,89 @@
...
+      </c:when>
+      <c:otherwise>
+        <button id="savePlain" onclick="Sample.hot.savePlainData();">Save Bulk</button>
+      </c:otherwise>
+      </c:choose>
+    </div>
+    <div id="hotContainer"></div>
+  
+    <script type="text/javascript">
+      Sample.hot.dropdownRef = JSON.parse('${referenceDataJson}');
```
apmasell: Since this already JavaScript, you can drop the `JSON.parse`.

miso-web/src/main/webapp/pages/editSample.jsp
```diff
@@ -329,6 +337,89 @@
...
+      Sample.hot.selectedProjectId = parseInt('${sample.project.id}') || null;
+      Sample.hot.detailedSample = JSON.parse(document.getElementById('HOTbulkForm').dataset.detailedSample);
+      if (Boolean("${detailedSample}")) {
```
apmasell: Again, you can probably convert this to `<c:if>`.

apmasell:
:+1:

djcooke:
I broke it again :smiling_imp:

miso-web/src/main/webapp/pages/editSample.jsp
callunity: `<c:if>` inside the <script> tag? That'll break it good. But I store the boolean one line up, so I"ll use that instead.

miso-web/src/main/webapp/pages/editSample.jsp
callunity: It errors in various ways if I try to drop the JSON.parse.

callunity:
@tonydebat will soon be merging in the fixes for SecurityProfile (an Identity sample is not assigned a Security Profile, which is causing @djcooke 's NPEs all over the place).

djcooke:
:+1: 

There are some issues with validation message not appearing, or being duplicated, but these can be fixed later on. It would be nice to see field-specific validation message as well.

PR #156, opened 2016-03-23
Add concentration to Poolable Elements tables

callunity:
A Library can have multiple dilutions which differ only in their concentrations. Show concentration of library dilutions when displaying all dilutions in a table.


apmasell:
:+1:

djcooke:
:+1:

PR #157, opened 2016-03-23
Made pool and ldi concentration labels consistent

djcooke:
Concentration was labelled "Results" in several places, and the units were missing in most places. Will now be labelled consistently: Concentration (nM)


callunity:
:+1:

apmasell:
:+1:

PR #161, opened 2016-03-30
GLT-519 Sample hierarchy transactions working again. Sample alias and name saved correctly.

tonydebat:
GLT-519 Update to allow hiberate and JDBC to work together within a single transaction.

GLT-519 Cascade saving child sample information.

GLT-519 Set the securityProfileId explicitly based on the provided SecurityProfile.

GLT-519 Avoid cloning hibernate specific portion of persisted Project.

GLT-519 Use existing transaction if one exists when retrieving sample valid relationships.

GLT-519 Use existing transaction if when is present when saving SecurityProfile.

GLT-519 Move call to query.list early in the transaction to avoid unnecessary auto flush.

GLT-519 Execute JDBC statements in session.doWork calls to avoid flushing and ending transaction.

GLT-519 Catching GenericJDBCException thrown by session.doWork calls.

GLT-519 addSample can now throw SQLException.

GLT-519 Handle giving samples temporary names and then generating actual names later.

GLT-519 Set temporary name for sample created with sample ws.

GLT-519 Retrieve alias list without attempting to connect other objects with JDBC calls. This is important since during sample creation some of these object can still be null.
Moved the hibernate save to the beginning of the method to ensure it runs early to create as much of the object as possible before other methods run. This ensure that the parent sample exists, if provided.

GLT-519 Added tests for Sample name generation.

GLT-519 Removed unnecessary set of security profile. Added test to suite.

### DIFF
--- a/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSampleDao.java
+++ b/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSampleDao.java
@@ -1,26 +1,32 @@
 package uk.ac.bbsrc.tgac.miso.persistence.impl;
 
 import java.io.IOException;
+import java.sql.Connection;
+import java.sql.SQLException;
 import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 
 import javax.persistence.CascadeType;
 
 import org.hibernate.Query;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
+import org.hibernate.jdbc.Work;
 import org.hibernate.type.StringType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 
 import com.eaglegenomics.simlims.core.Note;
 import com.eaglegenomics.simlims.core.SecurityProfile;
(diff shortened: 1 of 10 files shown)


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSampleDao.java
```diff
@@ -71,22 +77,50 @@
...
+        }
+
+      }
+    });
+
+    if (sample.getSecurityProfile() != null) {
+      sessionFactory.getCurrentSession().doWork(new Work() {
+
+        @Override
+        public void execute(Connection connection) throws SQLException {
```
apmasell: Is this async? Is that necessary?

apmasell:
If that is async, that's kind of ick, but otherwise :+1:.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSampleDao.java
tonydebat: The `sessionFactory.getCurrentSession().doWork(new Work() {` sections allow JDBC calls to run without completing the current ongoing transaction.

djcooke:
:+1:

Two-step save process for Samples is better because ID is generated before using it in the name. This results in two changelog entries - one for sample creation and one for the name change. Not a major issue, but it'd be nicer to suppress the second message

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSampleDao.java
djcooke: It's not async

PR #163, opened 2016-03-31
GLT-522 Bulk edit samples

callunity:
- provides a controller method for getting samples from a list of IDs from the database for bulk editing.
- creates a new page for bulk editing.

### DIFF
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
@@ -42,6 +42,7 @@ import org.springframework.stereotype.Controller;
 import org.springframework.ui.ModelMap;
 import org.springframework.web.bind.annotation.ModelAttribute;
 import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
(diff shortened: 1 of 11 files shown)


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
```diff
@@ -444,6 +448,37 @@ public ModelAndView setupForm(@PathVariable Long sampleId, @PathVariable Long pr
...
     return null;
   }
+  
+  @RequestMapping(value = "/bulk/", method = RequestMethod.POST, headers = "Accept=application/json")
+  public String getBulkSamples(@RequestBody JSONObject json) throws IOException {
+    JSONArray ids = JSONArray.fromObject(json.get("ids"));
+    StringBuffer stringBuffer = new StringBuffer();
+    for (int i = 0; i < ids.size(); ++i) {
+      stringBuffer.append(ids.getLong(i));
+      if (i + 1 < ids.size()) {
```
apmasell: This is easier to read as:
```
for (int i = 0; i < ids.size(); ++i) {
       if (i>0) { stringBuffer.append(","); }
      stringBuffer.append(ids.getLong(i));
 }
```

miso-web/src/main/webapp/scripts/sample_ajax.js
```diff
@@ -782,6 +783,27 @@ Sample.ui = {
...
+          } else if (!jQuery(this).parent().parent().hasClass('row_selected')) {
+            jQuery(this).parent().parent().addClass('row_selected');
+          }
+        });
+
+        jQuery("#listingSamplesTable").append("<button style=\"margin-left:0px;\" id=\"editBtn\" class=\"fg-button ui-state-default ui-corner-all\">Bulk Edit</button>");
+
+        jQuery("#editBtn").click(function() {
+          var ids = '';
+          jQuery("input[name=bulkCheckbox]").each( function (index) {
```
apmasell: Since this is already an array, you could do something like:

    jQuery("input[name=bulkCheckbox]").filter(function(x) { return x.checked; }).map(function(x) { return x.val(); }).join();

apmasell:
:+1:

miso-web/src/main/webapp/scripts/sample_ajax.js
callunity: Aha -- I can't do it in jQuery because ``jQuery("input[name=bulkCheckbox]")`` returns an object, but I can achieve the same thing using vanilla JS, which makes me happy.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
callunity: I like this pattern. I'm going to have to remember it.

tonydebat:
The SampleAdditionalInfo strStatus isn't working for detailed sample creation when an Identity needs to be created. This has been addressed in https://jira.oicr.on.ca/browse/GLT-533.

:+1:

PR #164, opened 2016-04-01
Standardise on BOOLEAN in the database

apmasell:


sqlstore/src/main/resources/db/migration/V0020__qcpassed_type.sql
```diff
@@ -0,0 +1,17 @@
...
+ALTER TABLE Library DROP COLUMN qcPassed;
+ALTER TABLE Library CHANGE qcPassedNew qcPassed BOOLEAN NULL;
+
+ALTER TABLE Pool ADD COLUMN qcPassedNew BOOLEAN NULL;
+UPDATE Pool Set qcPassedNew = TRUE WHERE qcPassed = 'true' OR qcPassed = '1';
+UPDATE Pool Set qcPassedNew = FALSE WHERE qcPassed = 'false' OR qcPassed = '0';
+ALTER TABLE Pool DROP COLUMN qcPassed;
+ALTER TABLE Pool CHANGE qcPassedNew qcPassed BOOLEAN NULL;
+
+ALTER TABLE Sample ADD COLUMN qcPassedNew BOOLEAN NULL;
```
callunity: BOOLEAN get translated to a TINYINT column like on ``Sample.emptied``, but we also have BIT columns elsewhere like on ``SampleAdditionalInfo.emptied``.  Not a dealbreaker for this PR, but worth settling on something consistent.

callunity:
:+1:

sqlstore/src/main/resources/db/migration/V0020__qcpassed_type.sql
apmasell: `BOOLEAN` is the right type to use. `TINYINT` is 1 byte, so it can vary be more than 0 or 1. `BIT(1)` does work, but since mySQL treats it as a special type, it likes to dump special characters to output. that breaks compatibility with H2 and is generally a pain on the console.

apmasell:
Changes all the `BOOLEAN`-ish things to `BOOLEAN`. No more inconsistent Booleshit.

djcooke:
:+1:

BIT fields get stored more efficiently, and are actually restricted to 0/1 (BOOLEAN is an alias for TINYINT(1), which oddly will store anything (-128)-127), but consistency is an improvement either way

PR #167, opened 2016-04-05
Fix for Handsontable save not working properly

callunity:
Samples no longer get stuck "saving forever" in Handsontable. Dates are also saved properly.


tonydebat:
:+1:

apmasell:
:+1:

PR #168, opened 2016-04-05
remove redundant DB relationships, edit sample elements via sample webservice

djcooke:
Removed foreign keys from Sample table pointing to SampleAdditionalInfo, Identity, SampleTissue, and SampleAnalyte. These tables now all share sampleId as their primary keys.

Removed the SampleHierarchy table, which mapped parents to children. This is already handled by the parentId field in Sample.

Fixed SampleCategoryController name as well - it was called SampleIdentityController before.


apmasell:
:+1:

callunity:
:+1:

PR #171, opened 2016-04-06
GLT-535 Save detailed samples sequentially

tonydebat:
Added a serial function to execute an array of functions in order. Used this to bulk samples.

### DIFF
--- a/miso-web/src/main/webapp/scripts/sample_hot.js
+++ b/miso-web/src/main/webapp/scripts/sample_hot.js
@@ -793,10 +793,11 @@ Sample.hot = {
     return sample;
   },
   
-  saveOneSample: function (data, index, messages, numberToSave) {
+  saveOneSample: function (data, index, messages, numberToSave, callback) {
     var xhr = new XMLHttpRequest();
     xhr.onreadystatechange = function() {
       if (xhr.readyState === XMLHttpRequest.DONE) {
+    	callback(); // Indicate request has completed.
         xhr.status === 201 ? Sample.hot.successSave(xhr, index, messages, numberToSave) : Sample.hot.failSave(xhr, index, messages, numberToSave);
       }
     };
@@ -817,7 +818,7 @@ Sample.hot = {
     }
     var rowIndex = rowIndex;
     console.log(rowIndex, columnIndex);
-    if (rowIndex !== undefined && columnIndex !== -1) {
+    if (rowIndex !== undefined && columnIndex !== -1 && columnIndex !== undefined) {
       Sample.hot.hotTable.setCellMeta(rowIndex, columnIndex, 'valid', false);
       Sample.hot.hotTable.render();
     }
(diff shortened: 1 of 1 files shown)


miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -914,6 +915,17 @@ Sample.hot = {
         // send it through the parser to get a sampleData array that isn't merely a reference to Sample.hot.hotTable.getSourceData()
         var sampleData = JSON.parse(JSON.parse(JSON.stringify(Sample.hot.hotTable.getSourceData())));
         
+        // Array of save functions, one for each line in the table.
+        var sampleSaveArray = new Array();
+
```
callunity: From what I understand, it's more idiomatic in JavaScript to use ``var sampleSaveArray = [];``

tonydebat: Updated.

miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -793,10 +793,11 @@ Sample.hot = {
     return sample;
   },
   
-  saveOneSample: function (data, index, messages, numberToSave) {
+  saveOneSample: function (data, index, messages, numberToSave, callback) {
     var xhr = new XMLHttpRequest();
     xhr.onreadystatechange = function() {
       if (xhr.readyState === XMLHttpRequest.DONE) {
+    	callback(); // Indicate request has completed.
```
apmasell: Is this guaranteed to execute in the desired order? Should the callback be invoked after the save function?

miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -934,6 +947,22 @@ Sample.hot = {
     });
   },
   
+  /**
+   * Serial execution of an array of functions. 
+   * @param {Function[]} arrayOfFunctions - Each function takes a single callback function parameter.
+   */
+  serial: function(arrayOfFunctions) {
+	  index = 0;
```
apmasell: Missing `var`.

miso-web/src/main/webapp/scripts/sample_hot.js
callunity: The callback starts sending the second sample to the server while the front end processes whether the first sample was successfully saved, so having those execute out of order shouldn't be a problem.

callunity:
:+1:

miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -934,6 +947,22 @@ Sample.hot = {
...
+	  
+	  function serialCallBack() {
+		  if(index < (arrayOfFunctions.length)) {
+		     arrayOfFunctions[index++](serialCallBack);
```
apmasell: I find this mutation of state to be really difficult to reason about. Can this be something like:

```
fuction(aof) {
   var invokeNext = function(index) {
       if (index < aof.length) { aof[index](function() { invokeNext(index + 1);} ); }
   }
   invokeNext(0);
}

miso-web/src/main/webapp/scripts/sample_hot.js
tonydebat: Yes. At this point the REST request is complete and either ended in success or failure. The next save can be triggered before the successSave or failSave callbacks update the page with the results.

apmasell:
:+1:

PR #172, opened 2016-04-06
GLT-274 add select checkboxes to List Libraries page

callunity:
- also adds additional bulk action options to Samples
- several of these are placeholders for now, to be implemented later.


apmasell:
:+1: Wow, that's hilariously (though necessarily) complicated for that feature.

tonydebat:
:+1:

PR #174, opened 2016-04-07
GLT-547 add infrastructure for bulk edit libraries

callunity:


apmasell:
:+1: I was really hoping we didn't need to put in any more of those `IN (?, ...)` builders, but here we are.

callunity:
I too had such high hopes for that... Incentive to hibernatize Libraries sooner?

apmasell:
One day. One day...

djcooke:
:+1:

tonydebat:
:+1:

PR #176, opened 2016-04-11
Add sequencing parameters to runs

apmasell:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
```diff
@@ -2,39 +2,50 @@
...
-  public Date getLastUpdated();
+  public void setName(String string);
 
-  public void setLastUpdated(Date lastUpdated);
+  public void setPlatform(Platform platform);
+
+  public void setUpdatedBy(User updatedBy);
+
+  void setXPath(String xpath);
 }
```
callunity: The re-alphabetizing of getters and setters makes this somewhat confusing, but it looks like everything is in (a new) order.

callunity:
:+1:

tonydebat:
:+1:

apmasell:
There autowiring badness, but it doesn't break, so I'm merging.

PR #177, opened 2016-04-12
GLT-537 && GLT-552 Bulk edit and bulk propagate samples

callunity:
- `/miso/bulk/create/{sampleIds}&scid={sampleClassId}` for propagating samples from other samples
- `/miso/bulk/edit/{sampleIds}` for editing samples in place

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSample.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSample.java
@@ -751,9 +751,10 @@ public abstract class AbstractSample extends AbstractBoxable implements Sample {
           return SampleImpl.sampleAnalyte(this);
         } else if (sampleTissue != null) {
           return SampleImpl.sampleTissue(this);
+        } else {
+          return SampleImpl.sampleTissueProcessing(this);
         }
       }
-      throw new IllegalArgumentException("No sample can be built with the specified parameters.");
     }
   }
 
--- a/miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
+++ b/miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
@@ -224,6 +224,7 @@ public class Dtos {
     }
     if (from.getParent() != null) {
       dto.setParentId(from.getParent().getId());
+      dto.setParentAlias(from.getParent().getAlias());
     }
     dto.setPassageNumber(from.getPassageNumber());
     dto.setTimesReceived(from.getTimesReceived());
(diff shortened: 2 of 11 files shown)


apmasell:
:+1:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSample.java
```diff
@@ -751,9 +751,10 @@ public Sample build() {
           return SampleImpl.sampleAnalyte(this);
         } else if (sampleTissue != null) {
           return SampleImpl.sampleTissue(this);
+        } else {
+          return SampleImpl.sampleTissueProcessing(this);
         }
       }
-      throw new IllegalArgumentException("No sample can be built with the specified parameters.");
```
tonydebat: Can't get rid of this as there are still ways to attempt to create a sample that will fail. Not everything that falls through will be a TissueProcessing sample.

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -283,6 +284,11 @@ public static SampleAdditionalInfo to(SampleAdditionalInfoDto from) {
       sampleClass.setSampleClassId(from.getSampleClassId());
       to.setSampleClass(sampleClass);
     }
+    if (from.getParentId() != null) {
```
tonydebat: This doesn't seem right. Since we've got the parent id we can lookup the actual parent. (Not here in this function, but later when it's required.)

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSample.java
callunity: I did it this way because our TissueProcessing samples contain zero extra information (nor db tables) and we don't have a SampleTissueProcessing interface/implementation. Is creating those going to be the best solution, in your opinion?

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
callunity: Should we get rid of this line then? (It's certainly possible that I was overgeneralizing here)

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSample.java
tonydebat: SampleTissueProcessing will be a detailed sample that contains only a SampleAdditionalInfo. So this works, though I would prefer to see some positive testable reason that a sample is a SampleTissueProcessing rather than relying on the fact that it doesn't fit in any other category.

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
tonydebat: Yes, if the code works without this block, then get rid of it.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSample.java
callunity: I could turn the `else` statement into an `else if (sampleAdditionalInfo.getSampleClass().getSampleCategory() == "Tissue Processing")` and keep the exception...

tonydebat:
:+1:

PR #178, opened 2016-04-12
GLT-553 fixes for Sample builder and Dto

callunity:
- fixes issues identified in PR #177


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSample.java
```diff
@@ -751,10 +751,11 @@ public Sample build() {
           return SampleImpl.sampleAnalyte(this);
         } else if (sampleTissue != null) {
           return SampleImpl.sampleTissue(this);
-        } else {
+        } else if (sampleAdditionalInfo.getSampleClass().getSampleCategory() == "Tissue Processing"){
```
apmasell: This needs to be `.equals`

callunity: Riiight. Java.

tonydebat:
:+1:

apmasell:
:+1:

PR #181, opened 2016-04-13
Glt 243 generate sample alias

djcooke:
OICR naming schemes added (not configured by default). When sample is saved with null alias, the alias will be generated


apmasell:
:+1: That's quite the regular expression. O_o

pom.xml
```diff
@@ -387,6 +387,13 @@
     </dependency>
 
     <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-all</artifactId>
+      <version>1.3</version>
+      <scope>test</scope>
+    </dependency>
```
tonydebat: This is a duplicate inclusion. 

The existing hamcrest (line 407) can be updated to be restricted to the 'test' scope.

tonydebat:
After merging to develop the unit test HibernateSampleDaoTest.parentNameModifiedTest fails.

djcooke:
Fixed the duplicate dependency and the test. Test was using the sampleNamingScheme instead of default namingScheme for the name field. Default namingScheme handles the SAMXX, LIBXX... type names

tonydebat:
:+1:

PR #188, opened 2016-04-15
allow blank alias to autogenerate on create single sample page

djcooke:
Allows you to leave alias blank on create sample (single) page to auto-generate IF there is a sample alias generator configured


apmasell:
:+1: This a very nice way to do this. :-)

callunity:
:+1:

PR #189, opened 2016-04-15
Fixes for GLT-559, GLT-561

callunity:
- GLT-559: fixes JSON parse error in bulk saving plain samples
- GLT-561: permits empty aliases if a sample alias generator is avilable


apmasell:
:+1:

djcooke:
:+1:

PR #194, opened 2016-04-20
GLT-585 Add Service Record only available to Admin and Tech in web ui

tonydebat:
The 'add service record' menu is only displayed for users who have permissions to create service records (ADMIN and TECH). This prevents 'permission denied' errors for regular users.


callunity:
:+1:

apmasell:
:+1:

PR #196, opened 2016-04-20
Fix problem with potential base64-encoded strings in searches

apmasell:
Rather than trying to guess (usually wrongly), it searches for the string as
provided and, if that yeilds nothing, searches using a base64-decoded string.
This removes all attempts to do base64 decoding in JavaScript.


callunity:
:+1:

tonydebat:
👍

PR #198, opened 2016-04-20
Avoid all HTML special characters in field names

apmasell:
In the original ticket, it was requested to do the encoding. Since we have so many encoding mechanism (JSON, JSON containing HTML, and HTML written by JSPs), getting encoding and decoding HTML special characters consistently right is going to be very difficult and error-prone. I've opted for the lazy approach of just banning them in Parsley.


callunity:
:+1:

djcooke:
:+1:

PR #200, opened 2016-04-20
GLT-581 after delete, displays table from which item was deleted

callunity:


miso-web/src/main/webapp/scripts/instituteDefaults_ajax.js
```diff
@@ -989,7 +994,12 @@ var Options = Options || {
   },
 
   confirmDelete: function (endpoint) {
+	var pieces = endpoint.split('/');
+	var singular = pieces[pieces.length - 2];
+	localStorage['tablediv'] = singular + 's';
     if (endpoint.indexOf("sampleclass") != -1) {
+      // need a modified tablediv because of pluralization
```
apmasell: Yikes.

apmasell:
:+1:

miso-web/src/main/webapp/scripts/instituteDefaults_ajax.js
callunity: You make a good point. I'll change the table div #ids to be the same as the singular resource which will avoid all pluralization messiness.

tonydebat:
👍

tonydebat:
Nice! The modified update/delete behaviour gets rid of all the problems with refreshing.

PR #201, opened 2016-04-21
GLT-580 unshift institute defaults

callunity:
- keeps tables from shifting up on Institute Defaults page while success checkmark is displayed
- merge this after GLT-581 PR


apmasell:
:+1:

tonydebat:
The update behaviour is correct. 

Noticed that 'delete' will open up another item and scroll to the incorrect location when the item appears after the hash in a url. Like this:
http://localhost:8080/miso/admin/instituteDefaults#samplepurpose

👍

callunity:
Good point, @tonydebat . I've updated PR #200 to fix this.

PR #204, opened 2016-04-21
GLT-550 remove duplicated code which was causing errors

callunity:


djcooke:
:+1:

apmasell:
:+1: Weird, given that's not copy-pasta.

PR #205, opened 2016-04-22
GLT-596 un-round library insert size

callunity:
Library insert size was mistakenly set to be rounded to 2 decimal places. This commit undoes that change.


tonydebat:
👍

apmasell:
:+1:

PR #207, opened 2016-04-22
GLT-595 remove null reference

callunity:
Removes a null element reference so that library bulk action actually work.


apmasell:
:+1:

tonydebat:
👍

PR #208, opened 2016-04-22
GLT-586 & GLT-617

callunity:
- GLT-586 display parent alias in Sample handsontable
- GLT-617 stop JS from freezing page during sample Handsontable creation
- also allows user to edit samples belonging to multiple samples of the same class so long as they are in the same category (ex. Primary & Reference **Tissue**)


apmasell:
:+1:

djcooke:
:+1:

PR #210, opened 2016-04-22
Add library design to edit library page

apmasell:


callunity:
Is the idea that a subset of propagation rules are also library designs? And how do library designs differ from propagation rules?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/LibraryPropagationRule.java
```diff
@@ -106,4 +106,8 @@ public static boolean validate(Library library, Iterable<LibraryPropagationRule>
     return first;
   }
 
+  public boolean isDesign() {
+    return name != null && !name.isEmpty() && platformName != null && paired != null && librarySelectionType != null
+        && libraryStrategyType != null;
+  }
```
djcooke: Or add DB column 'isDesign'?

apmasell: I will add a comment. A library design is one that has no wildcards in it. So, if you allow any selection strategy, they you can't be a design. So long as everything is locked, a rule is a design.

djcooke:
Is there a use for LibraryPropagationRules that are not designs?

apmasell:
I believe there is a use for rules that aren't designs. When we discussed it with the lab, there are certain library protocols that seem to be used for multiple things. It could be that we code them as separate designs, but wildcards seemed sensible when talking. Heather, do you recall this conversation?

PR #212, opened 2016-04-25
GLT-278 add runs to containers page & GLT-590 new run NPE

callunity:
On the Partition Containers page: displays a table of the runs associated with the container
- adds sequencing parameters to the view/edit run page only if the run already has an associated sequencer reference (ie. not on create, as that was causing a NPE -- GLT-590)
- also adds a check for container form existence before attaching a Parsley validator to the form (the form is only available on the create page)


apmasell:
:+1:

tonydebat:
👍

PR #216, opened 2016-04-27
GLT-568 improve save message

callunity:
- fixes "save forever" issue. (once and for all?)
- displays correct numbers of items which have been saved


apmasell:
:+1: Heh. Once and for all.

djcooke:
:+1:

PR #218, opened 2016-04-29
Fix bad verion numbers

apmasell:


callunity:
:+1:

djcooke:
Oh, because they're not modules modules of the main MISO pom, so neither mvn release:update-versions nor mvn versions:set will likely touch them when run in the main directory... :+1:

apmasell:
They are in the external profile, so we just have to fix the release instructions.

PR #221, opened 2016-05-02
Proper success/error message display in all saves

callunity:
Follows up on work in PR #216 and expands the correct save method to all saves/propagates in samples & libraries.


djcooke:
:+1:

apmasell:
:+1: Since these are all the same, you can collapse them down into a single function. It's fine as is though.

callunity:
I'm doing a refactoring, so they'll be collapsed there :)

PR #222, opened 2016-05-02
Refactor shared Handsontable code to own module

callunity:
- also includes some documentation for the JS functions

### DIFF
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/util/form/SelectTag.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/util/form/SelectTag.java
@@ -257,8 +257,9 @@ public class SelectTag extends AbstractHtmlInputElementTag {
     String valueProperty = ObjectUtils.getDisplayString(evaluate("itemValue", getItemValue()));
     String labelProperty = ObjectUtils.getDisplayString(evaluate("itemLabel", getItemLabel()));
 
-    Object comparisonValue = PropertyAccessorFactory.forBeanPropertyAccess(getBindStatus().getActualValue())
-        .getPropertyValue(valueProperty);
+    Object pathValue = getBindStatus().getActualValue();
+    Object comparisonValue = pathValue == null ? new Object()
+        : PropertyAccessorFactory.forBeanPropertyAccess(pathValue).getPropertyValue(valueProperty);
 
     for (Object item : (Iterable<? extends Object>) evaluate("items", getItems())) {
       tagWriter.startTag("option");


miso-web/src/main/webapp/scripts/shared_hot.js
```diff
@@ -0,0 +1,264 @@
...
+    })[0].id;
+  },
+
+  /**
+   * Gets the alias of an object from a given id and collection
+   */
+  getAliasFromId: function (id, referenceCollection) {
+    return referenceCollection.filter(function (item) {
+      return item.id == id;
+    })[0].alias;
```
apmasell: This might do well with a check for length to avoid the case where the id isn't present.

miso-web/src/main/webapp/scripts/shared_hot.js
```diff
@@ -0,0 +1,264 @@
...
+   prepopulateSaveMessages: function (itemsData) {
+     Hot.messages.success = itemsData.map(function (item) { return (item.saved === true ? item.alias : null); });
+   },
+
+  /**
+   * Removes bottom rows if bottom rows are empty (usually from the user dragging too far down)
+   */
+  removeEmptyBottomRows: function (tableData) {
+    while (Hot.startData.length > 1 && Hot.hotTable.isEmptyRow(tableData.length - 1)) {
+      Hot.hotTable.alter('remove_row', parseInt(tableData.length - 1), keepEmptyRows = false);
```
apmasell: Why is there a `parseInt` here?

miso-web/src/main/webapp/scripts/shared_hot.js
```diff
@@ -0,0 +1,264 @@
...
+  },
+
+  /**
+   * Creates the error message and adds it to the page. Marks cells invalid if necessary. Enables save button.
+   */
+  addErrors: function (messages) {
+    console.log(messages);
+    var errorMessages = document.getElementById('errorMessages');
+    var ary = ["<ul>"];
+    for (var i=0; i<messages.failed.length; i++) {
```
apmasell: This should probably just be a `map`.

miso-web/src/main/webapp/scripts/shared_hot.js
```diff
@@ -0,0 +1,264 @@
...
+      ary.push("<li>"+ messages.failed[i] +"</li>");
+    }
+    ary.push("</ul>");
+    errorMessages.innerHTML = '';
+    errorMessages.innerHTML = ary.join('');
```
apmasell: I'm not sure why this gets assigned twice.

apmasell:
Good. :-) Most of my comments can either be changed or addressed with comments if they need to be like that for some non-obvious reason.

miso-web/src/main/webapp/scripts/shared_hot.js
callunity: If there isn't an id present, then we have bigger issues...

PR #223, opened 2016-05-03
Handle nulls in new boxes

apmasell:


spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/LimsBindingInitializer.java
```diff
@@ -531,6 +537,32 @@ public void setAsText(String element) throws IllegalArgumentException {
       }
 
     });
+    binder.registerCustomEditor(BoxUse.class, new PropertyEditorSupport() {
+      @Override
+      public void setAsText(String element) throws IllegalArgumentException {
+        long id = Long.parseLong(element);
+        try {
+          setValue(sqlBoxDAO.getUseById(id));
+        } catch (IOException e) {
```
djcooke: Should this not be thrown? Seems the error will be hidden and bad boxes sent to the service/requestManager layer.

Same with below for BoxSize

callunity: It's now thrown.

djcooke:
:+1:

callunity:
:+1:

PR #225, opened 2016-05-05
Switch to use LibraryType instead of paired

apmasell:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/LibraryDesign.java
```diff
@@ -31,13 +29,11 @@ public static boolean validate(Library library, Iterable<LibraryDesign> rules) {
...
+  @JoinColumn(name = "libraryType", nullable = false)
+  private LibraryType libraryType;
   @Column(nullable = false)
   private String name;
-  @Column(nullable = false)
-  private Boolean paired;
-  @Enumerated(EnumType.STRING)
-  @Column(nullable = false)
-  private PlatformType platformName;
   @OneToOne(targetEntity = SampleClassImpl.class)
```
callunity: Since `paired` does need to be set on the object, can we use LibraryType to infer that? Need to talk to the lab...

apmasell: Yes. The library type will be either Single, Mate Paired, or Paired End. I'm still unclear on how mate paired is different from paired end. There isn't an explicit `paired` column in the `LibraryType`s, but I can add one.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/LibraryDesign.java
```diff
@@ -112,8 +100,7 @@ public void setSuffix(String suffix) {
   public boolean validate(Library library) {
     if (library.getSample().getSampleAdditionalInfo() == null) return true;
     if (library.getSample().getSampleAdditionalInfo().getSampleClass().getSampleClassId() != sampleClass.getSampleClassId()) return false;
-    if (!library.getPlatformName().equals(platformName.getKey())) return false;
-    if (library.getPaired().booleanValue() != paired.booleanValue()) return false;
+    if (library.getLibraryType().getLibraryTypeId() != libraryType.getLibraryTypeId()) return false;
     if (library.getLibrarySelectionType().getLibrarySelectionTypeId() != librarySelectionType) return false;
```
callunity: Masellclash!

callunity:
:+1:

tonydebat:
👍

PR #228, opened 2016-05-06
GLT-704 Fix bad migration

callunity:
V0022 failed because `text`-type columns cannot have defaults. Since this was a recent commit, we've agreed to update the migration itself, manually do the changes required to fix our db: 
- compile with the commit below
- `DROP TABLE LibraryDesign`
- re-run the `CREATE TABLE LibraryPropagationRule` from [V0010](https://github.com/oicr-gsi/miso-lims/commit/4bcfd21f3480f89e8470e691cac62b29f79b13ea#diff-eeebc84bf2b2309ec4e54271f1cd2b65R334)
- run `flyway:repair` on the database, which will roll it back to V0021
- run migrations as usual

This way, we fix our db once and nobody in the future will have to fix this failed migration.


apmasell:
:+1:

djcooke:
:+1:

PR #230, opened 2016-05-06
GLT-690 interpolate properties instead of waiting for MisoAppListener to set them

callunity:
MisoAppListener is sometimes too slow (ie. one naming scheme is autowired into the LibraryControllerHelperService, and another into the SQLLibraryDAO).

Also sets properties in the controllers in a far cleaner way.


apmasell:
:+1: Dark magic, but if it works.

callunity:
The weird part is that there's this same interpolation just ten lines above, where `${miso.fileStorageDirectory}` is set as a property value. We're just making the code base more consistent!

apmasell:
:+1: :100:

djcooke:
:+1:

PR #231, opened 2016-05-09
GLT-664 Refactor shared Handsontable code to own module

callunity:
Extracts code shared between Sample and Library uses of Handsontable. Includes changes made in response to comments on PR #222 .


apmasell:
:+1:  Mmm, code deletion.

djcooke:
:+1:

PR #232, opened 2016-05-09
GLT-698 Add sex field to Identity

callunity:


apmasell:
:+1: Should the default be unknown?

callunity:
The lab cares about this but SeqWare doesn't, so I'm inclined to leave "Unknown" as a user-entered field for when they don't know, and NULL for when the information hasn't been collected.

djcooke:
"Unknown" vs. null is confusing. If the information hasn't been collected then it's unknown, and if it has been collected then it shouldn't be unknown... right?

Would it be better to instead use the following?
* "Unspecified" for information collected and intentionally entered this way
* "Unknown" as a default to replace null

callunity:
Well, when you both put it that way... :) I'll change it to "Unknown" and "Unspecified" as per @djcooke 's suggestion.

djcooke:
:+1: once load and save is tested with fixes in #233

PR #234, opened 2016-05-11
Refactor tagged barcodes to live in the database

apmasell:


sqlstore/src/main/resources/db/migration/V0023__tag_barcode_refactor.sql
```diff
@@ -0,0 +1,45 @@
...
+  ('ILLUMINA', 'NEXTflex 6bp');
+
+ALTER TABLE TagBarcodes ADD COLUMN position int DEFAULT 1;
+ALTER TABLE TagBarcodes ADD COLUMN tagFamilyId bigint(20);
+UPDATE TagBarcodes SET tagFamilyId = (SELECT tagFamilyId FROM TagBarcodeFamily WHERE TagBarcodeFamily.platformType = UPPER(TagBarcodes.platformName) AND TagBarcodeFamily.name = TagBarcodes.strategyName);
+ALTER TABLE TagBarcodes CHANGE COLUMN tagFamilyId tagFamilyId bigint(20) NOT NULL;
+ALTER TABLE TagBarcodes ADD FOREIGN KEY (tagFamilyId) REFERENCES TagBarcodeFamily (tagFamilyId);
+ALTER TABLE TagBarcodes DROP COLUMN platformName;
+ALTER TABLE TagBarcodes DROP COLUMN strategyName;
+
```
callunity: The Nextera Dual Index N5** barcodes need to have `position = 2` set.

apmasell: Fixed.

callunity:
On the Edit Library page, the Barcoding Family dropdown still displays "Please select a platform", even after platform change.

callunity:
:+1:

sqlstore/src/main/resources/db/migration/V0023__tag_barcode_refactor.sql
```diff
@@ -0,0 +1,45 @@
...
+-- Set positions in dual index barcodes
+UPDATE TagBarcodes SET position = 2 WHERE name LIKE 'N5%' AND tagFamilyId = (SELECT tagFamilyId FROM TagBarcodeFamily WHERE name = 'Nextera Dual Index');
+UPDATE TagBarcodes SET position = 2 WHERE name LIKE 'S%7%' AND tagFamilyId = (SELECT tagFamilyId FROM TagBarcodeFamily WHERE name = '16S V4 Dual Index');
+
```
callunity: S5** barcodes are also Index 2 ([source](http://seq.liai.org/204-2/))

djcooke:
HibernateTagBarcodeDao should be accessed through RequestManager or a Service class (preferring the latter as that's what we've been transitioning towards). Otherwise, :+1:

apmasell:
I had the service before, but the interface is exactly identical to the DAO interface. Do you still want it?

djcooke:
It's better to have the service for consistency, and in case we need to add business logic (which should be in the service layer). For example, if we added a method to create tagbarcodes, this should be admin-only and the service layer should do authorization checks

PR #235, opened 2016-05-11
GLT-700 use Group ID and Group Description as text fields

callunity:
The original thought had been to control the vocabulary of Group ID and Group Description. However, this implementation would prove challenging for migration, and we will have to study the lab further in order to determine whether controlling this vocabulary would be useful.

With that in mind, this commit implements Group ID and Group Description as text fields on Sample Analyte (LibraryAdditionalInfo inherits these fields from Sample). The controlled vocabulary SampleGroup code has been commented out but left in place in case we end up making use of it later. (Thoughts on whether it should be deleted outright?)

### DIFF
--- a/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryAdditionalInfoService.java
+++ b/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryAdditionalInfoService.java
@@ -41,9 +41,6 @@ public class DefaultLibraryAdditionalInfoService implements LibraryAdditionalInf
   private TissueTypeDao tissueTypeDao;
   
   @Autowired
-  private SampleGroupDao sampleGroupDao;
-  
-  @Autowired
   private KitStore kitStore;
   
   @Autowired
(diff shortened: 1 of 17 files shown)


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryAdditionalInfoService.java
```diff
@@ -81,9 +87,13 @@ public void update(LibraryAdditionalInfo libraryAdditionalInfo)
     LibraryAdditionalInfo updated = get(libraryAdditionalInfo.getLibraryId());
     updated.setTissueOrigin(tissueOriginDao.getTissueOrigin(libraryAdditionalInfo.getTissueOrigin().getId()));
     updated.setTissueType(tissueTypeDao.getTissueType(libraryAdditionalInfo.getTissueType().getId()));
+    /**
```
apmasell: Please delete, rather than comment out. There are others too.

apmasell:
:+1:

tonydebat:
👍 Please delete all the commented out code.

callunity:
Commented out code is deleted.

PR #238, opened 2016-05-16
GLT-638 & GLT-619

callunity:
- GLT-619: remove stray breadcrumb on bulk create samples page
- GLT-638: use aliases in breadcrumbs (with names for hover-over)
- bonus: remove some unneeded imports


djcooke:
:+1:

tonydebat:
👍

PR #239, opened 2016-05-16
fix sample service create method for propagation

djcooke:
Makes the DefaultSampleService more flexible in how parent information is specified (may be in sample.sampleAddtionalInfo.parent or sample.identity). Also includes binding fix for sample.groupId


callunity:
:+1:

tonydebat:
👍

PR #241, opened 2016-05-17
GLT-720 fix set tagbarcode error

callunity:
plus bonus fix for libraryKit, which wasn't getting marked as required.


apmasell:
:+1:

djcooke:
:+1:

PR #244, opened 2016-05-18
fixed filtering of pools per platform request

djcooke:


callunity:
:+1:

apmasell:
:+1: This is one of those rare cases where the more efficient code is also shorter. ;-)

PR #245, opened 2016-05-19
GLT-676 fix partition container pencil bug

callunity:
- fix for the occasion when you click a second time on the "edit ID Barcode" pencil for a container, and it fills the container's ID Barcode text field with HTML to make the text field
- also contains JS formatting fixes


apmasell:
:+1: I wonder how many pencil-related tickets we've had so far.

djcooke:
:spaghetti: :+1:

PR #246, opened 2016-05-19
Allow multiple status in pool order completions

apmasell:
Instead of having simply requested and completed, provide pool order completion
counts for all possible sequencing states.

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/type/HealthType.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/type/HealthType.java
@@ -24,6 +24,7 @@
 package uk.ac.bbsrc.tgac.miso.core.data.type;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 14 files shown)


callunity:
:+1: to everything but "Completion", but I haven't come up with a better name.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/CompletionGroup.java
```diff
@@ -0,0 +1,41 @@
+package uk.ac.bbsrc.tgac.miso.core.data;
+
+import java.util.TreeMap;
+
+import uk.ac.bbsrc.tgac.miso.core.data.type.HealthType;
+
+public class CompletionGroup extends TreeMap<HealthType, PoolOrderCompletion> {
```
djcooke: Consider renaming to PoolOrderCompletionGroup to match PoolOrderCompletion and avoid ambiguity?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/type/HealthType.java
```diff
@@ -35,10 +36,20 @@
  * @since 0.0.2
  */
 public enum HealthType {
-  Unknown("Unknown"), Completed("Completed"), Failed("Failed"), Started("Started"), Stopped("Stopped"), Running("Running");
+  Unknown("Unknown", 0), Completed("Completed", -1), Failed("Failed", 0), Started("Started", -1), Stopped("Stopped", -1), Running("Running",
+      -1), Requested("Requested", 1);
```
djcooke: HealthType was meant to be used for Runs, which should still be limited to the same options. Probably not worth making a separate enum, so maybe just throw a RuntimeException from AbstractRun.setStatus if trying to set it to HealthType.Requested (and Javadoc this)?

apmasell: Sure. If added an extra property to `HealthType` where is describes if it can come from the sequencer and `StatusImpl` will throw if it's something crazy.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/CompletionGroup.java
apmasell: Done.

apmasell:
Updated.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/type/HealthType.java
djcooke: Even better :+1:

djcooke:
:+1:

PR #251, opened 2016-05-20
Add any missing barcode strategies

apmasell:


callunity:
:+1:

djcooke:
Will this not break migration since V0023 is already released?

apmasell:
Yes. But it won't migrate for the TGAC production DB, so it's easier to fix our DB's checksums.

djcooke:
:+1:

PR #252, opened 2016-05-20
GLT-693 show helpful message for HOT 500-series errors

callunity:
Well, more helpful than the existing nothing. At least it lets the user know it wasn't their fault.


apmasell:
:+1: Maybe “file a ticket” instead of email?

callunity:
"file a ticket" is much better.

djcooke:
:+1: after removing the log message

PR #253, opened 2016-05-20
GLT-661 hide HOT IDBC col if BCs are autogenerated

callunity:
- visibility of Matrix (Identification) Barcode column in Handontable (sample & library) is turned off if these barcodes are auto-generated (set in miso.properties)


apmasell:
:+1:

djcooke:
:+1:

PR #254, opened 2016-05-24
GLT-731 updated maven config for metrix dependency

tonydebat:
Also updated for one-jar and open symphony oscore as they are no longer available from googlecode.


apmasell:
:+1:

djcooke:
:+1:

PR #257, opened 2016-05-24
GLT-611 return "Pool not found" instead of NPE

callunity:
Fix for the NPE that's returned when looking up a pool by an invalid barcode.


apmasell:
:+1:

djcooke:
:+1:

djcooke:
PS: RunControllerHelperService is full of HTML StringBuilding :cry:

callunity:
I know, and it was a source of pain here.

PR #258, opened 2016-05-25
GLT-608 fix OxfordNanopore enum & LibAddInfo save

callunity:
- fixes the enum error found when adding an OxfordNanopore container
- also finishes the fix for GLT-729 (corrects the check for resource writability)


apmasell:
:+1:

djcooke:
:+1:

PR #261, opened 2016-05-26
added migration utility with load generator

djcooke:
Includes several minor changes outside of (but mostly required for) migration utility:
* added setters for a lot of Service and DAO fields
* fixed hashcode and equals in Dilutions (avoid NPEs and compare more fields)
* use lazyGet when reloading Project in sample service to avoid reloading all project samples unnecessarily
* added missing check for metrix status on Edit Run page
* added null checks to avoid NPE on cacheManager in Run and Project DAOs


miso-web/src/main/webapp/pages/editRun.jsp
```diff
@@ -696,7 +696,7 @@
     <c:choose>
     <c:when test="${not empty run.platformType}">
       Run.pool.poolSearch("", '${run.platformType.key}');
-      <c:if test="${run.id != 0}">
+      <c:if test="${run.id != 0 and metrixEnabled}">
```
apmasell: This was a bug the whole time?

apmasell:
:+1:

miso-web/src/main/webapp/pages/editRun.jsp
djcooke: Yes, but only a noticeable problem when a run is missing a status, which normally doesn't happen. I noticed it when generating runs before I was also generating status for them. There's another check for metrixEnabled for querying the interop files. This one is for updating run status

callunity:
:+1:

PR #262, opened 2016-05-26
report configured path instead of canonical from notification server

djcooke:
If you wish to have the canonical path reported, you can still accomplish this by setting the canonical path instead of using symbolic links in notification.properties.

Also replaced some Strings with constants


callunity:
:+1:

apmasell:
:+1:

PR #263, opened 2016-05-26
GLT-566 Make all barcodes available on page

callunity:
* update TagBarcodes ModelAttribute to contain all barcodes
 (and Handsontable code to match); reduces AJAX calls
* EditLibraryController: change (TagBarcode)Strategy to TagBarcodeFamily
* add "No barcode" option for each platform
* update LibAddInfoService to set library before save

### DIFF
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
@@ -34,6 +34,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


apmasell:
:+1: Thanks for killing the last use of strategy.

tonydebat:
With a fresh db migrating the OICR `V8001__standard_oicr_config.sql` causes the error:
```
[WARNING] DB: Unknown column 'strategyName' in 'where clause' (SQL State: 42S22 - Error Code: 1054)
[ERROR] Migration of schema `miso` to version 8001 failed! Please restore backups and roll back database and code!
```

callunity:
@tonydebat L hmm, that's a problem. I imagine you'll encounter that on any branch right now, as strategyName was changed in the last release. Can you make a ticket for that?

tonydebat:
👍

PR #264, opened 2016-05-27
Analysis server integration

frankie13:
I deleted the analysis server and added some bits to integrate it with my new thing.


integration-tools/src/main/java/uk/ac/bbsrc/tgac/miso/integration/AnalysisQueryService.java
```diff
@@ -62,7 +62,9 @@ public JSONArray getTask(String taskId) throws IntegrationException {
     q1.put("params", params);
     String query = q1.toString();
 
-    String response = IntegrationUtils.sendMessage(IntegrationUtils.prepareSocket(analysisServerHost, analysisServerPort), query);
+//    String response = IntegrationUtils.sendMessage(IntegrationUtils.prepareSocket(analysisServerHost, analysisServerPort), query);
```
apmasell: Please delete rather than comment out.

apmasell:
:+1: Can `sqlstore/src/main/resources/schemas/conan-schema-20120601.sql` also be removed?

miso-web/src/main/webapp/scripts/task_ajax.js
```diff
@@ -300,11 +242,12 @@ Tasks.ui = {
...
         for (var k = 0; k < process.parameters.length; k++) {
-          if (jQuery('input[name="'+process.parameters[k].name+'"]').length === 0) {
-            jQuery('#'+pipeline.name+'-reqParams > tbody:last')
-              .append(jQuery('<tr>')
-              .append("<td>"+process.parameters[k].name+"</td><td><input style='width:98%' optional='true' type='text' id='"+process.parameters[k].name+"' name='"+process.parameters[k].name+"' value=''/>")
-            );
+          if (jQuery('input[name="' + process.parameters[k].name + '"]').length == 0) {
+            jQuery('#' + pipeline.name + '-reqParams > tbody:last').append(
+                jQuery('<tr>').append("<td>" + process.parameters[k].name + "</td>").append(
+                    "<td><input style='width:98%' optional='true' type='text'").append(
```
callunity: Adding an `append` here will cause jQuery to close the `input` element before appending, so you'll end up with an input box followed by text of `id='name' name='name'...`.

Is there a particular reason for using `append`? It's is over twice as slow as `+`...

callunity:
:+1: after fixing the `append` in the middle of an `input` element

miso-web/src/main/webapp/scripts/task_ajax.js
frankie13: fixed!

PR #265, opened 2016-05-27
GLT-738 redirect to sequencer ref after saving it

callunity:
(instead of to awkward alternate List Sequencers page)


apmasell:
:+1:

tonydebat:
👍

PR #267, opened 2016-05-27
added missing migration target/destination package

djcooke:
Package was originally named uk.ac.bbsrc.tgac.miso.migration.target; because 'target/' is gitignored, this did not go over well and the package was missing from the previous PR


apmasell:
:+1:

tonydebat:
👍

callunity:
:+1:

PR #270, opened 2016-06-01
GLT-732 Correct communication between notification server and client

tonydebat:


apmasell:
I'm not sure notification server should be a dependency of MISO directly. Could those constants go into an interface that lives in core?

djcooke:
The notification-consumer-services module doesn't really make sense without notification-server, so I think that dependency makes sense

djcooke:
:+1:

apmasell:
:+1:

PR #271, opened 2016-06-02
Copy groupId and groupDescription to SampleAdditionalInfo

apmasell:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleAnalyteService.java
```diff
@@ -151,8 +151,8 @@ public void applyChanges(SampleAnalyte target, SampleAnalyte source) throws IOEx
     target.setRegion(source.getRegion());
     target.setTubeId(source.getTubeId());
     target.setStrStatus(source.getStrStatus());
-    target.setGroupId(source.getGroupId());
-    target.setGroupDescription(source.getGroupDescription());
+    target.getSample().getSampleAdditionalInfo().setGroupId(source.getSample().getSampleAdditionalInfo().getGroupId());
+    target.getSample().getSampleAdditionalInfo().setGroupDescription(source.getSample().getSampleAdditionalInfo().getGroupDescription());
```
djcooke: These two lines should be removed. This happens in DefaultSampleAdditionalInfoService

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -215,23 +212,30 @@ public Sample get(Long sampleId) throws IOException {
...
       if (!Identity.CATEGORY_NAME.equals(sample.getSampleAdditionalInfo().getSampleClass().getSampleCategory())) {
         sample.getSampleAdditionalInfo().setParent(findOrCreateParent(sample));
         validateHierarchy(sample);
+        if (sample.getSampleAdditionalInfo().getGroupId() == null) {
+          sample.getSampleAdditionalInfo().setGroupId(sample.getSampleAdditionalInfo().getParent().getSampleAdditionalInfo().getGroupId());
+        }
+        if (sample.getSampleAdditionalInfo().getGroupDescription() == null) {
+          sample.getSampleAdditionalInfo()
+              .setGroupDescription(sample.getSampleAdditionalInfo().getParent().getSampleAdditionalInfo().getGroupDescription());
+        }
```
djcooke: I don't think this behaviour is correct. From my understanding, it is fine for a sample to have no group ID if its parent has one. I think the "inherited but changeable" idea here is that the parent's group ID should be default in the UI when creating a sample, but clearing it is valid

miso-web/src/main/resources/external/miso.properties
```diff
@@ -78,7 +78,7 @@ miso.statsdb.enabled:false
 miso.autoGenerateIdentificationBarcodes:false
 
 ##config for classic vs detailed sample info
-miso.detailed.sample.enabled:false
+miso.detailed.sample.enabled:true
```
djcooke: should not be changed in develop

miso-web/src/main/resources/internal/miso.properties
```diff
@@ -70,11 +70,11 @@ miso.statsdb.enabled:false
 miso.autoGenerateIdentificationBarcodes:false
 
 ##config for classic vs detailed sample info
-miso.detailed.sample.enabled:false
+miso.detailed.sample.enabled:true
```
djcooke: should not be changed in develop

sqlstore/src/main/resources/db/migration/V0028__model_refactor.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE SampleAdditionalInfo ADD COLUMN groupId int(10);
+ALTER TABLE SampleAdditionalInfo ADD COLUMN groupDescription varchar(255);
+
+UPDATE SampleAdditionalInfo SET groupId = (SELECT groupId FROM SampleAnalyte WHERE SampleAdditionalInfo.sampleId = SampleAnalyte.sampleId), groupDescription = (SELECT groupDescription FROM SampleAnalyte WHERE SampleAdditionalInfo.sampleId = SampleAnalyte.sampleId);
```
djcooke: Is there a reason not to remove the columns from SampleAnalyte?

tonydebat:
👍

sqlstore/src/main/resources/db/migration/V0028__model_refactor.sql
apmasell: Yes, they will be removed in GLT-746 where Analyte will be split into Stock and Aliquot.

miso-web/src/main/resources/internal/miso.properties
apmasell: Fixed. :-/

miso-web/src/main/resources/external/miso.properties
apmasell: Fixed. :-/

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleAnalyteService.java
apmasell: Fixed.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
apmasell: Okay. I guess that was not true for Analytes. Fixed.

apmasell: Wait, no, I stand by this. This is in the `create` method, not `update`.

djcooke: So if the parent has a group ID, it is ok to set the child's group ID to something different, and this will be saved, but if you clear the group ID, it will get changed back to the parent's group ID in the save process, likely without the user noticing? Then when editing the sample, it's fine to clear it, and that will be saved?

apmasell: No. It's inherited once at the time of creation. If you change the parent the child won't see it. If you set it to null in the child, it will be null.

djcooke:
Application startup fails when attempting to precache samples:
```
2016-06-03 12:33:43 INFO: Precaching. This may take a while.
2016-06-03 12:33:43 INFO: \_ projects...
2016-06-03 12:33:43 INFO: 5
2016-06-03 12:33:43 INFO: \_ studies...
2016-06-03 12:33:43 INFO: 0
2016-06-03 12:33:43 INFO: \_ experiments...
2016-06-03 12:33:43 INFO: 0
2016-06-03 12:33:43 INFO: \_ samples...
2016-06-03 12:33:43 WARN: SQL Error: 1054, SQLState: 42S22
2016-06-03 12:33:43 ERROR: Unknown column 'sampleaddi0_.groupDescription' in 'field list'
2016-06-03 12:33:43 INFO: HHH000327: Error performing load command : org.hibernate.exception.SQLGrammarException: could not extract Resu
ltSet
2016-06-03 12:33:43 INFO: MISO Application Context Destroyed: Fri Jun 03 12:33:43 EDT 2016
```

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
apmasell: Fixed by removal.

djcooke:
:+1:

PR #273, opened 2016-06-03
corrected transactional annotations

djcooke:
In most places, transactions only rolled back on RuntimeExceptions before. Now they will roll back on all Exceptions. Added annotation to all Service and DAO classes (Hibernate and SQL). Removed method-level annotations as none of them required different handling than is set at the class level.

Notes:
* Default propagation is REQUIRED, which means the method must be run in a transaction. If there is an existing transaction, it will be joined; otherwise a new transaction will be started.
* Default readOnly is false
* Many tables in the MISO DB are still MyISAM, which does not support transactions, so these tables will never actually get rolled back. We should convert them all to InnoDB at some point


apmasell:
:+1:

tonydebat:
👍

callunity:
:+1:

PR #275, opened 2016-06-06
added config for miso.properties location

djcooke:
You can now provide an external miso.properties file by adding the following parameter to ROOT.xml:

`<Parameter name="miso.propertiesFile" value="file:${CATALINA_HOME}/conf/Catalina/localhost/miso.properties" override="false"/>`

value must begin with `file:` and you must include `override=false`; otherwise, the parameter will be overridden by the default in web.xml (`classpath:miso.properties`). If this parameter is absent in ROOT.xml, the default in web.xml will be used as well, so current deployments will not be affected and may continue to use the packaged miso.properties file


callunity:
:+1: nice. Can this get added to @apmasell 's documentation?

apmasell:
@callunity I'm doing that now.

PR #276, opened 2016-06-06
Updates to admin-manual.md

callunity:


admin-manual.md
```diff
@@ -26,18 +26,13 @@ Development Machine(s):
 - Eclipse
 - A merge tool such as Meld
 
-# Creating a Fork and Setting Configuration
+# Creating a Fork
 Use the GitHub interface or a private instance to create a forked repository.
 
-Proceed to set up a build environment, then edit
-`miso-web/src/main/webapp/WEB-INF/miso-config.xml` and set the naming schemes.
-The naming schemes will determine how MISO checks if object names (especially
```
apmasell: Writing naming schemes still matters. I'm not sure where it should go.

apmasell:
:+1:

admin-manual.md
callunity: Naming scheme beans in `miso-web/src/main/webapp/WEB-INF/miso-config.xml` are interpolations of the `miso.X.naming.scheme` values set in the miso.properties file. That's why I've moved this part to line 122 below.

admin-manual.md
```diff
@@ -109,21 +104,32 @@ and populate it with the following information:
       <Parameter name="miso.propertiesFile" value="file:${CATALINA_HOME}/conf/Catalina/localhost/miso.properties" override="false"/>
     </Context>
 
+Make sure the database path in the `server.xml` is correct for your install:
```
djcooke: ROOT.xml?

djcooke:
:+1: after server/ROOT.xml fix

PR #278, opened 2016-06-07
Remove unnecessary cloning

callunity:
- pools in watcher list are only ever accessed as read-only objects
- pool cloning is failing in production db


apmasell:
:+1:

djcooke:
:+1:

tonydebat:
👍

PR #279, opened 2016-06-08
sample hierarchy model and dto inheritance

djcooke:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SampleAnalyte.java
```diff
@@ -39,22 +25,6 @@
 
   void setTubeId(String tubeId);
```
callunity: TubeID needs to be deleted. (It's the same as ExternalInstituteIdentifier)

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SampleTissue.java
```diff
@@ -1,42 +1,11 @@
...
 
-  Long getId();
-
-  void setId(Long sampleId);
-
-  Sample getSample();
-
-  void setSample(Sample sample);
-
   Integer getCellularity();
```
callunity: Cellularity is to be deleted altogether.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SampleAnalyteImpl.java
```diff
@@ -46,40 +31,6 @@
   private String region;
```
callunity: Region is to be moved into SampleTissue (not sure if that's relevant to this PR).

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SampleAnalyteImpl.java
```diff
@@ -46,40 +31,6 @@
...
   private String tubeId;
```
callunity: TubeID is to be deleted.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SampleAnalyteImpl.java
```diff
@@ -121,46 +72,6 @@ public void setTubeId(String tubeId) {
...
-    return lastUpdated;
-  }
-
-  @Override
-  public void setLastUpdated(Date lastUpdated) {
-    this.lastUpdated = lastUpdated;
-  }
-
-  @Override
   public StrStatus getStrStatus() {
```
callunity: STR Status goes onto SampleStock.

callunity:
:+1:

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/SampleAdditionalInfoDto.java
```diff
@@ -30,6 +33,8 @@
   private String labUrl;
   private Long groupId;
   private String groupDescription;
+  private Long updatedById;
+  private String updatedByUrl;
```
djcooke: SampleDto could have a lastModifier field, but this updatedBy has been removed from SampleAdditionalInfo. The Dtos.asDetailedSample(SampleAdditionalInfo) method above is actually pulling the value from Sample, not SampleAdditionalInfo

apmasell: Fixed.

tonydebat:
Attempting to create a sample using the single sample page results in the failure:
`ERROR: Field 'createdBy' doesn't have a default value`

Attempting to create a sample using the bulk page results int he failure:
`Could not read JSON: Unrecognized field "externalName" (Class uk.ac.bbsrc.tgac.miso.dto.SampleDto), not marked as ignorable`

apmasell:
Did you apply the migration? All the createdBy columns should be dropped.

djcooke:
create/update user/time fields have been removed from all of the add-on tables since this data can be pulled from Sample change logs. You'll have to run the V0028 migration

apmasell:
Also, what kind of sample were you creating in the bulk upload page?

tonydebat:
Yes, definitely a problem when I migrated. Will do this again. 

I was creating a primary tumour on the bulk upload page.

apmasell:
The `externalName` problem seems to be caused by a missing `sampleOptions` value.

apmasell:
I think I have the `externalName` problem solved, but I'm not sure it will work for all things in the hierarchy since if receiving an aliquot, we have to create the entire ghost hierarchy.

djcooke:
It has never been possible to 'receive' an aliquot. Before this pull request, it was possible to create Tissue and Analyte Stock from the Create Sample page (/sample/new - single and bulk). This pull request should not disrupt those capabilities (or any others)

apmasell:
Well, analyte will have the same problem as tissues did... I'll add the same fields.

apmasell:
I've fixed the problem with the tissues. It won't work for analytes, but since that needs to be restructured, that's intentionally todo.

djcooke:
Making SampleAnalyteDto also extend SampleIdentityDto for now would prevent this PR from breaking anything though, right?

apmasell:
Err, I think it would still be broken since it wouldn't create a tissue.

djcooke:
Tissue is not yet required in the hierarchy. I am working on that in a branch that will be merged later

apmasell:
Okay. I can do that then.

apmasell:
Okay, fixed.

sqlstore/src/main/resources/db/migration/V0028__model_refactor.sql
```diff
@@ -2,3 +2,28 @@ ALTER TABLE SampleAdditionalInfo ADD COLUMN groupId int(10);
 ALTER TABLE SampleAdditionalInfo ADD COLUMN groupDescription varchar(255);
 
 UPDATE SampleAdditionalInfo SET groupId = (SELECT groupId FROM SampleAnalyte WHERE SampleAdditionalInfo.sampleId = SampleAnalyte.sampleId), groupDescription = (SELECT groupDescription FROM SampleAnalyte WHERE SampleAdditionalInfo.sampleId = SampleAnalyte.sampleId);
+
+ALTER TABLE SampleAdditionalInfo DROP FOREIGN KEY `FKeutn2473w3yr16khgalspuviw`; -- createdBy fkey
+ALTER TABLE SampleAdditionalInfo DROP FOREIGN KEY `FKp8bvx3e7jsmnyw51toi7mq7cq`; -- updatedBy fkey
```
tonydebat: The inline comments cause the flyway migration to fail for me. I get the error:
`[ERROR] Message    : You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'ALTER TABLE SampleAnalyte DROP FOREIGN KEY `FKprqyhv40bntjrf5l64mjdgl1j`; -- upd' at line 2`

If the comments occur at the beginning of the line they work okay with flyway. (I'm using flyway via maven.) Changing to something like the following would be fine:

```
-- createdBy fkey
ALTER TABLE Identity DROP FOREIGN KEY `FKauqylg2sle5eudy0tqabtlmsb`; 
-- updatedBy fkey
ALTER TABLE Identity DROP FOREIGN KEY `FKa8c6e56hg9iucguhr0dcse62h`; 
```

apmasell: So, having auto-generated FK names seems to be a problem because they aren't consistent. I have no idea how to address this. :-/

tonydebat: It's the not FK names that are the problem. It's the comment (`-- createdBy fkey`) location that's the problem. When they're inline the migration doesn't work. Putting them on their own line solves the problem.

djcooke: The FK names are declared in the migration script, so they are named consistently (at least the ones in question). They mostly are randomly generated and not nice names, but that's not a problem. I'll fix the comments

djcooke: fixed

callunity:
I attempted to update an existing Sample (Analyte Stock) through Handsontable and got a 500 Server error: 
`Invalid Donor Sex: null`. (Donor Sex was not visible on the UI, and I was not trying to change it.)

apmasell:
I think I fixed the `donorSex` issue...twice for good measure.

djcooke:
On propagate to sample, hitting save results in endless progress bar
```
TypeError: referenceCollection.filter(...)[0] is undefined
	Hot.getIdFromAlias()
	 header_script.js:20605
	Sample.hot.buildSampleDtosFromData()
	 sample_hot.js:972
	Sample.hot.getArrayOfNewObjects()
	 sample_hot.js:1232
	Sample.hot.propagateData/<()
	 sample_hot.js:1399
	ValidatorsQueue/<.checkIfQueueIsEmpty()
	 handsontable.full.js:5043
	ValidatorsQueue/<.removeValidatorFormQueue()
	 handsontable.full.js:5037
	Core/this.validateCells/<()
	 handsontable.full.js:5663
	done()
	 handsontable.full.js:5146
	Core/this.validateCell/</<()
	 handsontable.full.js:5161
	validateAlias()
	 sample_hot.js:839
	Core/this.validateCell/<()
```

djcooke:
Fixed the problem I mentioned above. Next problem when propagating analyte stock->aliquot:

```
uk.ac.bbsrc.tgac.miso.core.exception.MisoNamingException: Custom naming generator 'OicrSampleAliasGenerator' supplied for Sample field 'alias' generated an invalid name according to the validation scheme '^([A-Z\d]{3,5})_(\d{3,5})(_(Ad|Ap|Ag|Bm|Bn|Br|Bu|Cb|Cn|Du|Es|Fs|Gb|Hr|Ki|Le|Li|Ln|Lu|Lv|Lx|Ly|Md|Me|Nk|Oc|Om|Ov|Pa|Pb|Pr|Sa|Sg|Si|Sk|Sm|Sp|St|Ta|Tr|Mu|Wm|nn)_[BRPXMCFESATOn]_(nn|\d{2})_(\d{1,2})-(\d{1,2})(_(C|CV|HE|LCM|D_S|R_S|D_|R_(\d+_(MR|SM|WT)_)?)\d+)?)?$' (generated name: TEST_0018_Ad_P_nn_null_D_1)
        at uk.ac.bbsrc.tgac.miso.core.service.naming.OicrSampleNamingScheme.generateNameFor(OicrSampleNamingScheme.java:77)
        at uk.ac.bbsrc.tgac.miso.core.service.naming.OicrSampleNamingScheme.generateNameFor(OicrSampleNamingScheme.java:19)
        at uk.ac.bbsrc.tgac.miso.service.impl.DefaultSampleService.save(DefaultSampleService.java:259)
        ... 69 more
```

Not sure how this is happening, as the 'null' part is passageNumber, which should use 'nn' in place of null

create/edit/propagate single/bulk should all work now except analyte stock>aliquot propagation

callunity:
List Samples page: selected three entities of the same sample class and selected "Propagate (sample) selected". JS error:
`TypeError: Sample.sampleClasses is undefined
Stack trace:
Sample.ui.insertSampleClassesDropdown@http://localhost:8080/scripts/header_script.js:10699:9
Sample.ui.checkForPropagate@http://localhost:8080/scripts/header_script.js:10586:7
onchange@http://localhost:8080/miso/samples:1:1`

djcooke:
I believe it should all work now

callunity:
:+1: works as advertised.

morgantaschuk:
:smiley:

tonydebat:
👍

PR #281, opened 2016-06-09
allow project alias containing numbers in OICR sample alias

djcooke:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/OicrSampleNamingScheme.java
```diff
@@ -20,7 +20,7 @@
 public class OicrSampleNamingScheme implements RequestManagerAwareNamingScheme<Sample> {
   protected static final Logger log = LoggerFactory.getLogger(OicrSampleNamingScheme.class);
 
-  private static final String IDENTITY_REGEX_PART = "([A-Z\\d]{3,5})_(\\d{3,5})";
+  private static final String IDENTITY_REGEX_PART = "([A-Z0-9\\d]{3,5})_(\\d{3,5})";
```
callunity: Doesn't the `\\d` do the same thing as the `0-9`?

djcooke: Oh, haha. yes

PR #282, opened 2016-06-10
Refactor finding of temp folder for id barcodes

callunity:
This will hopefully solve the issue where prod is creating a folder called `ROOTtemp` at the same level as `ROOT` and saving the barcodes there.


apmasell:
:+1:

tonydebat:
👍

PR #284, opened 2016-06-14
GLT-702 fix recursion

callunity:
Adds annotations between Library & LibraryAdditionalInfo, and TagBarcode & TagBarcodeFamily.

### DIFF
--- a/core/src/test/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractLibraryTestSuite.java
+++ b/core/src/test/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractLibraryTestSuite.java
@@ -1,9 +1,14 @@
 package uk.ac.bbsrc.tgac.miso.core.data;
 
+import static org.hamcrest.core.StringContains.containsString;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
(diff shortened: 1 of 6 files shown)


djcooke:
:+1:

core/src/test/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractLibraryTestSuite.java
```diff
@@ -44,9 +53,16 @@ public void testInheritPermissionsException() {
...
+    Sample sample = new SampleImpl();
+    Library library = new LibraryImpl();
+    LibraryAdditionalInfo lai = new LibraryAdditionalInfoImpl();
+    library.setSample(sample);
+    lai.setTissueOrigin(new TissueOriginImpl());
+    lai.setTissueType(new TissueTypeImpl());
+    lai.setLibrary(library);
+    library.setLibraryAdditionalInfo(lai);
+    mapper.writer().writeValueAsString(library);
+  }
```
tonydebat: Please add an assert to this test. (I guess, you know it's working when it doesn't throw an exception at the moment. But better to also positively test that the string contains a value that you want.) You can use the `containsString` matcher to avoid needing to match the entire string exactly.

tonydebat:
👍 After test updated

callunity:
Test is updated.

PR #288, opened 2016-06-16
updated OICR sample alias generation

djcooke:
Changes made in advance of moving fields. Should not disrupt current name generation


callunity:
:+1:

apmasell:
:+1:

PR #290, opened 2016-06-17
require and create tissue sample in hierarchy

djcooke:
NOTE: Before this pull request, the create samples page (/miso/sample/new) can be used to create Tissue or Analyte Stock samples. This PR will disrupt the ability to create Analyte Stock until GLT-766 (single) and GLT-767 (bulk) are merged


apmasell:
:+1: Why is the delete excluded from the test?

djcooke:
A test would be good for future-proofing I guess, but the delete method is pretty simple, doesn't really do anything itself, and is and unlikely to change:
```
  public void delete(Long sampleId) throws IOException {
    authorizationManager.throwIfNonAdmin();
    Sample sample = get(sampleId);
    sampleDao.deleteSample(sample);
  }
```
Could assert that it calls authorizationManager.throwIfNonAdmin() and sampleDao.deleteSample(sample), but I'm not sure if there's a point?

tonydebat:
👍

apmasell:
Sorry, I meant in the `DELETE` in the SQL migration.

djcooke:
Oh! H2 (test database) doesn't like the MySQL "DELETE [table] from [table]" syntax. It could have been done with a subquery instead, but it wouldn't affect current tests anyway. When tests are added for the SampleValidRelationship DAO, the table should be cleared and populated with test data for better control, as is done for other entities.

PR #292, opened 2016-06-17
updated Edit Sample (single) for required tissue in hierarchy

djcooke:
Creating analyte stock on create single sample page will work again. Bulk page will be fixed in GLT-767


apmasell:
:+1:

callunity:
:+1:

PR #294, opened 2016-06-20
update Edit Sample (bulk) for required tissue in hierarchy

djcooke:
Creating stock analyte on the bulk create sample page works again, and causes a parent tissue to be created in addition to the identity. All create and propagate methods should function correctly again


apmasell:
:+1:

callunity:
:+1: works well. I still don't like having to enter `Tissue Class`, `Tissue Type` and `Sample Class`, so I've created a JIRA ticket to check with the lab about whether there's any way to streamline this.

PR #296, opened 2016-06-21
GLT-750 Add TissueProcessing entities

callunity:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SampleCVSlideImpl.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Override
+  public void setCuts(Integer cuts) {
+    this.cuts = cuts;
+  }
+
+  @Override
+  public Integer getCutsRemaining() {
+    Integer cutsConsumed = 0;
```
apmasell: This should be `int`.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SampleCVSlide.java
```diff
@@ -0,0 +1,21 @@
+package uk.ac.bbsrc.tgac.miso.core.data;
+
+public interface SampleCVSlide extends SampleTissueProcessing {
+
+  public static final String SAMPLE_CLASS_NAME = "CV Slide";
+
+  Integer getCuts();
```
apmasell: I thought we were going to have one CVSlide that had a pile of cuts in it...

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/DetailedSampleBuilder.java
```diff
@@ -745,10 +801,21 @@ public SampleAdditionalInfo build() {
...
+        cvSlide.setCuts(cuts);
+        cvSlide.setDiscards(discards);
+        cvSlide.setThickness(thickness);
+        sample = cvSlide;
+      } else if (sampleClass.getAlias() == SampleLCMTube.SAMPLE_CLASS_NAME) {
+        SampleLCMTube lcmTube = new SampleLCMTubeImpl();
+        lcmTube.setCutsConsumed(cutsConsumed);
+        sample = lcmTube;
+      } else {
+        SampleAdditionalInfo processing = new SampleAdditionalInfoImpl();
```
djcooke: Should be a SampleTissueProcessingImpl

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SampleCVSlideImpl.java
callunity: will do

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SampleCVSlide.java
callunity: We need to speak with Ilinca to confirm the actual requirements, as you and I got different information from her.

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -702,6 +712,60 @@ private static SampleTissue toTissueSample(SampleTissueDto from) {
     return to;
   }
 
+  private static SampleTissueProcessingDto asTissueProcessingSampleDto(SampleTissueProcessing from) {
+    SampleTissueProcessingDto dto = null;
+    if (from.getClass() == SampleCVSlideImpl.class) {
```
djcooke: getClass comparison is required for Dtos because of the way their inheritance works (analyte is also a tissue and an identity for the purpose of including parent data), but instanceof should be used here (and typically for the non-dto models), and check vs. interfaces for flexibility. e.g. we could add another implementation of SampleCVSlide that this method should still work for, or we could subclass SampleCVSlideImpl (however unlikely)

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/SampleTissueProcessingDto.java
```diff
@@ -0,0 +1,21 @@
...
+import uk.ac.bbsrc.tgac.miso.core.data.SampleTissueProcessing;
+
+@JsonSerialize(include = JsonSerialize.Inclusion.NON_NULL)
+@JsonTypeName(value = SampleTissueProcessing.CATEGORY_NAME)
+public class SampleTissueProcessingDto extends SampleTissueDto {
+
+  @Override
+  public void writeUrls(URI baseUri) {
+    super.writeUrls(baseUri);
+    setUrl(UriComponentsBuilder.fromUri(baseUri).path("/rest/sample/tissueprocessing/{id}").buildAndExpand(getId()).toUriString());
```
djcooke: url should just point to regular detailed sample url (`/rest/tree/sample/{id}`)

miso-web/src/main/webapp/pages/editSample.jsp
```diff
@@ -521,7 +565,47 @@
           </table>                               
         </div>
       </c:if>
-      
+
+      <c:if test="${sampleCategory eq 'Tissue Processing' or sampleCategory eq 'new'}">
```
djcooke: should only be included if `${sampleCategory eq 'Tissue Processing'}` since you can't "receive" a Tissue Processing at this time, and the 'single' mode of this page cannot be used for propagation

miso-web/src/main/webapp/scripts/sample_ajax.js
```diff
@@ -163,7 +163,34 @@ var Sample = Sample || {
...
+
+          // Thickness validation
+          jQuery('#thickness').attr('class', 'form-control');
+          jQuery('#thickness').attr('data-parsley-type', 'number');
+          break;
+        case 'LCM Tube':
+          jQuery('#cutsConsumed').attr('class', 'form-control');
+          jQuery('#cutsConsumed').attr('data-parsley-type', 'digits');
+          break;
+        }
```
djcooke: Missing break for outer case

miso-web/src/main/webapp/scripts/sample_ajax.js
```diff
@@ -709,18 +741,48 @@ Sample.ui = {
       break;
     }
   },
+
+  hideTissueProcessingFields: function() {
```
djcooke: showing/hiding tissueProcessing stuff in here probably isn't necessary, since it would be controlled in the JSP. This stuff is used when changing SampleClass, which you can only do when receiving (not propagating) a new sample

apmasell:
:+1:

miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -25,29 +25,23 @@ Sample.hot = {
...
-        if (sam.labId) {
-          sam.labComposite = Sample.hot.getLabCompositeFromId(sam.labId, Hot.sampleOptions.labsDtos);
-        }
+        if (sam.parentSampleClassId) sam.parentSampleClassAlias = Hot.getAliasFromId(sam.parentSampleClassId, Hot.sampleOptions.sampleClassesDtos);
+        if (sam.tissueOriginId) sam.tissueOriginAlias = Hot.getAliasFromId(sam.tissueOriginId, Hot.sampleOptions.tissueOriginsDtos);
+        if (sam.tissueTypeId) sam.tissueTypeAlias = Hot.getAliasFromId(sam.tissueTypeId, Hot.sampleOptions.tissueTypesDtos);
+        if (sam.labId) sam.labComposite = Sample.hot.getLabCompositeFromId(sam.labId, Hot.sampleOptions.labsDtos);
+        if (sam.prepKitId) sam.prepKitAlias = Hot.getAliasFromId(sam.prepKitId, Hot.sampleOptions.kitDescriptorsDtos);
+        if (sam.subprojectId) sam.subprojectAlias = Hot.getAliasFromId(sam.subprojectId, Hot.sampleOptions.subprojectsDtos);
+        if (sam.tissueMaterialId) sam.tissueMaterialAlias = Hot.getAliasFromId(sam.tissueMaterialId, Hot.sampleOptions.tissueMaterialsDtos);
```
djcooke: This field will be moving, but has not been moved yet, so should still be in analyte

miso-web/src/main/webapp/scripts/sample_ajax.js
callunity: Added.

djcooke:
:+1:

PR #300, opened 2016-06-22
GLT-758 Converted deep cloning of projects to partialCopy of projects.

tonydebat:
Converted deep cloning of pools to partialCopy of pools.
Corrected the generic types for all use of the type Pool.
Removed duplicated PoolConstructionCompleteResponderService.

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/event/manager/ProjectAlertManager.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/event/manager/ProjectAlertManager.java
@@ -26,6 +26,7 @@ package uk.ac.bbsrc.tgac.miso.core.event.manager;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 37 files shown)


apmasell:
:+1: You're a hero for fixing all those generic types.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/event/manager/ProjectAlertManager.java
```diff
@@ -296,4 +296,49 @@ public void updateGroupWatcher(Long userId) throws IOException {
...
+   * Creates a minimal copy of the ProjectOverview to be used for change tracking in the alerting system. Only relevant fields are copied
+   * 
+   * @param projectOverview
+   *          The ProjectOverview to copy
+   * @return the copy
+   */
+  private ProjectOverview partialCopy(ProjectOverview projectOverview) {
+    ProjectOverview clone = new ProjectOverview();
+    clone.setId(projectOverview.getId());
+    clone.setProject(projectOverview.getProject());
```
djcooke: Can the partialCopied project be used here instead? It shouldn't affect functionality, but it would prevent the full project (including samples, libraries, runs, etc.) from being kept in memory unnecessarily

djcooke:
Over 200 compiler warnings fixed! :100: 
:+1:

PR #301, opened 2016-06-22
Glt 787 project shortname

apmasell:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractProject.java
```diff
@@ -133,6 +137,19 @@ public String getAlias() {
   }
 
   @Override
+  public String getShortName() {
+    if (isStringBlankOrNull(shortName)) {
+      return alias;
```
djcooke: This doesn't work since it's for naming and the alias (full project name) is not likely to pass validation when used for naming

miso-web/src/main/webapp/scripts/project_ajax.js
```diff
@@ -35,6 +35,12 @@ var Project = Project || {
     jQuery('#alias').attr('data-parsley-maxlength', '100');
     jQuery('#alias').attr('data-parsley-pattern', Utils.validation.sanitizeRegex);
 
+    // Short name input field validation
+    jQuery('#shortName').attr('class', 'form-control');
+    jQuery('#shortName').attr('data-parsley-required', 'true');
+    jQuery('#shortName').attr('data-parsley-maxlength', '100');
```
djcooke: Should be limited to 5 characters, or will fail sample alias validation when used by generator

miso-web/src/main/webapp/scripts/lims.js
```diff
@@ -232,7 +232,8 @@ Utils.fileUpload = {
 Utils.validation = {
   dateRegex: '^(0[1-9]|[12][0-9]|3[01])\/(0[1-9]|1[012])\/(19|20)[0-9]{2}$',
   dateTimeRegex: '^(0[1-9]|[12][0-9]|3[01])\/(0[1-9]|1[012])\/(19|20)[0-9]{2} ([01][0-9]|2[0-3]):[0-5][0-9]$',
-  sanitizeRegex: '[^<>\'\"\/&]+',
+  sanitizeRegex: '^[^<>\'\"\/&]+$',
+  alnum: '^[A-Za-z0-9]+$',
```
djcooke: for shortName, according to OicrSampleNamingScheme, CAPS/numbers only

sqlstore/src/main/resources/db/migration/V0030__project_shortName.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE Project ADD COLUMN shortName varchar(20) DEFAULT NULL;
```
djcooke: Should be varchar(5). Can we default current project shortNames to their alias truncated to 5 characters too?

djcooke:
DefaultSampleService.createParentIdentity should use Project shortName instead of alias for generating Sample internalName/alias

sqlstore/src/main/resources/db/migration/V0030__project_shortName.sql
apmasell: Fixed. The string handling in MySQL makes this impractical. I'll do this in Java.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractProject.java
apmasell: Fixed.

miso-web/src/main/webapp/scripts/lims.js
apmasell: Fixed.

miso-web/src/main/webapp/scripts/project_ajax.js
apmasell: Okay. This is not universally true though. :-/

apmasell:
Fixed the alias generator.

sqlstore/src/main/resources/db/migration/V0030__project_shortName.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE Project ADD COLUMN shortName varchar(5) DEFAULT NULL;
```
djcooke: UPDATE Project SET shortName=UPPER(SUBSTRING(alias,1,5));
?

miso-web/src/main/webapp/scripts/project_ajax.js
djcooke: Fixing the Sample validation would also be acceptable if that's what is wrong. The point of this field is to be used in Sample alias generation though, so its constraints should match what it's used for

sqlstore/src/main/resources/db/migration/V0030__project_shortName.sql
apmasell: That could have non-alphanumeric characters in it. There's no `REGEX_REPLACE` for MySQL.

djcooke: It might not be perfect, but having something there that may fail validation is better than null, which will always fail validation.

I guess the solution may also be "when you upgrade to this release, you must define short names for all of your projects"

apmasell: Okay. I think the existing code would handle this, with the getter doing the substitution.

djcooke:
:+1: I'd still rather store something in the database (and the column not be nullable) than have the synthetic short name, which is unclear to users, but in production projects should just have a shortName anyway

tonydebat:
👍 If a user modifies the value of the short name this will cause trouble, but this was true of the alias before that as well. I'll create a separate ticket for this.

PR #302, opened 2016-06-22
misc sample form adjustments

djcooke:
List Samples
* removed Tissue classes from child class options for propagation

Edit Sample (single)
* volume is no longer a required field

Edit Sample (detailed)
* fixed Sample JSON serialisation, so groupId is no longer set to 0 when it should be null
* set STR Status default to 'Not Submitted' instead of null (invalid)
* removed receivedDate field when propagating
* removed hardcoded EST timezone from receivedDate

Misc
* removed some javascript that seemed to be setting default scientificName and sampleClassId, but that was actually doing nothing while the defaults were set elsewhere


apmasell:
:+1:

tonydebat:
👍

PR #305, opened 2016-06-30
Glt 684 nextseq support

apmasell:


tonydebat:
The notification server is processing runs slowly. It takes about 40 seconds to process a single run. In the case of production with 1359 runs, this will take 15 hours to complete.

tonydebat:
Pacbio processing works and completes quickly. ~1m for 144 runs.

tonydebat:
I tested the notification server with the NextSeq 550 in isolation (3 runs) and MISO generated a 500 internal server error. I've attached the logs to this jira issue. https://jira.oicr.on.ca/browse/GLT-799

djcooke:
Caching the Metrix stuff would be nice to speed up loading of that part of the Run pages, but could it be done in a separate thread from the run processing, so it doesn't take hours for the first round of runs to be sent after a notification server reset?

PR #307, opened 2016-07-06
GLT-796 Paginate List Samples

callunity:


callunity:
(I also need to delete the `listSamplesTable` method from the SampleControllerHelperService; will do that with the fix after the first round of comments)

miso-web/src/main/webapp/scripts/sample_ajax.js
```diff
@@ -971,73 +971,135 @@ Sample.ui = {
...
-    jQuery.fn.dataTableExt.oSort['no-sam-asc'] = function (x, y) {
-      var a = parseInt(x.replace(/^SAM/i, ""));
-      var b = parseInt(y.replace(/^SAM/i, ""));
-      return ((a < b) ? -1 : ((a > b) ? 1 : 0));
-    };
-    jQuery.fn.dataTableExt.oSort['no-sam-desc'] = function (x, y) {
-      var a = parseInt(x.replace(/^SAM/i, ""));
-      var b = parseInt(y.replace(/^SAM/i, ""));
-      return ((a < b) ? 1 : ((a > b) ? -1 : 0));
+    jQuery.fn.dataTableExt.oApi.fnSetFilteringDelay = function ( oSettings, iDelay ) {
```
callunity: This should get moved out of this method.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSampleDao.java
```diff
@@ -339,15 +346,39 @@ public boolean aliasExists(String alias) throws IOException {
 
   @Override
   public Collection<Sample> listBySearch(String querystr) throws IOException {
-    Query query = currentSession().createQuery(
-        "from SampleImpl where identificationBarcode like :query or name LIKE :query or alias like :query or description like :query or scientificName like :query");
-    query.setString("query", querystr);
+    Criteria criteria = currentSession().createCriteria(Sample.class);
+    criteria.add(Restrictions.or(Restrictions.ilike("identificationBarcode", querystr + "%"), Restrictions.ilike("name", querystr + "%"),
+        Restrictions.ilike("alias", querystr + "%")));
```
tonydebat: Consider adding a wildcard to the front of the string as well. Right now searching for `KGMT` will match `KGMT_0015`, but searching for `0015` will not. Since the query string is not escaped, it's currently possible to use SQL wildcards in queries (this is not something a general user will do.) So searching for `%0015` will match against `KGMT_0015`.

tonydebat:
👍 The page navigation is consistent, but appears to be backwards to my UI sensibility. (The First should be Last and the Last should be First.) Get some consensus from the team before making this change.

Also it would be nicer if the navigation fit nicely in the grey bar. Or appeared below the grey bar maybe.

image

callunity:
@tonydebat , I realize I thought GSLE LIMS ordered things by `ID DESC` (most recently created are at top of list) but upon looking further at it, they're actually ordered by `lastModified`. I will update the SQL to reflect this. (We've occasionally had requests for viewing order = 'last modified by me', but that could be a later fix).

callunity:
@tonydebat : re. column ordering, it looks like `lastUpdated` date is currently a property on SampleAdditionalInfo, which means it's not currently easy to order by it from Hibernate. I've created a ticket to move `lastUpdated` over to Sample (as `lastModifier` is already on Sample), and that should get completed after all the other changes to the sample object model. I'll leave it as ordering by `ID DESC` for right now, as they're slightly more likely to want to see things by most recently created as opposed to first created.

djcooke:
The lastUpdated field has been removed from all of the Sample add-on tables since the information is available already via changelogs. Instead of re-adding the redundant fields, it would be preferable to add to Sample a getLastModified() method that returns the most recent changeTime. Probably add it as a field on the Dto

apmasell:
I agree with the last modified via change logs plan. The change logs are ordered correctly in the list provided, so you don't have to rummage around.

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/DataTablesResponseDto.java
```diff
@@ -0,0 +1,49 @@
+package uk.ac.bbsrc.tgac.miso.dto;
+
+import java.util.List;
+
+import org.codehaus.jackson.annotate.JsonProperty;
+
+public class DataTablesResponseDto<T> {
```
djcooke: Method-level Javadocs would be nice to clarify what the fields are for without having to look through DataTables' docs (though you could include a link to them in the class-level Javadoc too)

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
```diff
@@ -116,16 +121,46 @@ private static SampleDto writeUrls(SampleDto sampleDto, UriComponentsBuilder uri
 
   @RequestMapping(value = "/samples", method = RequestMethod.GET, produces = { "application/json" })
   @ResponseBody
-  public ResponseEntity<Set<SampleDto>> getSamples(UriComponentsBuilder uriBuilder) throws IOException {
-    Set<Sample> samples = sampleService.getAll();
-    if (samples.isEmpty()) {
-      throw new RestException("No samples found", Status.NOT_FOUND);
-    } else {
-      Set<SampleDto> sampleDtos = Dtos.asSampleDtos(samples);
+  public ResponseEntity<?> getSamples(HttpServletRequest request, HttpServletResponse response, UriComponentsBuilder uriBuilder)
```
djcooke: Should return either `ResponseEntity<DataTablesResponseDto<SampleDto>>` or just `DataTablesResponseDto<SampleDto>`. Probably prefer the latter since the ResponseEntity isn't really doing anything. (`200 OK` is default response status)

miso-web/src/main/webapp/scripts/sample_ajax.js
```diff
@@ -971,73 +971,111 @@ Sample.ui = {
...
+          "mRender": function (data, type, full) {
+            return (data ? data : "Unknown");
+          }
+        },
+        {
+          "sTitle": "QC Result",
+          "mData": "url",
+          "mRender": function (data, type, full) {
+            return "NA";
+          } // TODO: FIX THIS PLACEHOLDER
```
djcooke: Was this intentionally left like this?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
callunity: It's not always DataTables requesting a representation of samples; sometimes it should return a regular `Set<SampleDtos>`, so `ResponseEntity<?>` is appropriate (if unnecessarily verbose, with the status code).

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/DataTablesResponseDto.java
callunity: Included.

djcooke:
:+1:

PR #308, opened 2016-07-07
Readme docker

morgantaschuk:
Added docker information to README and spruced up gh-pages


morgantaschuk:
Very undemocratically merging without review.

PR #310, opened 2016-07-13
Glt 745 refactor model

apmasell:


djcooke:
Create Sample (single) page is only showing Tissue classes in the Sample Class dropdown (should show Tissues and Stocks)

djcooke:
Sample Class dropdown issue is because of out-of-order migrations. I think either the V8001 migration in OICR needs to be modified, or a new (OICE) migration needs added that reapplies the SampleClass category changes

djcooke:
When trying to propagate to sample, the HOT doesn't show up and I see in Javascript console:
```
TypeError: sam.groupId is undefined
Sample.hot.modifySamplesForPropagate/<() sample_hot.js:99
Enumerable.collect/<()  prototype.js:669
Enumerable.each/<() prototype.js:628
forEach() self-hosted:232
Enumerable.each() prototype.js:627
Enumerable.collect() prototype.js:668
Sample.hot.modifySamplesForPropagate() sample_hot.js:79
Sample.hot.makeBulkCreateTable()  5&scid=11:247
Hot.fetchSampleOptions/xhr.onreadystatechange()
```

djcooke:
I manually set the proper SampleClasses' sampleCategoy to 'Stock' and the Create Sample (single) page is fine, but now the Stock classes aren't included in the dropdown on the bulk create tab. The canCreateNew function in sample_hot.js should do
`return sampleClass.sampleCategory === "Tissue" || sampleClass.sampleCategory === "Stock";`

miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -194,15 +207,62 @@ Sample.hot = {
     var classes = Hot.sortByProperty(Hot.sampleOptions.sampleClassesDtos, 'id');
     select.push('<select id="classDropdown">');
     select.push('<option value="">Select class</option>');
-    for (var i=0; i<classes.length; i++) {
-      if (classes[i].alias == "Identity") continue;
-      select.push('<option value="'+ classes[i].id +'">'+ classes[i].alias +'</option>');
+    var classOptions = Sample.hot.getNewSampleClassOptions();
+    for (var i=0; i<classOptions.length; i++) {
+     if (classes[i].alias == "Identity") continue;
```
djcooke: This line should be removed. It's looking at the wrong array (classes) and ends up omitting a sample class that should be included. The array it should be looking at (classOptions) is already filtered

miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -613,172 +648,244 @@ Sample.hot = {
...
+        data: 'timesReceived',
+        type: 'numeric',
+        validator: requiredText
+      },{
+        header: 'Tube Number',
+        data: 'tubeNumber',
+        type: 'numeric',
+        validator: requiredText
+      },{
+        header: 'Cellularity',
```
djcooke: Remove cellularity field

djcooke:
On bulk create sample, after choosing a project and sample class and making the table, Sample Class in the table was being automatically selected before, and it no longer is

djcooke:
Saving from bulk create sample fails and I get this stack trace:
```
2016-07-14 11:39:15 ERROR: 500error handling REST request
org.springframework.http.converter.HttpMessageNotReadableException: Could not read JSON: Unrecognized field "externalName" (Class uk.ac.bbsrc.tgac.miso.dto.SampleDto), not marked as ignorable
 at [Source: org.apache.catalina.connector.CoyoteInputStream@27328a44; line: 1, column: 371] (through reference chain: uk.ac.bbsrc.tgac.miso.dto.SampleDto["externalName"]);
```

callunity:
On a Tissue single sample page, adding text to "Region" field does not get persisted.

callunity:
:+1:

djcooke:
The region field should be nullable, but when trying to save a new tissue, I get `Row 1: Column 'region' cannot be null`. If I enter a region, it gives the same error.

The database column is also 'NOT NULL', but should allow nulls. There must be a problem in the javascript too though if it's not saving even when I enter a region

djcooke:
When trying to save a new stock on the bulk form, I get `Row 1: Parent Identity not permitted to have a child of type gDNA (stock)`, which means the Tissue parent is not getting created. DefaultSampleService looks fine, so most likely the problem is in either HOT or DTO conversion

djcooke:
When creating stocks (on bulk form) and when propagating tissue > stock, STR Status should default to 'Not Submitted'. I had fixed this previously in develop (GLT-779)

miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -953,126 +1067,117 @@ Sample.hot = {
     sample.projectId = (parseInt(obj.projectId) || parseInt(document.getElementById('projectSelect').value));
     sample.scientificName = obj.scientificName;
     if (obj.receivedDate && obj.receivedDate.length) {
-      sample.receivedDate = obj.receivedDate;
+      // the time string is added for detailedSample because the server is expecting a datetime value
+      sample.receivedDate = obj.receivedDate += "T00:00:00-05:00";
```
djcooke: Also in GLT-779, I had removed the hardcoded timezone from here. This should just be `sample.receivedDate = obj.receivedDate;`

djcooke:
Another GLT-779 change: when propagating, 'Date of Receipt' column should be hidden

djcooke:
I'm gone tomorrow and Monday. Please make sure the following cases work before merging:
* create (receive) new tissue via single sample form
* create (receive) new stock via single sample form
* create (receive) new tissue via bulk sample form
* create (receive) new stock via bulk sample form
* propagate tissue > stock
* propagate stock > aliquot
* edit (and save) tissue via single sample form
* edit (and save) stock via single sample form
* edit (and save) aliquot via single sample form
* edit (and save) tissue via bulk sample form
* edit (and save) stock via bulk sample form
* edit (and save) aliquot via bulk sample form

callunity:
Subproject is not persisted on the Single Sample page

callunity:
We went through the list from @djcooke and confirmed that samples are saveable and editable as intended. We have identified a few Handsontable changes that will be made in a later fix.

PR #311, opened 2016-07-14
Show the percentage of each dilution in a pool

apmasell:


spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/PoolControllerHelperService.java
```diff
@@ -745,6 +745,12 @@ public JSONObject getPoolableElementInfo(HttpSession session, JSONObject json) {
         Long elementId = json.getLong("elementId");
         Pool<? extends Poolable> pool = requestManager.getPoolById(poolId);
         StringBuilder info = new StringBuilder();
+        double total = 0;
+        for (Poolable p : pool.getPoolableElements()) {
+          if (p instanceof Dilution) {
+            total += ((Dilution) p).getConcentration();
```
callunity: The total concentration depends on volume as well (otherwise you end up with 100 ng in one uL), which doesn't get captured by this equation :)

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/PoolControllerHelperService.java
```diff
@@ -768,6 +774,7 @@ public JSONObject getPoolableElementInfo(HttpSession session, JSONObject json) {
                   + dilution.getLibrary().getAlias() + "(" + dilution.getLibrary().getName() + ")</a><br/>");
               info.append("<b>Sample:</b> <a href='/miso/sample/" + dilution.getLibrary().getSample().getId() + "'>"
                   + dilution.getLibrary().getSample().getAlias() + "(" + dilution.getLibrary().getSample().getName() + ")</a><br/>");
+              info.append("<b>Percentage:</b> " + Math.round(dilution.getConcentration() * 100 / total) + "%<br/>");
```
callunity: I don't think that the lab treats percent pool content like this -- whenever I've seen them discuss it, it's a "Please make this pool 25% Dilution A, 25% Dilution B, and 50% Dilution C." In this case, it has to be user-entered on a per-pool basis, not calculated.

PR #312, opened 2016-07-15
Update column filtering

apmasell:


miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -587,338 +587,290 @@ Sample.hot = {
...
+        include: detailedBool && idColBoolean
+      },
+      {
+        header: 'Tissue Class',
+        data: 'parentSampleClassAlias',
+        type: 'dropdown',
+        trimDropdown: false,
+        source: Sample.hot.getTissueClassesAliasOnly(),
+        validator: validateTissueClasses,
+        include: detailedBool && idColBoolean && (sampleCategory === 'Aliquot' || sampleCategory === 'Stock')
```
callunity: Shouldn't be necessary on aliquots

callunity:
:+1: logic looks good

apmasell:
Okay. This is ready for real review. ^_^

miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -640,45 +656,87 @@ Sample.hot = {
...
       {
-        header: 'External Name',
-        data: 'externalName',
-        type: 'text',
-        validator: requiredText,
-        include: detailedBool && idColBoolean
-      },{
-        header: 'Sex',
-        data: 'donorSex',
+        header: 'Kit',
```
callunity: Can you confirm that it's the Aliquots which have the kits applied to them, and not the Stocks? I would have thought that an extraction kit (chemical change) happens at the stock level, not the aliquot level.

miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -578,347 +563,339 @@ Sample.hot = {
...
+        type: 'numeric',
+        format: '0.00',
+        include: showQcs || show['Stock']
+      },
+      {
+        header: 'Conc.',
+        data: 'concentration',
+        type: 'numeric',
+        format: '0.00',
+        include: showQcs || show['Stock']
```
djcooke: Concentration is on stock only

djcooke:
When creating a Tissue, both 'Sample Class' and 'Tissue Class' columns are shown, and required. Setting them both results in:

`Row 1: Parent Primary Tumor Tissue not permitted to have a child of type Primary Tumor Tissue`

The Tissue Class column was meant to be shown only when receiving Stocks in order to determine the parent (ghost) tissue's SampleClass. TissueClass should be null if SampleClass is a tissue class

djcooke:
When propagating tissue > stock, STR Status 'Not Submitted' should be selected by default. Not sure if this was broken here or during merge of sample model refactor

apmasell:
No, it was working after the merge.

apmasell:
Okay. It wasn't working after the merge, because it never worked for propagate.

djcooke:
Ah, a test case I missed in my own fix then - my bad!

:+1:

PR #313, opened 2016-07-18
Avoid duplicate keys in the sequencer partition container DAO

apmasell:


callunity:
:+1:

djcooke:
:+1:

:100: for reducing ugly JdbcTemplate code. Hibernate will be a better solution one day...

callunity:
Ooh. Yeah, that's much nicer.

PR #314, opened 2016-07-18
GLT-556 update List Samples page

callunity:
Fixes for a few bugs:
- non-detailed sample options (propagate samples; empty; archive) are not shown if detailed sample is not enabled
- multiple samples can only be edited if they are all of the same sample category
- Tissue Processing samples can only be edited if their sample class-specific columns do not collide (ie. can't do (LCM Tubes || CV Slides) + anything else
- Propagating samles: checks that a sample class is selected before continuing.

### DIFF
--- a/miso-web/src/main/webapp/scripts/sample_ajax.js
+++ b/miso-web/src/main/webapp/scripts/sample_ajax.js
@@ -1005,14 +1005,17 @@ Sample.ui = {
   },
 
   createListingSamplesTable: function () {
-    if (Sample.detailedSample) Sample.ui.getSampleClasses();
+    if (Sample.detailedSample && Sample.sampleClasses === undefined) {
+      Sample.ui.getSampleClasses();
+      Sample.ui.getSampleValidRelationships();
+    }
     jQuery('#listingSamplesTable').html("<img src='../styles/images/ajax-loader.gif'/>");
 
     jQuery('#listingSamplesTable').html('');
     jQuery('#listingSamplesTable').dataTable({
       "aoColumns": [
         {
-          "sTitle": "Bulk Edit",
+          "sTitle": "",
           "mData": "id",
           "mRender": function (data, type, full) {
             return "<input type=\"checkbox\" value=\"" + data + "\" class=\"bulkCheckbox\" id=\"bulk_" + data + "\">"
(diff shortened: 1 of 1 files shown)


miso-web/src/main/webapp/scripts/sample_ajax.js
```diff
@@ -1243,32 +1274,138 @@ Sample.ui = {
...
+   * Returns an array of sample classes associated with selected samples
+   */
+  getSampleClassesForSelected: function () {
+    var selectedIds = Sample.ui.getSelectedIds();
+    
+    // get unique sample class IDs
+    var selectedSampleClasses = jQuery('#listingSamplesTable').dataTable().fnGetData().filter(function (row) {
+      return selectedIds.indexOf(row.id) != -1;
+    }).map(function (dtRow) {
+      return Sample.sampleClasses.filter(function (sc) {
```
apmasell: Shouldn't this return exactly one item? If it returned that, there'd be no need for the `concat`.

apmasell:
:+1:

miso-web/src/main/webapp/scripts/sample_ajax.js
callunity: Since the filter is within the map, it returns an array of one item (filter) inside the larger array (map).

apmasell: Right, so you could `filter()[0]` and lose the inner arrays.

callunity: Ah, I see what you mean. No, the initial filter might return multiple sample classes, as Primary Tumor Tissue and Reference Tissue have the same columns and should be able to be edited/propagated together.

miso-web/src/main/webapp/scripts/sample_ajax.js
```diff
@@ -1166,49 +1190,60 @@ Sample.ui = {
...
-    var selectedIdsArray = Sample.ui.getSelectedIds();
-    if (selectedIdsArray.length === 0) {
-      alert("Please select one or more Samples to propagate.");
-      return false;
+    jQuery('#go').attr('disabled', 'disabled');
+    if (Sample.detailedSample) {
+      // rejects propagation if samples are not of aliquot category
+      // TODO: add a more detailed check, or library designs?
+      if (Sample.ui.getUniqueCategoriesForSelected() !== 'Aliquot') {
+        jQuery('#errors').html("Tissue, Tissue Processing, and Stock samples cannot be propagated to libraries.");
```
djcooke: Nor can Identities

djcooke:
:+1:

miso-web/src/main/webapp/scripts/sample_ajax.js
callunity: ...yet :wink:

PR #315, opened 2016-07-19
Track synthetic (ghost entities) in the database

apmasell:


sqlstore/src/main/resources/db/migration/V0028__model_refactor.sql
```diff
@@ -85,9 +85,45 @@ DROP TABLE SampleAnalyte;
...
+  OPEN badChildren;
+
+  read_loop: LOOP
+    FETCH badChildren INTO badId;
+    IF done THEN
+      LEAVE read_loop;
+    END IF;
+    INSERT INTO Sample(name, description, securityProfile_profileId, sampleType, receivedDate, project_projectId, scientificName, taxonIdentifier, lastModifier, emptied) SELECT 'TEMP', description, securityProfile_profileId, sampleType, receivedDate, project_projectId, scientificName, taxonIdentifier, lastModifier, emptied FROM Sample WHERE sampleId = badId;
+    SELECT LAST_INSERT_ID() INTO newId;
+    UPDATE Sample SET name = CONCAT('SAM', newId);
```
djcooke: WHERE sampleId = newId

sqlstore/src/main/resources/db/migration/V0028__model_refactor.sql
```diff
@@ -85,9 +85,45 @@ DROP TABLE SampleAnalyte;
...
+  DECLARE CONTINUE HANDLER FOR NOT FOUND SET done = TRUE;
+
+  OPEN badChildren;
+
+  read_loop: LOOP
+    FETCH badChildren INTO badId;
+    IF done THEN
+      LEAVE read_loop;
+    END IF;
+    INSERT INTO Sample(name, description, securityProfile_profileId, sampleType, receivedDate, project_projectId, scientificName, taxonIdentifier, lastModifier, emptied) SELECT 'TEMP', description, securityProfile_profileId, sampleType, receivedDate, project_projectId, scientificName, taxonIdentifier, lastModifier, emptied FROM Sample WHERE sampleId = badId;
```
djcooke: Alias never gets set. While there isn't a great way of doing this without the naming scheme, maybe it could be set to something similar to the name field so it is at least not null

callunity:
:+1:

djcooke:
:+1:

PR #316, opened 2016-07-19
Glt 785 migration (and related) updates

djcooke:


migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -102,110 +153,128 @@ private void saveSample(Sample sample) throws IOException {
...
-          library.setLastUpdated(now);
-          library.getLibraryAdditionalInfo().setCreatedBy(user);
-          library.getLibraryAdditionalInfo().setCreationDate(now);
-          library.getLibraryAdditionalInfo().setUpdatedBy(user);
-          library.getLibraryAdditionalInfo().setLastUpdated(now);
-          library.setId(serviceManager.getLibraryDao().save(library));
-          log.debug("Saved library " + library.getAlias());
-        }
-        return null;
+    Date now = new Date();
```
apmasell: Should this be set for the whole migration?

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/ValueTypeLookup.java
```diff
@@ -0,0 +1,545 @@
...
+      }
+      mapByName.put(kit.getName(), kit);
+      mapById.put(kit.getId(), kit);
+    }
+    this.kitById = mapById;
+    this.kitByName = mapByName;
+  }
+  
+  private void setSamplePurposes(Collection<SamplePurpose> samplePurposes) {
+    Map<Long, SamplePurpose> mapById = new HashMap<>();
```
apmasell: Is it worth creating an implementation of map that throws a duplicate key on `put` to avoid doing this so many times?

apmasell:
:+1:

callunity:
:+1:

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
djcooke: Yes. Will set a timestamp in the constructor and use that everywhere instead

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/ValueTypeLookup.java
djcooke: Sure, I like that

apmasell:
:+1:

callunity:
👍

PR #321, opened 2016-07-25
GLT-815 Paginate List Libraries page

callunity:


apmasell:
:+1:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractLibrary.java
```diff
@@ -131,7 +131,7 @@
   @JoinColumn(name = "lastModifier", nullable = false)
   private User lastModifier;
 
-  @Formula(value = "(SELECT MAX(lcl.changeTime) FROM LibraryChangeLog lcl WHERE lcl.libraryId = libraryId)")
+  @Formula(value = "(SELECT MAX(lmod.changeTime) FROM LibraryChangeLog lmod WHERE lmod.libraryId = libraryId)")
```
djcooke: I believe this currently has no effect since Libraries aren't loaded via Hibernate currently. When they are, the same pattern as is used for Samples will perform better. Is there any reason to keep the annotation here?

callunity: I kept this line because when I tried the Sample pattern, the JDBC Library was unhappy, and when I tried without the `@Formula` annotation, the Hibernate LibraryAdditionalInfo was unhappy.

callunity: (Also, the Hibernate-no-longer-loading-Libraries change was merged after I opened the PR ;) but I'll update this to remove the annotation.)

djcooke:
:+1:

PR #324, opened 2016-07-26
Fix seqencing partition container change log id

apmasell:


callunity:
Hah - I've got this change in my next PR, so I know it works. :+1:

djcooke:
:+1:

PR #325, opened 2016-07-26
GLT-818 Paginate List Partition Containers page

callunity:
(depends on changes in #324)

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/UserAuthMisoRequestManager.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/UserAuthMisoRequestManager.java
@@ -2615,4 +2615,51 @@ public class UserAuthMisoRequestManager implements RequestManager {
   public Long countRunsBySearch(String querystr) throws IOException {
     return backingManager.countRunsBySearch(querystr);
   }
+
+  @Override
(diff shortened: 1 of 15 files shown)


apmasell:
:+1:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/UserAuthMisoRequestManager.java
```diff
@@ -2615,4 +2615,51 @@ public Long countRuns() throws IOException {
...
+        .getContainersByPageAndSize(offset, limit, sortDir, sortCol)) {
+      if (spc.userCanRead(user)) {
+        accessibles.add(spc);
+      }
+    }
+    return accessibles;
+  }
+
+  @Override
+  public Long countContainersBySearch(String querystr) throws IOException {
```
djcooke: Since this doesn't filter based on permissions, it may not match the number actually returned by the search query. I'm not sure what the solution is since looking up all of the containers would be slow

callunity: Let's leave it as is, as I think most of our internal users have access to most items. We can re-examine it if it ends up glitching due to users lacking permissions on a large number of items.

PR #326, opened 2016-07-26
Prevent null aliases from being generated

apmasell:


callunity:
:+1:

djcooke:
Null alias will break downstream alias generation. Why are null aliases necessary?

apmasell:
They aren't necessary, but they are being (not) generated for me when I create a first-receipt stock.

djcooke:
Alias should be generated properly with this naming scheme config in `miso.properties`:
```
## configs for naming validation
miso.naming.scheme.nameable:DefaultEntityNamingScheme
miso.naming.scheme.sample:OicrSampleNamingScheme
miso.naming.scheme.library:OicrLibraryNamingScheme

## configs for alias generation
miso.naming.generator.sample.alias:OicrSampleAliasGenerator
```

I think we probably shouldn't allow null alias because it will cause more problems, but we should have a way to generate a ghost tissue alias in the absence of a configured sample alias generator

apmasell:
I agree. How do you propose going about that?

djcooke:
Just copy the name (e.g. `SAM123`) when generating name+alias in DefaultSampleService? If aliases aren't being generated, and the tissue will be hidden, its alias doesn't really matter. It would be good to keep it constrained to being unique and non-null though

apmasell:
Okay. Name copied.

djcooke:
Could do this in the pre-save name generation part of the `create(Sample)` method instead to avoid mention in the second changelog entry, but it doesn't really matter. And Sample.alias is nullable in the DB... :crying_cat_face: 

:+1:

apmasell:
Yeah, if the name is a temporary name I don't want to copy that to the alias though, so it has to happen after. :-(

callunity:
:+1:  if it's a ghost entity, we won't have to care about the additional changelog entry, since it won't ever be seen anyway.

djcooke:
@apmasell Oh yeah. duh!

PR #330, opened 2016-07-29
Add a script to compact migrations as part of the release process

apmasell:


compact-migrations
```diff
@@ -0,0 +1,53 @@
...
+set -eu
+
+cd "$(dirname $0)/sqlstore/src/main/resources/db/migration"
+
+# Determine the last migration in the working directory
+LAST="$(basename "$(ls V[0-5][0-9][0-9][0-9]_*.sql | sort | tail -n 1)")"
+echo "Previous migration is ${LAST}."
+LAST_ID="$(echo "${LAST}" | cut -c 2-5 | sed -e 's/^0*//')"
+
+# Determine if we are on the mainline branch or a sit-specific branch.
```
callunity: Nitpick: "site-specific"

sqlstore/src/main/resources/db/migration/readme.md
```diff
@@ -4,10 +4,15 @@ Flyway Migrations
...
 
 Additionally, the following rules apply:
 - neither standard nor site-specific migrations should never DROP and then re-create a data table since this table might be already modified by a site-specific migration (schema or data)
 - if new columns are added, they must be nullable or have defaults; UPDATEing is not sufficient since there might be site-specific data
 - site-specific migrations are responsible for creating column and table names that do not conflict with the main schema.
+
+During each release cycle, all the 8xxx migrations will be compacted into a
+single migration by the `compact-migrations` script. This is also true for
+site-specific migrations during releases. The scripts detects what the correct
+mode is by determining if the branch is `origin/develop`.
```
callunity: Worth specifying that it's checking if the branch is TGAC's `origin/develop` ?

callunity:
:+1:

tonydebat:
This will work while `develop` and `site-specific` are released in lock step. But for a mature (or new) deployment the `site-specific` branch could be updated and released multiple times in order to update the site configuration and this could cause out of order migration scripts.

apmasell:
The empty file is intentional to enforce that a site-specific change will always have a mainline/develop change to release on top of. I have no idea how to handle multiple site-specific releases in between mainline. I could change the address space so that there can be _k_ releases site-specific releases between mainline releases (something like mainline is always xxx0 and site-specific are xxx[1-9]). Is 10 a good number?

tonydebat:
👍I like the number 10, it has that BASIC feel about it.

callunity:
:+1: I like 10 as well.

PR #332, opened 2016-08-03
Issue a warning when a sample is synthetic

apmasell:


callunity:
This is not very prominent, to my eyes. I'd like to see it much larger, perhaps in a different colour.

apmasell:
Here's something a little more prominent. I call it eye-rend chic.

callunity:
:+1:

tonydebat:
👍

PR #335, opened 2016-08-08
GLT-275 update List Pools page

callunity:
- hides/shows pooled elements
  - shows tagbarcodes on libraries
- adds a `description` field to Pool

### DIFF
--- a/sqlstore/src/main/resources/db/migration/V0035__add_description_to_pool.sql
+++ b/sqlstore/src/main/resources/db/migration/V0035__add_description_to_pool.sql
@@ -1,0 +1,2 @@
+ALTER TABLE Pool ADD COLUMN description varchar(255) DEFAULT NULL;
+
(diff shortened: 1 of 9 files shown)


apmasell:
:+1: Glad to see this ticket fixed.

sqlstore/src/main/resources/db/migration/V0035__add_description_to_pool.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE Pool ADD COLUMN description varchar(255) DEFAULT 'Pool';
```
djcooke: The field is nullable. Any reason for the default of 'Pool' instead of NULL?

djcooke:
:+1:

sqlstore/src/main/resources/db/migration/V0035__add_description_to_pool.sql
callunity: Nope, must be vestigial from when I was considering making it not nullable. I'll update that.

PR #336, opened 2016-08-09
GLT-994 improve UX for Handsontable bulk entry

callunity:
- no longer jitters and freezes when filling down rows by double-clicking cell corner
  - incidentally, supports multi-column fill-down
- no longer adds new rows when user drags down


apmasell:
:+1:

djcooke:
:+1:

PR #339, opened 2016-08-09
Prometheus monitoring

apmasell:
Exports per URL stats and per DAO method stats via `/metrics` for Prometheus to scrape.

### DIFF
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/context/PrometheusCollectionFilter.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/context/PrometheusCollectionFilter.java
@@ -1,0 +1,248 @@
+package uk.ac.bbsrc.tgac.miso.webapp.context;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.util.Locale;
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.ServletOutputStream;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.springframework.web.filter.GenericFilterBean;
+
+import io.prometheus.client.Counter;
+import io.prometheus.client.Histogram;
+import uk.ac.bbsrc.tgac.miso.core.util.CoverageIgnore;
+
+/**
+ * Collect per-URL performance statistics and export them to Prometheus
+ */
+public class PrometheusCollectionFilter extends GenericFilterBean {
+
+  // Wrapper for the servlet response in order to collect the status code
+  private static class ServletResponseWrapper implements HttpServletResponse {
+    private final HttpServletResponse backingResponse;
+    private int status = SC_OK;
+
+    public ServletResponseWrapper(HttpServletResponse httpResponse) {
+      backingResponse = httpResponse;
+    }
+
+    @Override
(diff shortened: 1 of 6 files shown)


callunity:
:+1:

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/context/PrometheusCollectionFilter.java
```diff
@@ -0,0 +1,246 @@
...
+import org.springframework.web.filter.GenericFilterBean;
+
+import io.prometheus.client.Counter;
+import io.prometheus.client.Histogram;
+import uk.ac.bbsrc.tgac.miso.core.util.CoverageIgnore;
+
+/**
+ * Collect per-URL performance statistics and export them to Prometheus
+ */
+@SuppressWarnings("deprecation")
```
djcooke: What's deprecated, and is there not a way to avoid using it?

apmasell: A bunch of methods and types in the `HttpServletResponse` are deprecated, but I can't avoid them to use the interface. If I put the suppression on just that class, it still whines about the imports.

miso-web/src/main/webapp/WEB-INF/event-config.xml
```diff
@@ -28,6 +28,17 @@
...
   default-autowire="byName">
 
+  <bean name="prometheusInterceptor" class="uk.ac.bbsrc.tgac.miso.webapp.context.PrometheusInterceptor" />
+   <bean id="prometheusAdvisor"
+      class="org.springframework.aop.support.RegexpMethodPointcutAdvisor">
+      <property name="advice">
+         <ref local="prometheusInterceptor"/>
+      </property>
+      <property name="pattern">
+         <value>uk\.ac\.bbsrc\.tgac\.miso\.(sqlstore\.SQL|persistence\.impl\.Hibernate).*</value>
```
djcooke: Can the SQL DAOs not be included in here, rather than in db-config.xml?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/context/PrometheusCollectionFilter.java
djcooke: I see 3 deprecated methods used - encodeUrl(String), encodeRedirectUrl(String), and setStatus(int, String).

Looking at the HttpServletResponse javadoc: https://javaee-spec.java.net/nonav/javadocs/javax/servlet/http/HttpServletResponse.html

encodeUrl and encodeRedirectUrl can be changed to encodeURL and encodeRedirectURL, which do the same thing (deprecated to change method names...)

setStatus:

> [deprecated] due to ambiguous meaning of the message parameter. To set a status code use setStatus(int), to send an error with a description use sendError(int, String). Sets the status code and message for this response.

Since you're just forwarding the call to the wrapped response object, changing this may not be reasonable, but the SuppressWarnings annotation could be moved to this method alone instead of the entire class

miso-web/src/main/webapp/WEB-INF/event-config.xml
apmasell: Yes! I had it in before when it wasn't working, but it no longer seems necessary.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/context/PrometheusCollectionFilter.java
apmasell: Fixed.

PR #344, opened 2016-08-10
Fix incorrect library URL and confusing name

apmasell:
Get the correct library id in the URL. Also, rename the collection so the
function doesn't have a variable with the same name in it.


callunity:
:+1:

djcooke:
:+1:

PR #352, opened 2016-08-12
GLT-479 Added configuration for Active Directory authentication

tonydebat:
Active directory does not return a password. Provide a fake password in this case
it is a requirement of the User service.
Created a separate Active directory configuration file that overrode one bean from the
general ldap configuration file.

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/security/ad/ActiveDirectoryAuthoritiesMapper.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/security/ad/ActiveDirectoryAuthoritiesMapper.java
@@ -1,0 +1,39 @@
+package uk.ac.bbsrc.tgac.miso.core.security.ad;
+
+import java.util.Collection;
+import java.util.EnumSet;
+import java.util.Set;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.authority.mapping.GrantedAuthoritiesMapper;
+
+import uk.ac.bbsrc.tgac.miso.core.security.MisoAuthority;
+
+/**
+ * When using active directory some roles may have a prefix required by the AD administrator. The class will map these prefixed roles to the
+ * actual MISO roles.
+ */
+public class ActiveDirectoryAuthoritiesMapper implements GrantedAuthoritiesMapper {
+
+  @Value("${security.ad.stripRolePrefix}")
+  private String rolePrefix = "";
+
+  @Override
+  public Collection<? extends GrantedAuthority> mapAuthorities(Collection<? extends GrantedAuthority> authorities) {
+    Set<MisoAuthority> roles = EnumSet.noneOf(MisoAuthority.class);
+
+    for (GrantedAuthority a : authorities) {
+      if ((rolePrefix + MisoAuthority.ROLE_ADMIN).equals(a.getAuthority())) {
+        roles.add(MisoAuthority.ROLE_ADMIN);
+      } else if ((rolePrefix + MisoAuthority.ROLE_INTERNAL).equals(a.getAuthority())) {
+        roles.add(MisoAuthority.ROLE_INTERNAL);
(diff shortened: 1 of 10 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/security/util/LimsSecurityUtils.java
```diff
@@ -72,24 +81,38 @@ public static User fromLdapUser(InetOrgPerson details) {
...
-
     final List<String> roles = new ArrayList<String>();
     for (final GrantedAuthority ga : details.getAuthorities()) {
-      roles.add(ga.toString());
+      roles.add(removePrefix(ga.toString(), LimsSecurityUtils.rolePrefix));
     }
     user.setRoles(roles.toArray(new String[0]));
 
+    user.setActive(details.isAccountNonExpired());
+    user.setAdmin(roles.contains(new GrantedAuthorityImpl("ROLE_ADMIN")));
```
apmasell: Can these `GrantedAuthorityImpl`s get rolled up into constants?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/security/util/LimsSecurityUtils.java
```diff
@@ -72,24 +81,38 @@ public static User fromLdapUser(InetOrgPerson details) {
...
+    user.setExternal(roles.contains(new GrantedAuthorityImpl("ROLE_EXTERNAL")));
+    user.setInternal(roles.contains(new GrantedAuthorityImpl("ROLE_INTERNAL")));
+    user.setLoginName(details.getUsername().toLowerCase());
+    if (details.getPassword() == null) {
+      log.debug("This directory has been configured correctly and has properly returned a null password. Setting fake password "
```
apmasell: So, the database is set up to allow null passwords, but the SHA encoder doesn't do a null check. I'm wondering if it's better to fix that code (or use the passthrough encoder, which does no checks). Or did you see other errors?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/security/util/LimsSecurityUtils.java
tonydebat: I've made this change. Unfortunately I forgot about making it a separate commit so that the changes would be obvious.

apmasell:
:+1:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/security/ad/ActiveDirectoryAuthoritiesMapper.java
```diff
@@ -0,0 +1,39 @@
...
+
+  @Value("${security.ad.stripRolePrefix}")
+  private String rolePrefix = "";
+
+  @Override
+  public Collection<? extends GrantedAuthority> mapAuthorities(Collection<? extends GrantedAuthority> authorities) {
+    Set<MisoAuthority> roles = EnumSet.noneOf(MisoAuthority.class);
+
+    for (GrantedAuthority a : authorities) {
+      if ((rolePrefix + "ROLE_ADMIN").equals(a.getAuthority())) {
```
djcooke: Could use MisoAuthority values instead of String role names

djcooke:
:+1:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/security/util/LimsSecurityUtils.java
tonydebat: Null password is now permitted when Active Directory is used.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/security/ad/ActiveDirectoryAuthoritiesMapper.java
tonydebat: Done.

PR #353, opened 2016-08-12
GLT-1011 Handsontable correct sources after autofill

callunity:
Still somewhat slow after Platform Name autofill, but at least it all works.


apmasell:
:+1:

tonydebat:
👍

PR #354, opened 2016-08-15
skip library alias validation during migration

djcooke:
During migration, library alias validation must be skipped, and names such 'LIB123' must be generated. It was not possible to configure the existing naming schemes to achieve this, so I created a very simple one to use for migration instead


apmasell:
:+1:

callunity:
:+1:

PR #356, opened 2016-08-16
Switch to RollingFileAppender and reduce ehcache log level

apmasell:
While date-based searching is nice, there's not automatic way to expire old
logs and log files can grow to an unbounded size. In particular, production
`ehcache.log` files can be large (over 1G) and tend to consume more prod disk
than they are worth. This will limit each log to 500GB and automaticall remove
old logs.


callunity:
:+1:

tonydebat:
👍

PR #359, opened 2016-08-16
GLT-658 bulk edit libraries (Handsontable)

callunity:


miso-web/src/main/webapp/scripts/library_hot.js
```diff
@@ -476,108 +493,112 @@ Library.hot = {
...
-            function (bc2Data) {
-              Library
-              Hot.dropdownRef.tagBarcodes[strat]['2'] = bc2Data.tagBarcodes;
+  getBarcodePositionsOnly: function (tbFamily) {
+    if (tbFamily) {
+      jQuery.get("../../barcodePositionsJson", {tagBarcodeFamily : tbFamily},
+        function(posData) {
+          // set tagBarcodeData
+          Hot.dropdownRef.tagBarcodes[tbFamily]['1'] = {};
+          if (posData['numApplicableBarcodes'] == 2) {
```
apmasell: Consider rolling this into a loop: `for (var pos = 1; pos <= posData['numApplicableBarcodes']; pos++)`

miso-web/src/main/webapp/scripts/library_hot.js
```diff
@@ -476,108 +493,112 @@ Library.hot = {
...
+        } else {
+          lib.libraryAdditionalInfo.archived = false;
+        }
+      }
+
+      lib.qcPassed = obj.qcPassed;
+
+      // TODO: add qcCols
+    } catch (e) {
+      console.log(e.error);
```
apmasell: Shoudl this still return the partially constructed library? Or should it return null or something else?

apmasell:
:+1:

miso-web/src/main/webapp/scripts/library_hot.js
callunity: Are you thinking of doing that for the sake of the user? To my mind, this try/catch block is really only useful during development; the only time things fail within it is if we've made an error in the logic (ie. forgotten to null check something before calling one of its attributes). If that happens, it won't be obvious to the user why it happened, and nothing they do will fix it. Returning a partially-constructed or null library will cause the function to try to save that, which will definitely break.

apmasell: So long as you will be sure it will definitely break if everything isn't processed. I'm generally a fan of returning something definitely broken in this case in case there's some error in the future. I wouldn't want the user to see a successful save that's eaten half their data.

callunity: Fair enough. I will change it to return null as there is zero downside to that.

tonydebat:
👍

PR #363, opened 2016-08-19
Allow customising the “file a ticket” URL

apmasell:


miso-web/src/main/resources/internal/miso.properties
```diff
@@ -27,6 +27,9 @@ miso.baseDirectory:/storage/miso/
 miso.fileStorageDirectory:/storage/miso/files/
 miso.submissionStorageDirectory:/storage/miso/files/submission/
 
+# Where to send bug reports for your organisation
+miso.bugUrl:https://github.com/TGAC/miso-lims/issues/new
```
callunity: Shouldn't this be in `/external`?

apmasell: Yes.

callunity:
:+1: after adding `bugUrl` to external build properties file as well

miso-web/src/main/resources/internal/miso.properties
callunity: Might be worth keeping it in both? I'm not sure what distinctions EI makes between internal and external, or why internal exists since all the documentation specifies external.

djcooke:
:+1:

tonydebat:
👍

PR #364, opened 2016-08-19
resolve pool and run foreign keys, allow partial run migrations

djcooke:


apmasell:
:+1:

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -249,6 +264,34 @@ public void saveRuns(final Collection<Run> runs) throws IOException {
...
   
+  private void updateRun(Run from, Run to) throws IOException {
+    log.debug("Updating run " + to.getId());
+    to.getStatus().setCompletionDate(to.getStatus().getCompletionDate());
+    to.getStatus().setHealth(to.getStatus().getHealth());
+    if (to.getSequencerPartitionContainers().size() != 1) {
+      throw new IOException("Existing run " + to.getName() + " has unexpected number of sequencerPartitionContainers");
+    }
+    if (from.getSequencerPartitionContainers().size() != 1) {
+      throw new IOException("Migrating run " + from.getName() + " has unexpected number of sequencerPartitionContainers");
```
tonydebat: The 454 machine has multiple flowcells. We have a few runs from these, but there's so old I'm pretty sure we won't need to migrate them. Hmmm... actually we have exactly one run.

tonydebat:
👍

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
djcooke: Will deal with that when/if needed

PR #366, opened 2016-08-24
GLT-1044 allow only numbers for timesReceived & tubeNumber

callunity:


miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -987,6 +987,14 @@ Sample.hot = {
       }
     }
     
+    function validatePosReqdNumber (value, callback) {
+      if (Handsontable.helper.isNumeric(value) && value > 0) {
```
apmasell: This can also be:
`callback(Handsontable.helper.isNumeric(value) && value > 0);`

apmasell:
:+1:

miso-web/src/main/webapp/scripts/sample_hot.js
callunity: I'll implement this in a PR coming soon!

djcooke:
:+1:

PR #368, opened 2016-08-24
GLT-1005 Handle non-standard aliases

callunity:
Legacy malformed aliases will need to be maintained, so they and their children can't be run through the alias validators during save.

### DIFF
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
@@ -798,7 +798,8 @@ public class EditLibraryController {
   public ModelAndView editPropagateSamples(@PathVariable String sampleIds, ModelMap model) throws IOException {
     try {
       List<Long> idList = getIdsFromString(sampleIds);
-      JSONArray libraries = new JSONArray();
+      ObjectMapper mapper = new ObjectMapper();
+      List<LibraryDto> libraryDtos = new ArrayList<LibraryDto>();
       SampleClass sampleClass = null;
       for (Sample sample : requestManager.getSamplesByIdList(idList)) {
         SampleAdditionalInfo detailed = (SampleAdditionalInfo) sample;
@@ -813,12 +814,13 @@ public class EditLibraryController {
 
         if (isDetailedSampleEnabled()) {
           LibraryAdditionalInfoDto lai = new LibraryAdditionalInfoDto();
+          lai.setNonStandardAlias(((SampleAdditionalInfo) sample).hasNonStandardAlias());
           library.setLibraryAdditionalInfo(lai);
         }
-        libraries.add(library);
+        libraryDtos.add(library);
       }
       model.put("title", "Bulk Create Libraries");
-      model.put("librariesJSON", libraries);
+      model.put("librariesJSON", mapper.writeValueAsString(libraryDtos));
       JSONArray libraryDesigns = new JSONArray();
       libraryDesigns.addAll(requestManager.listLibraryDesignByClass(sampleClass));
       model.put("libraryDesignsJSON", libraryDesigns.toString());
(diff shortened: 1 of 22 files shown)


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
```diff
@@ -757,18 +755,14 @@ public ModelAndView setupForm(@PathVariable Long sampleId, @PathVariable Long pr
...
   public ModelAndView editBulkSamples(@PathVariable String sampleIds, ModelMap model) throws IOException {
     try {
-      String[] split = sampleIds.split(",");
-      List<Long> idList = new ArrayList<Long>();
-      for (int i = 0; i < split.length; i++) {
-        idList.add(Long.parseLong(split[i]));
-      }
+      List<Long> idList = getIdsFromString(sampleIds);
       ObjectMapper mapper = new ObjectMapper();
       JSONArray samplesDtos = new JSONArray();
```
apmasell: Can we do this the other way around and make `sampleDtos` a `List<SampleDto>` and use the Jackson mapper?

miso-web/src/main/webapp/pages/editLibrary.jsp
```diff
@@ -210,6 +210,13 @@
     </c:choose>
 
     <span id="aliasCounter" class="counter"></span>
+    <c:if test="${detailedSample && library.hasNonStandardAlias()}">
+      <ul class="parsley-errors-list filled" id="nonStandardAlias">
+        <li class="parsley-custom-error-message">
+        Double-check thsi alias -- it will be saved even if it is duplicated or malformed!
```
apmasell: “malformed” is very jargon-y. “does not follow the standard”?

miso-service/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,27 @@
+#
```
apmasell: What's this for?

callunity: I think it's something I added to try to get my test suite working again. Will remove.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -813,12 +814,13 @@ public ModelAndView editPropagateSamples(@PathVariable String sampleIds, ModelMa
...
           LibraryAdditionalInfoDto lai = new LibraryAdditionalInfoDto();
+          lai.setNonStandardAlias(((SampleAdditionalInfo) sample).hasNonStandardAlias());
           library.setLibraryAdditionalInfo(lai);
         }
-        libraries.add(library);
+        libraryDtos.add(library);
       }
       model.put("title", "Bulk Create Libraries");
-      model.put("librariesJSON", libraries);
+      model.put("librariesJSON", mapper.writeValueAsString(libraryDtos));
```
djcooke: We should add a static utility method somewhere for writing JSON into the model like this, so we don't end up doing it in several different ways again (not necessarily in this PR)

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLLibraryDAO.java
```diff
@@ -387,7 +387,8 @@ public long save(Library library) throws IOException {
           String name = libraryNamingScheme.generateNameFor("name", library);
           library.setName(name);
           if (libraryNamingScheme.validateField("name", library.getName())
-              && libraryNamingScheme.validateField("alias", library.getAlias())) {
+              && (libraryNamingScheme.validateField("alias", library.getAlias())
+                  || library.getLibraryAdditionalInfo().hasNonStandardAlias())) {
```
djcooke: * Needs null check on LibraryAdditionalInfo
* Same below (L420)
* Above (L380), there is a check for duplicates, which I think should check the nonStandardAlias attribute as well

djcooke:
:+1:

apmasell:
:+1:

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
callunity: I'll make a separate PR for this once I've finished refactoring TagBarcode -> Index

PR #369, opened 2016-08-25
Pool improvements

apmasell:


callunity:
I like the implementation, but I'd like to alert our users higher up the page . Right now the alert is only displayed after save (after it's already added to the pool) and a scrolldown is required to see the alert.

tonydebat:
👍

apmasell:
@callunity, is this a better spot for the warning?

callunity:
:+1:

PR #373, opened 2016-08-25
GLT-1038 set null if attribute string is blank

callunity:
When Samples and Libraries are created through the Edit Sample/Edit Library page, empty fields were saved to the database as empty strings. This fix checks if the value being set is a blank string and if it is, the value is set to null.


apmasell:
:+1:

tonydebat:
👍

PR #375, opened 2016-08-26
GLT-1042 add ctrl-v message to HOT pages

callunity:
Right-click to paste does not work, so make it very obvious to users what they need to do to paste data.


apmasell:
:+1: You could also ⌘-V if so inclined.

tonydebat:
👍 and +1 for including ⌘-V

PR #376, opened 2016-08-26
GLT-662 Containers addition and editing

tonydebat:
This works when creating an individual container and also when manually adding a run.


apmasell:
:+1:

callunity:
When I try to replace one container with another, I get the following message:

> Replace container with existing '251ADXX container?'. Current values will be overridden.
>
> Container Properties
> (Partitions 1-X)
>
> Choose cancel to save current container.

It's not entirely clear to me which container is referred to each time. What do you think of the following?

> Are you sure you wish to replace run's container ${savedContainerName} with container ${newContainerName}? 
>
>This will associate the following items with run ${runAlias (runName)}:
> (Partitions 1-X)
>
> Choose "Cancel" to discard this change and preserve container ${savedContainerName}.

PR #377, opened 2016-08-26
GLT-1053 don't obliterate values on paste

callunity:
Handsontable no longer completely obliterates values after paste. Also, improved TagBarcode validators.


apmasell:
:+1:

tonydebat:
👍

PR #382, opened 2016-09-01
GLT-1041 Improve UI for bulk create samples

callunity:
Fields now have labels and samples can be batch entered without having to refresh the page.


apmasell:
:+1:

tonydebat:
👍

PR #385, opened 2016-09-01
GLT-1064 replace Partition Container on front end

callunity:
Removes all instances of it from the user's view.

### DIFF
--- a/spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/ContainerControllerHelperService.java
+++ b/spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/ContainerControllerHelperService.java
@@ -88,8 +88,9 @@ public class ContainerControllerHelperService {
   public JSONObject getPlatformTypes(HttpSession session, JSONObject json) throws IOException {
     StringBuilder b = new StringBuilder();
     for (String p : PlatformType.platformTypeNames(requestManager.listActivePlatformTypes())) {
-      b.append("<input type='radio' name='platformTypes' id='platformTypes" + p + "' value='" + p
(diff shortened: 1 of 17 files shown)


spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/ContainerControllerHelperService.java
```diff
@@ -430,7 +432,7 @@ public JSONObject changeIlluminaLane(HttpSession session, JSONObject json) {
     int container = json.getInt("container");
     StringBuilder b = new StringBuilder();
     b.append(
-        "<i class='italicInfo'>Click in a partition box to beep/type in barcodes, or double click a pool on the right to sequentially add pools to the container</i>");
+        "<i class='italicInfo'>Click in a lane box to beep/type in barcodes, or double click a pool on the right to sequentially add pools to the container</i>");
```
apmasell: Wait...are we going with lane given PacBio?

callunity: When asked about it, Karen shrugged and said that since there's only one lane/partition per SMRT Cell, it didn't really matter to her.

apmasell:
:+1:

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/ContainerControllerHelperService.java
apmasell: Fair enough.

tonydebat:
👍

PR #389, opened 2016-09-06
Allow compact-migrations to work with HTTP

apmasell:
The URL it was detecting was specific to SSH. This URL detection will match both

### DIFF
--- a/compact-migrations
+++ b/compact-migrations
@@ -34,7 +34,6 @@ fi
 
 NEXT="V$(printf "%04d" ${ID})__auto_${TAIL}.sql"
 echo "Next migration will be ${NEXT}."
-exit 1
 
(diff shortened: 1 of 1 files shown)


compact-migrations
```diff
@@ -34,6 +34,7 @@ fi
 
 NEXT="V$(printf "%04d" ${ID})__auto_${TAIL}.sql"
 echo "Next migration will be ${NEXT}."
+exit 1
```
djcooke: I'm guessing this was for testing and should be removed now?

apmasell: Yes, fixed.

djcooke:
:+1:

tonydebat:
👍

PR #392, opened 2016-09-07
GLT-1090 remove accidental overgeneralization of language in Java

callunity:
Too many `partition` -> `lane` conversions in places they shouldn't be (ie. the Java)


apmasell:
:+1:

djcooke:
:+1:

PR #393, opened 2016-09-07
GLT-1087 show kit on library dropdown

callunity:
Library Kits have names, not aliases.


apmasell:
:+1: Yes, yes, use your hate of inconsistent names/aliases.

djcooke:
:+1:

PR #394, opened 2016-09-07
Allow adding low quality libraries to new pools

apmasell:


callunity:
I think this column should go beside the "Add" column, rather than between the Sample Alias and Indices columns.

callunity:
:+1:

djcooke:
:+1:

PR #396, opened 2016-09-07
Show login name in change logs

apmasell:


callunity:
:+1:

djcooke:
Should the user's name even be in the message? The changelog tables already link to the user, and we could display that in any way we like in a separate column

apmasell:
Lars requested this. A separate column is fine if you prefer.

djcooke:
I think a separate column would be much better. It would also be good for migrated changelogs, as they are linked to the "Geospiza" user, so at least there would be an indication of where they came from in the UI

apmasell:
Rewritten. Please re-review.

callunity:
:+1: looks good.

djcooke:
User name should be removed from changelog messages

apmasell:
Fixed.

djcooke:
:+1:

PR #397, opened 2016-09-07
GLT-1078 check for selected samples right before changing page

callunity:
fixes bug where users could select more samples after selecting the sample class for child samples, but those additional selections wouldn't be carried over to the propagate page.

### DIFF
--- a/miso-web/src/main/webapp/scripts/sample_ajax.js
+++ b/miso-web/src/main/webapp/scripts/sample_ajax.js
@@ -1245,6 +1245,16 @@ Sample.ui = {
       jQuery('#errors').css('display', 'block');
       return false;
     }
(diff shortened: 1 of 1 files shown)


apmasell:
:+1:

djcooke:
New problem:

1. Select a Tissue Sample
2. Choose Propagate (Sample)
3. Select a Tissue Processing class for child type
4. Select a Stock Sample
5. Hit Go

It lets you proceed, though it shouldn't. Steps 3 and 4 are interchangeable, but if you had selected both samples before selecting Propagate (sample), you'd see an error since the possible child classes differ between tissues and stocks.

djcooke:
:+1:

PR #401, opened 2016-09-08
GLT-1071 Display error message for duplicate barcode.

tonydebat:


apmasell:
:+1: I'm concerned about this approach. There were a bunch of catch blocks for `DuplicateKeyException` but they were not thrown in the right cases.

djcooke:
:+1: Makes (probably safe) assumptions, but better than a stacktrace. It would be preferable to try a "getAnythingByBarcode" first and return error if anything (other than same sample) is returned. That would prevent duplicates across tables as well.

PR #406, opened 2016-09-08
GLT-1091 make service records CRUDable by all

callunity:


apmasell:
The code is good, but policy question: do we want to allow non-internal users to edit?

callunity:
The only non-internal users are techs and external users, right? Techs should be able to add them (at least, they had this capacity prior to this commit) and external users can only see the external webapp, which is read-only. Service Records also aren't available to edit via a REST API, so I don't believe it'll be an issue.

apmasell:
Okay. :+1:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/UserAuthMisoRequestManager.java
```diff
@@ -1778,10 +1778,8 @@ public void deleteSequencerReference(SequencerReference sequencerReference) thro
   }
 
   @Override
-  public void deleteSequencerServiceRecord(uk.ac.bbsrc.tgac.miso.core.data.SequencerServiceRecord serviceRecord) throws IOException {
-    if (getCurrentUser().isAdmin()) {
-      backingManager.deleteSequencerServiceRecord(serviceRecord);
-    }
+  public void deleteSequencerServiceRecord(SequencerServiceRecord serviceRecord) throws IOException {
+    backingManager.deleteSequencerServiceRecord(serviceRecord);
```
djcooke: Should all users be able to delete?

djcooke:
I think delete should be admin only, but :+1: if I'm wrong

callunity:
I can get behind admin-only delete, though that means that the sequencing team would be emailing us to do it. Any objections?

callunity:
(For what it's worth, non-admins are able to edit and delete records in the current LIMS)

djcooke:
Having to delete a service record should be pretty rare. I assume that lab managers will be admins in MISO as well though?

callunity:
@tonydebat was suggesting that maybe lab managers won't need to be admins in MISO, since there aren't that many lab-facing admin things. And right now the only admins in GSLE are Lee, Danielle, and Faridah.

djcooke:
The question is whether all users should be able to delete service records. If yes, merge as is. If no, there are several solutions:

1. lab asks us to delete them in the rare case that it's needed
2. make lab managers admin so they can handle it instead
3. make another security group for managers that has extra privileges, but not as much as admin. Let that group delete the records. Either
   * do this now (not likely)
   * use one of the other options for now and do this later

I don't have a strong opinion on what is done. Just being wary of "everyone can delete"

PR #419, opened 2016-09-12
Glt 1056 hot library design

apmasell:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -845,11 +846,24 @@ public ModelAndView editBulkLibraries(@PathVariable String libraryIds, ModelMap
           lai = libraryAdditionalInfoService.get(library.getId());
         }
         libraryDtos.add(Dtos.asDto(library, lai));
+        if (!isDetailedSampleEnabled()) {
+          // Do nothing about sample classes.
+        } else if (sampleClass == null) {
+          sampleClass = ((SampleAdditionalInfo) library.getSample()).getSampleClass();
+        } else if (((SampleAdditionalInfo) library.getSample()).getSampleClass().getId() != sampleClass.getId()) {
+          throw new IOException("Can only update libraries when samples all have the same class.");
```
callunity: This check should also be performed in the JS on List Libraries so users see an alert rather than a stacktrace.

apmasell: Fixed.

apmasell:
Also address GLT-1110

miso-web/src/main/webapp/scripts/shared_hot.js
```diff
@@ -102,33 +106,48 @@ var Hot = {
     });
   },
 
+  findFirstOrNull: function (predicate, referenceCollection) {
+    var results = referenceCollection.filter(predicate);
+    return results.length > 0 ? results[0] : null;
+  },
+  maybeGetProperty: function(obj, propertyName) {
```
apmasell: Not my idea: https://en.wikipedia.org/wiki/Option_type

apmasell:
`saveLibrary` will now throw if mismatched to the design.

PR #422, opened 2016-09-12
Persist sequencing partition container when run is saved

apmasell:


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLRunDAO.java
```diff
@@ -502,6 +502,9 @@ public long save(Run run) throws IOException {
               l.setId(sequencerPartitionContainerDAO.save(l));
             }
             SEQ_PART_CONTAINER_WRITER.saveAll(template, run.getId(), run.getSequencerPartitionContainers());
+            for (SequencerPartitionContainer<SequencerPoolPartition> container : run.getSequencerPartitionContainers()) {
+              sequencerPartitionContainerDAO.save(container);
```
djcooke: These are already saved 4 lines higher up

Was the problem with saving the containers in a run, or with saving the pools within lanes?

apmasell: I idiotically put this in the `saveAll` method where it already existed instead of the `save` method where it is missing. :-/

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLRunDAO.java
```diff
@@ -377,6 +377,9 @@ public long save(Run run) throws IOException {
     }
 
     if (this.cascadeType != null) {
+      for (SequencerPartitionContainer<SequencerPoolPartition> container : run.getSequencerPartitionContainers()) {
+        container.setId(sequencerPartitionContainerDAO.save(container));
```
djcooke: Should be in the [if PERSIST] block below just before updating the join table

apmasell: Fixed.

PR #426, opened 2016-09-15
GLT-1112 & GLT-1113 HoT UI update

callunity:
- increases arrow contrast in autocomplete cells
- visually highlights required cells
  - thoughts on colour scheme? Light orange is already used for nonstandard aliases
- fixes error where second index is marked as required for dual index families


apmasell:
:+1:

PR #430, opened 2016-09-15
Allow updating sample identification barcode from HoT (GLT-1122)

apmasell:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -506,6 +506,9 @@ private void applyChanges(Sample target, Sample source) throws IOException {
     target.setDescription(source.getDescription());
     target.setEmpty(source.isEmpty());
     target.setVolume(source.getVolume());
+    if (!isStringBlankOrNull(source.getIdentificationBarcode())) {
```
djcooke: This will prevent changing to null, which should be allowed

djcooke: Wrong barcode was added, correct one is unknown. Barcode accidentally added to ghost sample.

I think you should be able to remove it. If not, then the UI needs to prevent it too and show an error when you try to clear it

djcooke: Also, if you can't remove it and someone wants to anyway, they'll just set it to "asdfjkl;"

callunity: I'm going with @djcooke on this one.

apmasell: Fixed. The amount of discussion is inversely proportional to the patch size. ;-)

PR #433, opened 2016-09-15
GLT-1120 shift focus after barcode lookup

callunity:
So you can scan the barcode, hit enter, and select your next box position.

### DIFF
--- a/miso-web/src/main/webapp/pages/editBox.jsp
+++ b/miso-web/src/main/webapp/pages/editBox.jsp
@@ -281,14 +281,14 @@
   });
 
   jQuery('#selectedBarcode').keyup(function(event) {
-    if(event.keyCode == 13){
-      Box.lookupBoxableByBarcode();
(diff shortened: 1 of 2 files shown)


miso-web/src/main/webapp/pages/editBox.jsp
```diff
@@ -281,16 +281,22 @@
...
   });
 
   jQuery('#selectedBarcode').on('paste', function(e) {
     window.setTimeout(function() {
-      Box.lookupBoxableByBarcode();
+      jQuery('#lookupBarcode').click();
     }, 100);
   });
+
+  jQuery('#updateSelected').keyup(function(event) {
```
djcooke: Is there a case where hitting Enter on a button doesn't already fire a click event?

callunity: Looking into it further, doesn't seem like it. I will remove this block.

PR #434, opened 2016-09-16
Allow archiving library types

apmasell:


sqlstore/src/main/resources/db/migration/V8000__archive_library_type.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE LibraryType ADD COLUMN archived boolean DEFAULT FALSE;
+UPDATE LibraryType SET archived = TRUE WHERE platformType = 'Illumina' AND description NOT IN ('Paired End', 'Mate Pair', 'Single End');
```
callunity: This should go into `oicr`, unless we're certain EI wants it.

apmasell: Fixed. https://github.com/oicr-gsi/miso-lims/pull/95

sqlstore/src/main/resources/db/migration/V8000__archive_library_type.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE LibraryType ADD COLUMN archived boolean DEFAULT FALSE;
```
djcooke: Should not be nullable

callunity:
In testing this out, I found a (perhaps pre-existing?) bug where if you change the Library Design on an Edit Library page, it wipes out the Index Family info (since Platform is being changed).

callunity:
If you are bulk editing and attempt to remove a Library Design from an item, Handsontable marks the empty Library Design field as invalid and saving is not possible.

apmasell:
Fixed everybody's concerns. ;-)

sqlstore/src/main/resources/db/migration/V8000__archive_library_type.sql
frankie13: Could someone explain the functionality to me please?  It might help with some of our performance concerns...

Also, would this not be best enforced at the DAO level?

callunity: @frankie13 , this will work to keep all the Library Types in the database, but we can now reduce the list shown to the users to only the 3 Types they regularly interact with, so they don't have to scroll through 26 items every time.

djcooke:
Seeing some weird behaviour, though I'm not sure if it directly relates to this PR:
* Edit Library (single), library using an archived libraryType and a libraryDesign that doesn't match: Hitting 'Save' after no changes causes the libraryType to change silently
* Edit Library (single), non-archived libraryType, libraryDesign matches: Change libraryType (to not match libraryDesign), then save. Change reverts, again silently
* Edit Library (bulk), archived libraryType, libraryDesign doesn't match: on load, the libraryDesign's libraryType is shown instead of the saved library's libraryType. Saving again causes it to be changed without notice

apmasell:
Okay. I've changed the single page so that you can't change the library type independently of the design (that was a bug). If design and type don't match, that's a data problem, and I think beyond this PR.

djcooke:
Fix only works after changing Library Design. This affects librarySelection and libraryStrategy the same way too. Need to call Library.ui.changeDesign on page load or something.

miso-web/src/main/webapp/scripts/library_hot.js
```diff
@@ -8,12 +8,16 @@ Library.hot = {
...
-    Hot.dropdownRef.libraryTypes.forEach(function (lt) {
-      if (Library.hot.libraryTypeAliases[lt.platform]) {
-        Library.hot.libraryTypeAliases[lt.platform].push(lt.alias);
-      } else {
-        Library.hot.libraryTypeAliases[lt.platform] = [lt.alias];
+    var usedTypes = Library.hot.librariesJSON.map(function (lib) {
+      return lib.libraryTypeId;
+    });
+    Hot.dropdownRef.libraryTypes.filter(function (lt) {
+      return !lt.archived || usedType.indexOf(lt.id) != -1;
```
callunity: should be `usedTypes`

callunity: You'll want to add `archived` to the list of LibraryType attributes which get written in the EditLibraryController's `referenceDataJsonString` method (or refactor the whole method?).

miso-web/src/main/webapp/scripts/library_hot.js
```diff
@@ -8,12 +8,16 @@ Library.hot = {
...
+    }).forEach(function (lt) {
+      if (!Library.hot.libraryTypeAliases[lt.platform]) {
+        Library.hot.libraryTypeAliases[lt.platform] = [];
       }
+      Library.hot.libraryTypeAliases[lt.platform].push(lt.alias);
```
callunity: Bulk create libraries is fine, but for bulk edit libraries, after `Library.hot.getLibraryTypeAliasLists()` is called on `bulkEditLibraries.jsp`, you'll need to go through each LibraryType cell and update its source to `Library.hot.libraryTypeAliases[${platformName}]`. In bulk edit, no source is being set for library type.

apmasell: I believe this happens indirectly through `changePlatform` on 432.

callunity: But if you don't change the platform then it doesn't set it, so if the user is editing a library and tries to change the libraryType without changing the platform then they don't see any libraryType dropdown options.

apmasell: I see. Fixed.

PR #435, opened 2016-09-16
GLT-1140: create Studies during migration

djcooke:


migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -134,7 +136,20 @@ public void saveProjects(Collection<Project> projects) throws IOException {
...
+        Study study = new StudyImpl(project, user);
+        study.setAlias(project.getShortName() + " study");
+        study.setDescription("");
+        study.setStudyType("Other");
+        study.setLastModifier(user);
+        project.getStudies().add(study);
+      }
+      project.setId(serviceManager.getProjectDao().save(project));
+      for (Study study : project.getStudies()) {
+        study.setId(serviceManager.getStudyDao().save(study));
```
callunity: Do you need to set the study's project before saving?

djcooke: Maybe... More importantly, permissions would need to be inherited if Studies were provided in the migration data. Fixed

PR #436, opened 2016-09-16
Add uniqueness contraints to boxPositionId columns (GLT-1121)

apmasell:
This is to reduce the presentation of a race condition where concurrent
sample/library creation leads to duplicate boxPositionIds. This neither fixes
the underlying race condition in `nextval` nor ensures that there isn't a race
between a sample and a library.


callunity:
Why this approach as opposed to fixing the underlying race condition?

apmasell:
Because fixing the underlying race condition is much harder since we don't know why it occurs and it will likely involve fiddling with transactions, which we aren't using properly in SQLStore.

morgantaschuk:
And this way, if it happens again, it won't make "List Boxes" unreachable.

apmasell:
Err, it's less likely to make _List Boxes_ unreachable, but that's not a guarantee.

callunity:
Please add the JIRA issue name to PR title.

tonydebat:
I like this better than what we have now, but still think we should guarantee uniqueness. Okay approving it if we leave another ticket open that refers to this ticket.

frankie13:
looks good to me 👍

PR #438, opened 2016-09-16
Make all search boxes trim spaces and ignore case

apmasell:
GLT-1129

### DIFF
--- a/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/util/DbUtils.java
+++ b/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/util/DbUtils.java
@@ -34,6 +34,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Matcher;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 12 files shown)


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/util/DbUtils.java
```diff
@@ -66,6 +67,13 @@
   protected static final Logger log = LoggerFactory.getLogger(DbUtils.class);
   private static final HashCodeCacheKeyGenerator hashCodeCacheKeyGenerator = new HashCodeCacheKeyGenerator();
 
+  public static String convertStringToSearchQuery(String input) {
+    if (input == null) {
+      return "%";
+    }
+    return "%" + input.trim().toUpperCase().replaceAll("[\\s,%]+", " ").replaceAll("_", Matcher.quoteReplacement("\\_")) + "%";
```
tonydebat: The % should be escaped to \\% just like the _ character. It is possible to use % as a character in some searchable fields.

tonydebat: Will need '\\s+' outside character class in order to collapse white space.

PR #441, opened 2016-09-20
Cope with missing change logs when listing samples

apmasell:


frankie13:
suggest something like ```INSERT INTO SampleChangeLog SELECT sampleId, 'init' as columnsChanged, 0 as userId, 'init' as message, '1970-01-01 00:00:01' as changeTime FROM Sample;```
as a post migration step?

frankie13:
actually, even that post migration step causes the following error :(
```DataTables warning (table id = 'listingSamplesTable'): Requested unknown parameter 'sampleClassId' from the data source for row 0```

djcooke:
Just handle the nulls to fix the NPEs?

PR #442, opened 2016-09-20
GLT-1127 Tabularize pooled elements

callunity:


spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/PoolControllerHelperService.java
```diff
@@ -808,15 +658,15 @@ public JSONObject createElementSelectDataTable(HttpSession session, JSONObject j
         for (LibraryDilution libraryDilution : requestManager.listAllLibraryDilutionsByPlatform((PlatformType.get(platform)))) {
           JSONArray pout = new JSONArray();
           pout.add(libraryDilution.getName());
-          pout.add(libraryDilution.getConcentration());
+          pout.add(String.format("%.1f", libraryDilution.getConcentration()));
```
djcooke: Is there a reason to limit to 1 decimal place? If so, should input be validated similarly?

miso-web/src/main/webapp/pages/editPool.jsp
```diff
@@ -406,33 +406,63 @@
...
+	    <th>Remove</th>
+	  </tr>
+	  </thead>
+	  <tbody>
+		<c:forEach items="${pool.poolableElements}" var="dil">
+		  <tr id="pooled_${dil.name}">
+		    <td>${dil.name}</td>
+		    <td>${dil.concentration}</td>
+		    <td>${dil.library.alias} (${dil.library.name})</td>
+		    <td>${dil.library.sample.alias} (${dil.library.sample.name})</td>
```
djcooke: Library and Sample should be hyperlinked

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/PoolControllerHelperService.java
callunity: I prefer consistent numbers of decimal places, but prod seems to have up to five decimal places and I refuse to apply that standard to everything else. I'll change it so it's not limited to any decimal places.

PR #446, opened 2016-09-21
Glt 1128+glt 1131 order improvements

apmasell:


apmasell:
Fixed.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/PoolOrderRestController.java
```diff
@@ -54,6 +67,35 @@
...
+    ObjectNode node = mapper.createObjectNode();
+    ArrayNode headings = mapper.createArrayNode();
+    for (HealthType health : LimsUtils.getUsedHealthTypes(completions)) {
+      headings.add(health.getKey());
+    }
+    node.put("headings", headings);
+    ArrayNode data = mapper.createArrayNode();
+    for (Entry<SequencingParameters, PoolOrderCompletionGroup> entry : LimsUtils.groupCompletions(completions).values().iterator().next()
+        .entrySet()) {
+      ObjectNode completion = mapper.createObjectNode();
```
callunity: The `Platform` column seems to have been lost in translation

sqlstore/src/main/resources/db/migration/V8101__change_seqparam_names.sql
```diff
@@ -0,0 +1 @@
+UPDATE SequencingParameters SET name = REPLACE(name, 'x', '×');
```
djcooke: Should be OICR-specific since it fixes changes made there, and it's otherwise fine for there to be an 'x' in the name... Or those changes should be pulled into develop?

apmasell: Either is fine. I'll remove it from this branch.

PR #450, opened 2016-09-21
GLT-1098 Fix run & container count in search

callunity:


apmasell:
Well, that would be a problem.

PR #452, opened 2016-09-22
GLT-1116 hide unused visualisations on List Project page& Edit Project page

callunity:


tonydebat:
The ticket mentions that EI is okay with hiding visualisations. This is more of a delete. If they're really not used though, deleting is good.

callunity:
Dan has said that they don't use the functionality. I'll update the ticket.

PR #459, opened 2016-09-26
GLT-1165: only allow aliquot DetailedSamples to parent Libraries

djcooke:


callunity:
Can we get a check on the front end for this? You can still get to the bulk create page from a stock parent, and it's only on save that the exception pops up.

PR #460, opened 2016-09-27
Allow aliquot receipt (GLT-1055)

apmasell:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/DetailedSampleBuilder.java
```diff
@@ -824,17 +827,14 @@ public void setNonStandardAlias(boolean nonStandardAlias) {
     this.nonStandardAlias = nonStandardAlias;
   }
 
-  public DetailedSample build() {
+  public DetailedSample build(InferStockFromAliquot inferencer) {
```
djcooke: This is not always needed. Please keep build method parameterless. There can be a setter for this and an exception thrown if it's needed and not set

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/DetailedSampleBuilder.java
```diff
@@ -875,23 +871,24 @@ public DetailedSample build() {
...
-        if (tissueClass == null) throw new NullPointerException("Missing tissue class");
+      if (!SampleTissue.CATEGORY_NAME.equals(sampleClass.getSampleCategory()) && tissueClass != null) {
         SampleTissue tissue = buildTissue();
+        tissue.setParent(parent);
         tissue.setSampleClass(tissueClass);
-        tissue.setParent(identity);
-        sample.setParent(tissue);
+        parent = tissue;
+      }
+      if (!SampleStock.CATEGORY_NAME.equals(sampleClass.getSampleCategory())) {
```
djcooke: If the sample being built is a Tissue or Identity, won't this result in a Stock being made as parent of the Tissue?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/DetailedSampleBuilder.java
```diff
@@ -43,8 +43,11 @@
 import uk.ac.bbsrc.tgac.miso.core.exception.ReportingException;
 import uk.ac.bbsrc.tgac.miso.core.security.SecurableByProfile;
 
-public class DetailedSampleBuilder implements DetailedSample, SampleAliquot, SampleStock, SampleTissue, SampleTissueProcessing,
-    SampleCVSlide, SampleLCMTube, Identity {
+public class DetailedSampleBuilder
+    implements DetailedSample, SampleAliquot, SampleStock, SampleTissue, SampleTissueProcessing, SampleCVSlide, SampleLCMTube, Identity {
+  public interface InferStockFromAliquot {
```
djcooke: Non-noun class name :crying_cat_face:

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -291,7 +298,7 @@ private static DetailedSampleDto asDetailedSampleDto(DetailedSample from) {
     return dto;
   }
 
-  private static DetailedSample toDetailedSample(DetailedSampleDto from) {
+  private static DetailedSample toDetailedSample(DetailedSampleDto from, SampleClassCategorizerer categorizer) throws IOException {
```
djcooke: The Service layer is responsible for looking up DB data. Dto conversion should not do this

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleClassService.java
```diff
@@ -78,4 +85,25 @@ public void delete(Long sampleClassId) throws IOException {
...
+  @Override
+  public SampleClass inferStockFromAliquot(SampleClass sampleClass) {
+    SampleClass aliquotClass = sampleClassDao.getSampleClass(sampleClass.getId());
+    if (aliquotClass == null || !aliquotClass.getSampleCategory().equals(SampleAliquot.CATEGORY_NAME)) {
+      throw new IllegalArgumentException("Sample class " + sampleClass.getId() + " is not a valid aliquot class.");
+    }
+    SampleClass stockClass = null;
+    for (SampleValidRelationship relationship : sampleValidRelationshipDao.getSampleValidRelationship()) {
+      if (relationship.getChild().getId() == aliquotClass.getId()
+          && relationship.getParent().getSampleCategory().equals(SampleStock.CATEGORY_NAME)) {
```
djcooke: Make sure it's also not archived. cDNA (aliquot) can be parented to cDNA (stock) or whole RNA (stock), the latter of which is an archived relationship

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
```diff
@@ -820,7 +828,13 @@ public String processSubmit(@ModelAttribute("sample") Sample sample, ModelMap mo
       if (builder.getTissueClass() != null) {
         builder.setTissueClass(sampleClassService.get(builder.getTissueClass().getId()));
       }
-      sample = builder.build();
+      sample = builder.build(new InferStockFromAliquot() {
+
+        @Override
+        public SampleClass infer(SampleClass sampleClass) {
+          return sampleClassService.inferStockFromAliquot(sampleClass);
```
djcooke: If there's no other reason to use this, SampleClassService could just have a method that returns an inferencer, so that a new one that does the same thing isn't created for every request. Or SampleClassService could just implement the interface itself

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/DetailedSampleBuilder.java
apmasell: The caller has no way of knowing when it isn't.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/DetailedSampleBuilder.java
apmasell: http://steve-yegge.blogspot.ca/2006/03/execution-in-kingdom-of-nouns.html

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
apmasell: There's no way to do the inference with out it.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleClassService.java
apmasell: Fixed.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/DetailedSampleBuilder.java
apmasell: Fixed.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/DetailedSampleBuilder.java
djcooke: From a quick skim, this person doesn't seem to understand what verbs and nouns even are...

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/DetailedSampleBuilder.java
djcooke: The caller is capable of making it unnecessary. To infer in a nice way requires DAO access, which the caller may not have

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
apmasell: Fixed. The performance implications of creating the extra object are negligible.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/DetailedSampleBuilder.java
apmasell: How do you recommend doing it?

djcooke:
Is there a reason the inference has to happen within the builder? Couldn't it happen before build like lookup of SampleClasses (e.g. `EditSampleController#processSubmit(Sample, ModelMap, SessionStatus`)?

apmasell:
Only the builder knows what data is has and how to slice itself into a hierarchy. Either the calling code with DAO access needs to pick apart the sample builder into the hierarchy or the builder needs to know about all the sample classes and relationships. I can't see a way to do this without violating one type of encapsulation. If you think the other is better, I can spill large chunks of the builder's logic into the controllers. Same argument for the DTO conversion.

djcooke:
>spill large chunks of the builder's logic into the controllers

```
if (builder.getSampleClass().getSampleCategory() == SampleAliquot.CATEGORY_NAME
    && builder.getParent() == null && builder.getStockClass() == null) {
  builder.setStockClass(SamplelassDao.inferStockClassFromAliquot(builder.getSampleClass()))
}
```
DetailedSampleBuilder needs a `stockClass` field, and it can be handled in the `build()` method, and in DefaultSampleService the same way tissueClass is. The only difference is that it isn't user-provided.

Does this work?

apmasell:
Sure. And the DTOs?

djcooke:
DTOs should be WYSIWYG. Loading database entities is the responsibility of the Service layer, which DTO conversion should not have access to. Controllers can load the SampleClass via Service when necessary

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/DetailedSampleBuilder.java
```diff
@@ -875,23 +882,28 @@ public DetailedSample build() {
...
+        throw new IllegalArgumentException("Sample has no parent and cannot infer order from sample category.");
       }
-      identity.setExternalName(externalName);
-      identity.setInternalName(internalName);
-      identity.setDonorSex(donorSex);
-      if (SampleTissue.CATEGORY_NAME.equals(sampleClass.getSampleCategory())) {
-        sample.setParent(identity);
-      } else {
-        if (tissueClass == null) throw new NullPointerException("Missing tissue class");
+      if (categoryIndex > 0 && externalName != null) {
```
djcooke: Instead of using indices explicitly, could this be refactored into a method like `needsParentCategory(String parentCategory)`?

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -339,14 +340,11 @@ private static DetailedSample toDetailedSample(DetailedSampleDto from) {
...
-   * Extracts parent details from the DTO, according to three possible cases:
+   * Extracts parent details from the DTO, according to these possible cases:
    * 
    * <ol>
    * <li>parent ID is provided. This implies that the parent exists, so no other parent information will be required</li>
-   * <li>identity information and parentSampleClassId are provided. This implies that a tissue parent should be created, and that the
-   * identity may or may not yet exist</li>
-   * <li>identity information is provided, but no parentSampleClassId. This implies that the direct parent of this sample will be an
-   * identity, which may or may not yet exist</li>
+   * <li>there is no parent ID provided, but there is the information to reconstruct the hierarchy</li>
```
djcooke: What information? This documentation is much more useful if it's specific since it's easy to miss something when writing JSON

miso-dto/src/test/java/uk/ac/bbsrc/tgac/miso/dto/DtosTestSuite.java
```diff
@@ -15,7 +20,7 @@
 public class DtosTestSuite {
```
djcooke: Please add test for converting to aliquot with new parent (stock, tissue, identity) info

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -356,21 +354,29 @@ private static DetailedSample toDetailedSample(DetailedSampleDto from) {
...
-      } else {
-        // Case 3
-        parent = identity;
+    } else {
+      if (childDto instanceof SampleIdentityDto && childDto.getClass() != SampleIdentityDto.class) {
+        parent = toIdentitySample((SampleIdentityDto) childDto);
+      }
+
+      if (childDto instanceof SampleTissueDto && childDto.getClass() != SampleTissueDto.class) {
+        if (childDto.getParentSampleClassId() == null) {
```
djcooke: Please rename DetailedSample.parentSampleClassId to indicate that it refers to the tissue-level parent only, and not always the direct parent

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -350,6 +357,8 @@ private DetailedSample findOrCreateParent(DetailedSample sample) throws IOExcept
       }
     } else if (isTissueSample(tempParent)) {
       return createParentTissue((SampleTissue) tempParent, sample);
+    } else if (isStockSample(tempParent)) {
```
djcooke: Add test cases for this similar to the existing ones in DefaultSampleServiceTestSuite

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
```diff
@@ -142,6 +151,21 @@ public SampleDto getSample(@PathVariable("id") Long id, UriComponentsBuilder uri
...
     try {
+      if (sampleDto instanceof SampleAliquotDto) {
+        SampleAliquotDto dto = (SampleAliquotDto) sampleDto;
+        if (dto.getParentId() != null) {
+          // Pass
+        } else if (dto.getSampleClassId() == null) {
+          throw new IllegalStateException("No parent and no target sample class.");
+        } else {
+          SampleClass sampleClass = sampleClassService.get(dto.getSampleClassId());
+          if (!sampleClass.getSampleCategory().equals(SampleAliquot.CATEGORY_NAME)) {
```
djcooke: Missing null check on sampleClass

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleClassService.java
```diff
@@ -78,4 +85,25 @@ public void delete(Long sampleClassId) throws IOException {
...
   }
 
+  @Override
+  public SampleClass inferStockFromAliquot(SampleClass sampleClass) {
+    SampleClass aliquotClass = sampleClassDao.getSampleClass(sampleClass.getId());
+    if (aliquotClass == null || !aliquotClass.getSampleCategory().equals(SampleAliquot.CATEGORY_NAME)) {
+      throw new IllegalArgumentException("Sample class " + sampleClass.getId() + " is not a valid aliquot class.");
+    }
+    SampleClass stockClass = null;
+    for (SampleValidRelationship relationship : sampleValidRelationshipDao.getSampleValidRelationship()) {
```
djcooke: Could make a query in SampleValidRelationship DAO for this instead of getting all and filtering in Java. Should perform a bit better, which may be relevant when bulk creating a large number (96) of samples

miso-web/src/main/webapp/scripts/sample_ajax.js
```diff
@@ -1384,32 +1389,60 @@ Sample.ui = {
...
   },
   /**
-   * Returns an array of sample classes that correspond to given sample class IDs (for child samples)
+   * Returns an array of sample classes that correspond to given sample class IDs (for descendent samples)
+   * There must exist a path (of the same length) between all the starting sample classes and the targets.
    */
-  getChildSampleClasses: function (sampleClassIDs) {
-    return Sample.sampleClasses.filter(function (sc) {
-      return sampleClassIDs.indexOf(sc.id) != -1;
+  getMatchedDescentSampleClasses: function (sampleClassIDs) {
```
djcooke: This results in being able to choose to propagate from tissue to aliquot, but after entering aliquot details, and hitting save, it fails with the message "Parent Primary Tumor Tissue not permitted to have a child of type gDNA (aliquot)." I'm not sure if propagating from Tissue to Aliquot should be allowed, but my guess is no.

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
apmasell: Fixed.

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
apmasell: Fixed.

miso-dto/src/test/java/uk/ac/bbsrc/tgac/miso/dto/DtosTestSuite.java
```diff
@@ -15,7 +20,7 @@
...
 
   @Test
-  public void testToStockWithNewParents() {
+  public void testToStockWithNewParents() throws IOException {
```
apmasell: Fixed.

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
apmasell: Reverted.

miso-dto/src/test/java/uk/ac/bbsrc/tgac/miso/dto/DtosTestSuite.java
apmasell: Added.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
apmasell: Fixed.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleClassService.java
apmasell: The SVR set is small enough not to necessitate any optimisation.

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
apmasell: Fixed.

miso-web/src/main/webapp/scripts/sample_ajax.js
apmasell: I don't know. I'll get feedback. I suspect it should be allowed.

apmasell:
Fixed. I had to rebase due to the database schema being out of date, but the last changes are in a separate patch.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
```diff
@@ -142,6 +151,24 @@ public SampleDto getSample(@PathVariable("id") Long id, UriComponentsBuilder uri
     }
     Long id = null;
     try {
+      if (sampleDto instanceof SampleAliquotDto) {
+        SampleAliquotDto dto = (SampleAliquotDto) sampleDto;
+        if (dto.getParentId() != null) {
+          // Pass
+        } else if (dto.getSampleClassId() == null) {
+          throw new IllegalStateException("No parent and no target sample class.");
```
djcooke: This and the others below should be RestExceptions to form a clear error response with the correct status. This will end up with 500 status (Internal Server Error) when it is really a client-side error if the DTO is missing data

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -348,14 +349,14 @@ private static DetailedSample toDetailedSample(DetailedSampleDto from) {
...
    * 
    * <ol>
    * <li>parent ID is provided. This implies that the parent exists, so no other parent information will be required</li>
-   * <li>identity information and parentSampleClassId are provided. This implies that a tissue parent should be created, and that the
+   * <li>identity information and parentTissueSampleClassId are provided. This implies that a tissue parent should be created, and that the
    * identity may or may not yet exist</li>
-   * <li>identity information is provided, but no parentSampleClassId. This implies that the direct parent of this sample will be an
-   * identity, which may or may not yet exist</li>
+   * <li>identity information is provided, but no parentTissueSampleClassId. This implies that the direct parent of this sample will be an
+   * identity, which may or may not yet exist. If the sampleClassId is an aliquot, a stockClassId must be provided.</li>
```
djcooke: But if the sampleClassId is an aliquot, then a missing parentTissueSampleClassId would be invalid... right? I think the "If the sampleClassId is an aliquot, a stockClassId must be provided." bit belongs with the previous point ("identity information and parentTissueSampleClassId are provided.")

djcooke:
Creating smRNA, mRNA, or rRNA_depleted still results in the error "Aliquot class 20 has no stock class."

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
apmasell: The addition belongs with the second case and the third case no longer exists.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
apmasell: Fixed.

apmasell:
I've removed smRNA and friends from the list of options.

apmasell:
Fixed.

callunity:
The Create single Sample page has the option to select one of the derived RNA classes (smRNA, rRNA_depleted, mRNA) which then errors with `Aliquot class 20 has no stock class.`

apmasell:
Fixed.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
```diff
@@ -154,14 +154,14 @@ public SampleDto getSample(@PathVariable("id") Long id, UriComponentsBuilder uri
...
+          throw new RestException("No parent and no target sample class.", Status.BAD_REQUEST);
         } else {
           SampleClass sampleClass = sampleClassService.get(dto.getSampleClassId());
           if (sampleClass == null) {
-            throw new RestException("Cannot find sample class: " + dto.getSampleClassId());
+            throw new RestException("Cannot find sample class: " + dto.getSampleClassId(), Status.BAD_REQUEST);
           }
           if (!sampleClass.getSampleCategory().equals(SampleAliquot.CATEGORY_NAME)) {
-            throw new RestException("Class and type mismatch.");
+            throw new RestException("Class and type mismatch.", Status.BAD_REQUEST);
```
callunity: Will the user know what "class" and "type" are in these cases?

apmasell: No, but the user shouldn't be able to produce this error. If they do, it's a UI bug. This means that the front end picked a sample class of type aliquot and then wrote the wrong type attribute into the JSON.

callunity:
I'm a little leery of sorting sample classes by alias, since the stock and aliquot ones will be side-by-side and it'll be very easy to mistake them. What about sorting by category?

apmasell:
I immediately regret mixing that in this PR. I'll make a separate one for that and delete it from here.

PR #462, opened 2016-09-28
Allow marking different instances of MISO (GLT-1171)

apmasell:


miso-web/src/main/webapp/styles/style.css
```diff
@@ -1057,6 +1057,14 @@ table td.headertable {
...
 }
 
+td.headertable {
+    font-size: 50px;
+}
+
+td.headertable img {
+	vertical-align: top;
+}
+
```
callunity: What do you think of the following styling? 
- Wrap the `${initparam['miso.name']}` in a div#instanceName (or whatever name)
- ```td.headertable #instanceName {
  font: Helvetica;
  padding: 0 30px;
  color: #666666;
  display: inline;
  font-size: 40px;
}```

apmasell: Thank you. I have no aesthetic skills.

PR #464, opened 2016-09-28
GLT-1065 hide qcPassed & use detailedQcStatus

callunity:
- changes `QcPassedDetail` to `DetailedQcStatus`
- hides QC Passed if `detailedSample` is enabled, and uses `DetailedQcStatus` in lieu.
- null `DetailedQcStatus` is shown to user as `Not Ready`
- some formatting changes on JSPs


sqlstore/src/main/resources/db/migration/V8092__qcPassedDetails_to_detailedQcStatus.sql
```diff
@@ -0,0 +1,39 @@
...
+    SELECT s.sampleId FROM Sample s
+    WHERE s.qcPassed IS NULL);
+UPDATE DetailedSample ds
+  SET ds.qcPassedDetailId = (SELECT qcPassedDetailId FROM QcPassedDetail WHERE description = 'Failed: QC')
+  WHERE ds.sampleId IN (
+    SELECT s.sampleId FROM Sample s
+    WHERE s.qcPassed = false);
+
+ALTER TABLE DetailedSample ADD COLUMN detailedQcStatusNote VARCHAR(500) DEFAULT NULL;
+UPDATE DetailedSample SET detailedQcStatusNote = 'Default Note' WHERE qcPassedDetailId IN (
```
apmasell: Could this be “details not available” or something more informative?

callunity: It could be, but this query could also be removed entirely -- there are zero items in the prod db right now which have a DetailedQcStatus which requires a note, and it's a nullable field. Would that be more clear?

apmasell: Sure. That works.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -226,6 +233,10 @@ public Long create(Sample sample) throws IOException {
           throw new IOException(e.getMessage(), e);
         }
         validateHierarchy(detailed);
+
+        if (detailed.getDetailedQcStatus() != null) {
```
djcooke: This field is present on Identities too, so should be treated the same (outside of the `if (!isIdentitySample...` block). Alternately, it could be done in `loadChildEntities` instead, which is called during both create and update

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
```diff
@@ -453,15 +453,15 @@ public int compare(TissueType o1, TissueType o2) {
...
-    Collections.sort(list, new Comparator<QcPassedDetail>() {
+  @ModelAttribute("detailedQcStatuses")
+  public List<DetailedQcStatus> getDetailedQcStatuses() throws IOException {
+    List<DetailedQcStatus> list = new ArrayList<>(detailedQcStatusService.getAll());
+    Collections.sort(list, new Comparator<DetailedQcStatus>() {
       @Override
-      public int compare(QcPassedDetail o1, QcPassedDetail o2) {
-        return o1.getDescription().compareTo(o2.getDescription());
+      public int compare(DetailedQcStatus o1, DetailedQcStatus o2) {
+        return o1.getId().compareTo(o2.getId());
```
djcooke: Would it be better to sort by status (order: true > null > false), noteRequired (false > true) and description (A > Z)?

sqlstore/src/main/resources/db/migration/V8092__qcPassedDetails_to_detailedQcStatus.sql
```diff
@@ -0,0 +1,37 @@
...
+INSERT INTO QcPassedDetail (status, description, noteRequired, createdBy, creationDate, updatedBy, lastUpdated) 
+  SELECT TRUE, 'Ready', FALSE, (SELECT userId FROM User WHERE loginName = 'admin'), NOW(), (SELECT userId FROM User WHERE loginName = 'admin'), NOW() 
+  WHERE NOT EXISTS (SELECT description FROM QcPassedDetail WHERE description = 'Ready' LIMIT 1);
+INSERT INTO QcPassedDetail (status, description, noteRequired, createdBy, creationDate, updatedBy, lastUpdated) 
+  SELECT FALSE, 'Failed: QC', FALSE, (SELECT userId FROM User WHERE loginName = 'admin'), NOW(), (SELECT userId FROM User WHERE loginName = 'admin'), NOW() 
+  WHERE NOT EXISTS (SELECT description FROM QcPassedDetail WHERE description = 'Failed: QC' LIMIT 1);
+
+-- remove unnecessary option, as it will be the null case.
+ALTER TABLE DetailedSample DROP FOREIGN KEY `FKa2t38wms0eer896xo4fw76tw0`;
+DELETE FROM QcPassedDetail WHERE description = 'Not Ready';
```
djcooke: This migration will fail when re-adding the foreign key if anything uses that value. Skip dropping the foreign key and first do:
```
UPDATE DetailedSample SET qcPassedDetailId = NULL WHERE qcPassedDetailId = (
  SELECT qcPassedDetailId FROM QcPassedDetail WHERE description = 'Not Ready';
);
```

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
callunity: Hmm, looks like that code is already present in `loadChildEntities`. I'll delete it from `create` and `update`.

sqlstore/src/main/resources/db/migration/V8092__qcPassedDetails_to_detailedQcStatus.sql
callunity: changed

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
callunity: changed

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
callunity: changed

tonydebat:
I'm using the latest production db and am unable to migrate.
MySql version 5.7.15

```
[ERROR] ---------------------------------------------------------------
[ERROR] SQL State  : 42S01
[ERROR] Error Code : 1050
[ERROR] Message    : Table './miso/detailedsample' already exists
[ERROR] Location   : db/migration/V8092__qcPassedDetails_to_detailedQcStatus.sql (/Users/tdebat/code/gsi/miso-lims/sqlstore/target/classes/db/migration/V8092__qcPassedDetails_to_detailedQcStatus.sql)
[ERROR] Line       : 40
[ERROR] Statement  : ALTER TABLE DetailedSample ADD FOREIGN KEY (detailedQcStatusId) REFERENCES DetailedQcStatus (detailedQcStatusId)
[ERROR] -> [Help 1]
```

callunity:
@tonydebat I'm not sure why that's happening; I'm able to copy both the latest production and staging dbs and am unable to replicate the error. That's very unusual that it's failing there, at the very end of the file -- if there was a "table already exists" error, shouldn't it be thrown earlier?

Looking at the prod and staging dbs, I'm not sure that they got migrated after the latest release (the table `SampleAdditionalInfo` isn't renamed to `DetailedSample`, for instance).

callunity:
(removed previous comment about dbs not being migrated, because the backups happen before migration so of course they won't contain migrated data)

PR #468, opened 2016-09-30
Fix bug where Spring erases the contents of a pool

apmasell:


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLPoolDAO.java
```diff
@@ -535,7 +535,13 @@ public String mapRow(ResultSet rs, int pos) throws SQLException {
         ldc = cacheManager.getCache("lazy" + type + "Cache");
       }
 
+      Set<String> elementIds = new HashSet<>();
```
callunity: Maybe rename this to make it clear it's a unique set of elementIds?

PR #471, opened 2016-09-30
GLT-1178 Preventing containers with duplicate Serial Numbers from being

tonydebat:
added to the database.
- Added Container constraints to db including FK and UNIQUE for the serial
  number.
- When deleting a container ensured that the container is unlinked from
  other tables before the delete takes place to satisfy the FKs.
- Prevented form buttons from submitting the form. We'll let js
  handle that.
- Looked up the container serial number to ensure it's unique. If
  not report an error via parsley.

### DIFF
--- a/miso-web/src/main/webapp/scripts/sequencer_partition_container_ajax.js
+++ b/miso-web/src/main/webapp/scripts/sequencer_partition_container_ajax.js
@@ -137,17 +137,51 @@ var Container = Container || {
     jQuery('#container-form').parsley();
     jQuery('#container-form').parsley().validate();
     
-    console.log(jQuery('div[id^="studySelectDiv"]'));
+    if(jQuery("#identificationBarcode").length == 0) {
+    	// Serial number is not being modified.
+        Validate.updateWarningOrSubmit('#container-form', Container.validateStudyAdded);
+    } else {
+    	// Ensure provided serial number is unique.
+    	var serialNumber = jQuery("#identificationBarcode").val();
+    	var containerId = null;
+		if(jQuery("#containerId").length > 0) {
+		  containerId = jQuery("#containerId").val();
+		}
+
+	    Fluxion.doAjax(
+        'containerControllerHelperService',
+        'isSerialNumberUnique',
+        {
+          'serialNumber': serialNumber,
+          'containerId': containerId,
+          'url': ajaxurl
+        },
+        {
+          'doOnSuccess': function(json) {
+        	  if(json.isSerialNumberUnique === "true") {
+                  Validate.updateWarningOrSubmit('#container-form', Container.validateStudyAdded);
+        	  } else {
+        		  // Serial number is not unique.
+                  var serialNumberField = jQuery('#identificationBarcode').parsley();
+            	  window.ParsleyUI.addError(serialNumberField, "serialNumberError", 'This serial number is already in use. Please choose another.');
+            	  return false;
+        	  }
+          },
+          'doOnError': function(json) {
+        	// Unable to perform lookup.
+        	alert(json.error);
+        	window.ParsleyUI.addError(serialNumberField, "serialNumberError", 'Unable to determine if serial number is unique.');
+        	return false;
+           }
+        }
+      );
+    }
     
-    Validate.updateWarningOrSubmit('#container-form', Container.validateStudyAdded);
-    return false;
   },
   
   validateStudyAdded: function () {
     var ok = true;
     if (jQuery('div[id^="studySelectDiv"]').length > 0) {
-      console.log(jQuery('div[id^="studySelectDiv"]'));
-      console.log("whoops");
       if (!confirm("You haven't selected a study for one or more pools. Are you sure you still want to save?")) {
         alert("Please select a Study for each Pool added.\n");
         ok = false;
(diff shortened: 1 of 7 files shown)


sqlstore/src/main/resources/db/migration/V8000__container_constraints.sql
```diff
@@ -0,0 +1,14 @@
+--StartNoTest
+ALTER TABLE Run_SequencerPartitionContainer ADD FOREIGN KEY (Run_runId) REFERENCES Run (runId);
+DELETE FROM Run_SequencerPartitionContainer where containers_containerId = 0;
+ALTER TABLE Run_SequencerPartitionContainer ADD FOREIGN KEY (containers_containerId) REFERENCES SequencerPartitionContainer (containerId);
+
+ALTER TABLE SequencerPartitionContainer_Partition ADD FOREIGN KEY (container_containerId) REFERENCES SequencerPartitionContainer (containerId);
+ALTER TABLE SequencerPartitionContainer_Partition ADD FOREIGN KEY (partitions_partitionId) REFERENCES _Partition (partitionId);
+
+ALTER TABLE SequencerPartitionContainer ADD UNIQUE (identificationBarcode);
```
callunity: Won't this cause the migration to fail for EI since they have multiple nulls and several duplicates?

tonydebat: Yes, anyone performing the migration will need to remove duplicates and nulls first.  You can see the list of EI duplicates in the ticket. https://jira.oicr.on.ca/browse/GLT-1178

sqlstore/src/main/resources/db/migration/V8000__container_constraints.sql
```diff
@@ -0,0 +1,14 @@
+--StartNoTest
```
callunity: This file will need to be renamed, as there is already a V8000 in develop.

sqlstore/src/main/resources/db/migration/V8000__container_constraints.sql
callunity: This will also fail on our production db, as we have two containers with identificationBarcode `C8G67` (containers 48 and 512). Both of these containers are associated with pools, though only container 48 is associated with a run. What do you think will be the best course of action for handling this duplicate?

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/ContainerControllerHelperService.java
```diff
@@ -931,4 +931,36 @@ public void setRequestManager(RequestManager requestManager) {
...
+          // The serial number is unique. Lookup returned the container being edited.
+          responseMap.put("isSerialNumberUnique", true);
+        } else {
+          responseMap.put("isSerialNumberUnique", false);
+        }
+      }
+      return JSONUtils.JSONObjectResponse(responseMap);
+    } catch (IOException e) {
+      log.error("unable to lookup barcode", e);
+      return JSONUtils.JSONObjectResponse("error", "Unable to lookup barcode.");
```
callunity: `barcode` should be changed to `serial number` as this is user-facing text.

miso-web/src/main/webapp/scripts/sequencer_partition_container_ajax.js
```diff
@@ -137,17 +137,50 @@ var Container = Container || {
...
+        		  // Serial number is not unique.
+                  var serialNumberField = jQuery('#identificationBarcode').parsley();
+            	  window.ParsleyUI.addError(serialNumberField, "serialNumberError", 'This serial number is already in use. Please choose another.');
+            	  return false;
+        	  }
+          },
+          'doOnError': function(json) {
+        	// Unable to perform lookup.
+        	alert(json.error);
+        	return false;
```
callunity: Please add Parsley error on failure, similar to line 166 above.

callunity:
Changing the serial number text to a different (unique) string and then clicking "Save" (_without_ clicking "Lookup") keeps the same container associated with the run, but changes its serial number. We may wish to make another ticket out of this though.

sqlstore/src/main/resources/db/migration/V8000__container_constraints.sql
tonydebat: The latest release reset things, so this number is valid once again.

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/ContainerControllerHelperService.java
tonydebat: Changed strings to serial number.

miso-web/src/main/webapp/scripts/sequencer_partition_container_ajax.js
tonydebat: Added parsley error to ensure user cannot continue after error occurs.

tonydebat:
Heather, with regards to providing a unique serial number and saving: I'm going to call that correct behaviour. No ticket required.

tonydebat:
This PR has been updated. The serial number is no longer unique, so no db corrections are required by EI before this PR can be merged.

miso-web/src/main/webapp/scripts/sequencer_partition_container_ajax.js
callunity: I'm not seeing the line where the parsley error is added on the 'doOnError' condition, am I missing something? (this should also likely go in the 'doOnError' condition in the run_ajax.js file on line 112)

PR #473, opened 2016-10-03
GLT-1180 DB and SampleService changes

callunity:
- remove internalName from Identity
- change externalName to Set<String>
- external names must now be unique on a per-project basis (possible to turn this requirement off during migration)
- when creating a new sample, existing identity parents are looked up by ID if available, and created anew otherwise.

### DIFF
--- a/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/SampleDao.java
+++ b/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/SampleDao.java
@@ -52,6 +52,6 @@ public interface SampleDao extends SampleStore {
 
   Long countBySearch(String querystr) throws IOException;
 
-  Identity getIdentityByExternalName(String externalName) throws IOException;
+  Collection<Identity> getIdentitiesByExternalNameOrAlias(String externalName) throws IOException;
 
 }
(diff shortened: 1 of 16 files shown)


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/SampleDao.java
```diff
@@ -52,6 +52,6 @@
 
   Long countBySearch(String querystr) throws IOException;
 
-  Identity getIdentityByExternalName(String externalName) throws IOException;
+  Collection<Identity> getIdentitiesByExternalNameOrAlias(String externalName) throws IOException;
```
djcooke: Add Javadoc detailing parameter expectations and what exactly will be returned. Judging by the implementation in HibernateSampleDao, it looks like this is intended to find by a single external name (not comma-separated list), and return anything `LIKE '%externalName%'`. It would be easy to assume externalName should be a comma-separated list, since that's what `Identity#getExternalName()` returns

sqlstore/src/main/resources/db/migration/afterMigrate.sql
```diff
@@ -217,15 +217,13 @@ FOR EACH ROW
   DECLARE log_message varchar(500) CHARACTER SET utf8;
   SET log_message = CONCAT_WS(', ',
      CASE WHEN NEW.donorSex <> OLD.donorSex THEN CONCAT('donor sex: ', OLD.donorSex, ' → ', NEW.donorSex) END,
-     CASE WHEN NEW.externalName <> OLD.externalName THEN CONCAT('external name: ', OLD.externalName, ' → ', NEW.externalName) END,
-     CASE WHEN NEW.internalName <> OLD.internalName THEN CONCAT('internal name: ', OLD.internalName, ' → ', NEW.internalName) END);
+     CASE WHEN NEW.externalName <> OLD.externalName THEN CONCAT('external name: ', OLD.externalName, ' → ', NEW.externalName) END);
```
djcooke: Column no longer exists. Identity_ExternalName changes should be logged to SampleChangeLog though

callunity:
Closing until the UI changes are ready to go too.

PR #474, opened 2016-10-04
Fix problem where null library volume changed to zero (GLT-1137)

apmasell:


miso-web/src/main/webapp/scripts/library_hot.js
```diff
@@ -619,7 +619,7 @@ Library.hot = {
       if (obj.volume) {
         lib.volume = obj.volume;
       } else {
-        lib.volume = 0;
+        lib.volume = null;
```
djcooke: If obj.volume is 0 here, it ends up getting set to null. 0 should be a savable value as well though

apmasell:
Fixed.

djcooke:
Now volume is always 1.0 on load of Edit Library (bulk) page, whether the volume is actually null, 0, or 1. If you change it and save, it saves correctly, but then always displays 1.0 on load still. Single Edit page works fine

apmasell:
That's once again due to Spring poisoning the cache by changing the model object which is also the cached copy.

PR #475, opened 2016-10-04
Glt 614 empty pool

apmasell:


tonydebat:
During testing I noticed that Pool change logs are not displayed until the cache is cleared. I've opened a separate issue for this. https://jira.oicr.on.ca/browse/GLT-1195

apmasell:
No. That variable is never read, so I'll just not load it any time.

PR #478, opened 2016-10-05
GLT-1188: add/revise unique constraints for value types

djcooke:


djcooke:
@frankie13: please test this migration on EI database since it adds unique constraints to a few old tables

PR #479, opened 2016-10-06
GLT-1172 Converted 'trash' and 'empty' to 'discard'. 

tonydebat:
This harmonizes the language used when throwing away tubes.


callunity:
One of the items on the ticket is `checking "discarded" on a single item disables its volume field`. Has that been implemented here?

PR #480, opened 2016-10-07
Note sample name before saving and show purpose problems

apmasell:


migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -180,6 +180,7 @@ private void saveSample(Sample sample) throws IOException {
       // already saved
       return;
     }
+    log.debug("Saving sample " + sample.getAlias());
```
djcooke: Actually, should move this below the next block because saving this sample will not happen until after saving the parent (if necessary)

callunity: :+1:

PR #484, opened 2016-10-11
GLT-1180 & GLT-1179 improved Identity lookup

callunity:
- when receiving new detailed samples via Handsontable, users must select a parent identity before saving the new sample.


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
```diff
@@ -178,4 +185,48 @@ public SampleDto getSample(@PathVariable("id") Long id, UriComponentsBuilder uri
...
+    JSONArray searchTerms = JSONArray.fromObject(identitiesSearches.get("identitiesSearches"));
+
+    List<Set<SampleDto>> foundIdentities = new ArrayList<Set<SampleDto>>();
+    for (int i = 0; i < searchTerms.size(); i++) {
+      Set<Sample> resultsForOneParent = new HashSet<Sample>();
+      Set<SampleDto> resultsForOneParentDtos = new HashSet<SampleDto>();
+
+      for (String term : IdentityImpl.getSetFromString(((String) searchTerms.get(i)).replaceAll(";", ","))) {
+        Collection<Identity> selectedIdentities = sampleService.getIdentitiesByExternalNameOrAlias(term);
+        if (!selectedIdentities.isEmpty()) {
```
apmasell: This `if` is redundant. The loop will cover an empty set.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
```diff
@@ -178,4 +185,48 @@ public SampleDto getSample(@PathVariable("id") Long id, UriComponentsBuilder uri
...
+          for (Identity selectedIdentity : selectedIdentities) {
+            if (resultsForOneParent.contains(selectedIdentity)) {
+              continue;
+            } else {
+              resultsForOneParentDtos.add(Dtos.asDto(selectedIdentity));
+              resultsForOneParent.add(selectedIdentity);
```
apmasell: Can these go directly into the `foundIdentities` set?

miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -1457,5 +1483,101 @@ Sample.hot = {
...
+   */
+  setIdentitySources: function () {
+    var identityColIndex = Hot.getColIndex('identityAlias');
+    var rowCount = Sample.hot.foundIdentities.length;
+    var selectedProjectId = Sample.hot.selectedProjectId;
+    for (var i = 0; i < rowCount; i++) {
+      var hasIdentityInProject = false;
+      var sortedIdentities = Sample.hot.foundIdentities[i].sort(function (a, b) {
+        hasIdentityInProject = (a.projectId == selectedProjectId);
+        return (a.projectId == selectedProjectId || a.projectId > b.projectId) ? 1 : ((b.projectId > a.projectId) ? -1 : 0);
```
apmasell: The mutation of `hasIdentityInProject` from inside the comparison is not guaranteed to work. If only one item is returned, the list is sorted, but the comparison function will never be called. The comparison itself is also hard to reason about. Might I suggest:
```
var sortedIdentities = Sample.hot.foundIdentities[i].sort(function (a, b) {
  var aSortId = a.projectId == selectedProjectId ? 0 : a.projectId;
  var bSortId = b.projectId == selectedProjectId ? 0 : b.projectId;
  return aSortId - bSortId;
});
var hasIdentityInProject = (sortedIdentities.length > 0 && sortedIdentities[0].projectId != selectedProjectId);
var identityNames = sortedIdentities.map(Sample.hot.getIdentityLabel);
if (!hasIdentityInProject) sortedIdentities.unshift("First Receipt");
```
I would also avoid changing the contents of a variable from one type to another. It's better to create a new variable.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
callunity: No, since `resultsForOneParent` is the set of possibilities for a given externalName string, whereas `foundIdentities` corresponds to the ordered list of given externalNames.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
callunity: fixed

miso-web/src/main/webapp/scripts/sample_hot.js
callunity: fixed

callunity:
Realized I didn't make the update available for Create Single Sample. Working on that now.

apmasell:
Looks good for this part of though.

callunity:
PR for docs update: https://github.com/oicr-gsi/miso-docs-oicr/pull/26

PR #486, opened 2016-10-12
Cope with nulls in the list of indices

apmasell:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractLibrary.java
```diff
@@ -229,7 +229,13 @@ public String getLabelText() {
   }
 
   @Override
-  public void setIndices(List<Index> indices) {
+  public void setIndices(List<Index> originalIndices) {
```
djcooke: Odd parameter name from caller's perspective

apmasell: Caller never sees it due to the interface.

djcooke: Ah, right

PR #487, opened 2016-10-12
GLT-1203: fixed list cache updating

djcooke:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Idable.java
```diff
@@ -0,0 +1,12 @@
+package uk.ac.bbsrc.tgac.miso.core.data;
+
+public interface Idable {
```
apmasell: Identifiable

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Identifiable.java
```diff
@@ -0,0 +1,12 @@
+package uk.ac.bbsrc.tgac.miso.core.data;
+
+public interface Identifiable {
+
+  /**
+   * Returns the unique ID of this Idable object.
```
callunity: `Idable` -> `Identifiable`

PR #488, opened 2016-10-12
more GLT-1206: remove unused box methods; update UI in JSPs

callunity:
- remove unused box methods
- update location of "Location" row and improve pluralization on JSPs for Sample, Library, Pool


callunity:
The location update is fixed (better!) in PR #487 . I am therefore modifying this PR to remove the caching fixes and keep the other minor updates.

djcooke:
Was there a reason to move "Location" further down on the pages? I'd think this should be kept prominently near the top for quicker reference, and also to keep it grouped with the basic attributes like ID, name, alias, and description

callunity:
My thoughts were that it doesn't really belong squished between Library ID and Name (both semantically and it also doesn't stand out visually), and it belongs semantically with Volume and Discarded, and has a slight advantage for standing out by being the last item in the list.

PR #489, opened 2016-10-13
Sort all the things alphabetically

apmasell:


callunity:
Please also change the sampleClass sort in the EditSampleController (~L418)

apmasell:
Fixed

miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -220,10 +225,17 @@ Sample.hot = {
   },
 
   /**
+   * Returns true if a new sample of the provided SampleClass can be created without an existing parent
+   */
+  canCreateNew: function (sampleClass) {
```
djcooke: Shouldn't this now include some (not all) aliquots? sampleClass.canCreateNew probably replaces the need for this function though

apmasell: Oh, this was a bad rebase. I'll remove it.

PR #491, opened 2016-10-17
added KitDescriptor unique key, removed DetailedSample dao and service

djcooke:


callunity:
ticket number in PR name?

PR #493, opened 2016-10-18
GLT-1209: update SampleNumberPerProject when migrating Identities

djcooke:


callunity:
ticket number in PR name?

PR #495, opened 2016-10-18
drop unique constraint on external name

callunity:
NoTest because I was unable to find syntax to make HQL happy.


callunity:
I've just offended it further by mis-naming it, haven't I.

PR #497, opened 2016-10-19
GLT-1021: description field modifications

djcooke:
* Removed description field from
  * SamplePurpose
  * TissueMaterial
* Made description nullable for
  * Sample
  * Library
  * Pool
  * Run
  * Experiment
  * Platform
  * Subproject
  * Study
* Made Run alias NOT nullable

### DIFF
--- a/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
+++ b/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
@@ -381,7 +381,6 @@ public class DefaultSampleService implements SampleService {
   private SampleTissue createParentTissue(SampleTissue tissue, DetailedSample child) throws IOException {
     log.debug("Creating a new Tissue to use as a parent.");
     tissue.setProject(child.getProject());
-    tissue.setDescription("Tissue");
     tissue.setSampleType(child.getSampleType());
     tissue.setScientificName(child.getScientificName());
     tissue.setVolume(0D);
@@ -394,7 +393,6 @@ public class DefaultSampleService implements SampleService {
   private SampleStock createParentStock(SampleStock stock, DetailedSample child) throws IOException {
     log.debug("Creating a new Stock to use as a parent.");
     stock.setProject(child.getProject());
-    stock.setDescription("Stock");
     stock.setSampleType(child.getSampleType());
     stock.setScientificName(child.getScientificName());
     stock.setVolume(0D);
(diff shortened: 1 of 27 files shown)


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -381,7 +381,6 @@ private DetailedSample findOrCreateParent(DetailedSample sample) throws IOExcept
   private SampleTissue createParentTissue(SampleTissue tissue, DetailedSample child) throws IOException {
     log.debug("Creating a new Tissue to use as a parent.");
     tissue.setProject(child.getProject());
-    tissue.setDescription("Tissue");
```
callunity: Might it be worth keeping a description in here to indicate that these are ghost tissues/stocks?

djcooke: That's what the 'isSynthetic' field is for, and a big grey box explaining it shows up in the UI. The ghosts' old descriptions didn't mention that they were ghosts anyway - just the sample category, which again is available elsewhere

djcooke:
I think the difference between null and empty description doesn't really matter. I could remove the statements changing empty->null. There's no changelog entry when changing null->empty, and Pinery will see it the same either way

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
djcooke: Want to make a separate ticket for making better use of the 'isSynthetic' field? I don't think storing the same information in two places - one of which is freely-editable free text (description) - is the solution

callunity:
Ok, it looks like the issue I'm thinking of got fixed a while ago.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
callunity: I will make the ticket. I agree it's not the solution, or even a good solution, but I think adding "ghost" to the description is a user-friendly stopgap until we ultimately make better use of the 'isSynthetic' field.

miso-web/src/main/webapp/scripts/pool_ajax.js
```diff
@@ -52,8 +52,7 @@ var Pool = Pool || {
 
     // Description input field validation
     jQuery('#description').attr('class', 'form-control');
-    jQuery('#description').attr('data-parsley-required', 'true');
-    jQuery('#description').attr('data-parsley-maxlength', '100');
+    jQuery('#description').attr('data-parsley-maxlength', '255');
```
callunity: I've just noticed that Pool volume is also set to `required`, would you mind changing that in this PR?

callunity:
I extra approve this PR... (oops)

miso-web/src/main/webapp/scripts/pool_ajax.js
djcooke: changed

PR #499, opened 2016-10-19
Migration parent npe

apmasell:


migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -191,7 +191,7 @@ private void saveSample(Sample sample) throws IOException {
         if (detailed.getParent().getSampleClass() == null && detailed.getParent().getPreMigrationId() != null) {
           // find previously-migrated parent
           detailed.setParent((DetailedSample) serviceManager.getSampleDao().getByPreMigrationId(detailed.getParent().getPreMigrationId()));
-          if (detailed.getParent() == null) {
+          if (detailed.getParent() != null) {
```
djcooke: Was correct before the change. This will make it fail whenever a valid pre-migrated parent is found

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -376,6 +377,7 @@ private void saveLibraryChangeLog(Library library, Collection<ChangeLog> changes
   public void saveLibraryDilutions(final Collection<LibraryDilution> libraryDilutions) throws IOException {
     log.info("Migrating library dilutions...");
     for (LibraryDilution ldi : libraryDilutions) {
+      log.debug("Saving library dilution " + ldi.getName());
```
djcooke: Name will always be null before save

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
djcooke: A change is needed though to avoid the NPE (and actually throw the IOException) though. Either store `preMigrationId` in a separate variable before lookup, and use that for the log message, or `getByPreMigrationId` into a new variable and throw the exception if null before actually changing `detailed`'s parent

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
apmasell: Fixed.

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
apmasell: Fixed.

miso-web/src/main/webapp/pages/editExperiment.jsp
```diff
@@ -216,7 +216,7 @@
                    onMouseOut="this.className='dashboard'" class="dashboard"
                    ondblclick="Experiment.pool.experimentSelectPool(this);">
                       <span style="float:left">
-                        <b>${p.name}</b> (${fn:length(p.dilutions)} dilutions)
+                        <b>${p.name}</b> (${fn:length(p.poolableElements)} dilutions)
```
callunity: I think this is on the wrong branch...

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -376,6 +378,7 @@ private void saveLibraryChangeLog(Library library, Collection<ChangeLog> changes
   public void saveLibraryDilutions(final Collection<LibraryDilution> libraryDilutions) throws IOException {
     log.info("Migrating library dilutions...");
     for (LibraryDilution ldi : libraryDilutions) {
+      log.debug("Saving library dilution " + ldi.getPreMigrationId());
```
djcooke: preMigrationId is useful when using PinerySource, but may be null if using a MigrationSource other than Pinery (e.g. LoadGenerator), resulting in a lot of "Saving library dilution null."

Maybe use "Saving library dilution of library $libraryAlias" if preMigrationId is null, and "...formerly $preMigrationId" otherwise? Dilutions don't have any good identifier before save, but if there's an error, knowing the Library would help narrow it down. After save, the "formerly ..." bit could be left out if preMigrationId is null.

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -376,6 +378,11 @@ private void saveLibraryChangeLog(Library library, Collection<ChangeLog> changes
   public void saveLibraryDilutions(final Collection<LibraryDilution> libraryDilutions) throws IOException {
     log.info("Migrating library dilutions...");
     for (LibraryDilution ldi : libraryDilutions) {
+      String friendlyName = ldi.getName() == null ? " of " + ldi.getLibrary().getAlias() : ldi.getName();
```
apmasell: Right. Fixed.

PR #502, opened 2016-10-21
Move action columns from last to first in edit pool

apmasell:


tonydebat:
On the edit pool page the column headings need to be updated to match the new column contents.
image

apmasell:
Fixed and rebased.

PR #511, opened 2016-10-26
GLT-1233: replaced .settings symlinks with copies to avoid issues in Windows

djcooke:


djcooke:
(1 symlink removed + 4 settings files added) * 15 modules

PR #517, opened 2016-10-28
Pull bulk library creation tab back in.

frankie13:
A quick test and this seems to work good.
Unless @apmasell has other plans?


apmasell:
So, I have other plans, but only this code works for the “Add Libraries” menu from the Edit Project page.

PR #518, opened 2016-10-28
Make HoT work with plain samples and libraries

apmasell:


miso-web/src/main/webapp/pages/bulkEditLibraries.jsp
```diff
@@ -114,15 +114,12 @@
...
-        
-        if (Library.hot.propagateOrEdit == 'Propagate') {
-          Hot.fetchSampleOptions(Library.hot.makeBulkCreateTable);
-  	    } else {
-  	      Hot.fetchSampleOptions(Library.hot.makeBulkUpdateTable);
-	      }
+      Hot.saveButton.addEventListener('click', Library.hot.saveData, true);
+
+      if (Library.hot.propagateOrEdit == 'Propagate') {
+        Hot.fetchSampleOptions(Library.hot.makeBulkCreateTable);
```
callunity: `Hot.fetchSampleOptions` is unnecessary for plain samples

apmasell: Err, yes and no. There are many data paths that still want to access that data. It's much easier to have them access a list of empty values than it is to protect all of them with guards that return an empty list.

PR #519, opened 2016-10-28
create LibraryDesignCode table

callunity:


callunity:
Closed because Dillan figured out a more clever way of populating the table.

PR #520, opened 2016-11-01
GLT-1085: fix siblingNumber generation

djcooke:
also removes the option `miso.db.caching.mappers.enabled` which was never actually used for anything, and will not be useful with the switch-over to Hibernate

### DIFF
--- a/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSampleDao.java
+++ b/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSampleDao.java
@@ -125,13 +125,14 @@ public class HibernateSampleDao implements SampleDao, SiblingNumberGenerator {
 
   @Override
   public int getNextSiblingNumber(String partialAlias) throws IOException {
+    // Find highest existing siblingNumber matching this partialAlias
     Query query = currentSession().createQuery("select max(siblingNumber) from DetailedSampleImpl as ds"
             + " where alias IN (concat(:alias, ds.siblingNumber), concat(:alias, '0', ds.siblingNumber))");
     query.setString("alias", partialAlias);
     Number result = ((Number) query.uniqueResult());
     int next = result == null ? 0 : result.intValue();
 
-    // verify uniqueness and fix siblingNumbers for existing samples if necessary
+    // Increment and verify uniqueness. If alias is used, fix siblingNumber for existing sample. Repeat until unique
     Query verifyQuery = null;
     do {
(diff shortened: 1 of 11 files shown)


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSampleDao.java
```diff
@@ -123,13 +124,23 @@ public Long addSample(final Sample sample) throws IOException {
...
-    query.setLong("sampleClassId", childClass.getId());
+  public int getNextSiblingNumber(String partialAlias) throws IOException {
+    Query query = currentSession().createQuery("select max(siblingNumber) from DetailedSampleImpl as ds"
+            + " where alias IN (concat(:alias, ds.siblingNumber), concat(:alias, '0', ds.siblingNumber))");
+    query.setString("alias", partialAlias);
     Number result = ((Number) query.uniqueResult());
-    return result == null ? 1 : result.intValue() + 1;
+    int next = result == null ? 0 : result.intValue();
+
+    // verify uniqueness and fix siblingNumbers for existing samples if necessary
```
apmasell: This could use some explanation.

First, the scope of a sibling number is a tissue, right?

As I understand it, you find the highest sibling number that presently exists and then set the sibling number of the all the samples that share that prefix (i.e., the same tissue?) and any of their descendants who already have a higher sibling number? Yes?

apmasell: Okay. Put this description in the code. :-)

djcooke: Javadoc already exists for this method on the interface. Does this leave something unclear?

apmasell: Yes. It's doesn't explain the scoping and the conflict resolution and because it's in the interface, it doesn't describe this implementation.

djcooke: Updated javadoc/comments to clarify a bit. Did not explain scope because scope is not a real thing here, and there is nothing to suggest that it is. With our naming scheme, there is an imagined/effective scope, which is different depending on class and category, but sibling number is really just the lowest possible number that forms a unique alias. With another naming scheme, the imagined/effective scope could be completely different.

PR #521, opened 2016-11-03
Add wiring for HibernateSequencingParametersRunDAO

apmasell:


migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/MisoServiceManager.java
```diff
@@ -1376,4 +1379,26 @@ private void updatePoolQcDaoDependencies() {
...
+  public HibernateSequencingParametersDao getSequencingParametersDao() {
+    return sequencingParametersDao;
+  }
+
+  public void setSequencingParametersDao(HibernateSequencingParametersDao sequencingParametersDao) {
+    this.sequencingParametersDao = sequencingParametersDao;
+    updateSequencingParametersDao();
+  }
+
+  private void updateSequencingParametersDao() {
```
callunity: method should be `updateSequencingParametersDaoDependencies`

callunity: Another idiomatic change: `setDefaultRunDao()` should contain a line of `dao.setSequencingParametersDao(sequencingParametersDao)`

apmasell: Fixed.

PR #530, opened 2016-11-09
Copy required parts of the simlims package for MISO

apmasell:


froggleston:
Sublicensing this code under MISO is fine. All good with me.

PR #531, opened 2016-11-10
Glt 513 test securityprofile dao

apmasell:


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLSecurityProfileDAO.java
```diff
@@ -207,8 +201,8 @@ public int count() throws IOException {
   @Cacheable(cacheName = "securityProfileCache", keyGenerator = @KeyGenerator(name = "HashCodeCacheKeyGenerator", properties = {
       @Property(name = "includeMethod", value = "false"), @Property(name = "includeParameterTypes", value = "false") }))
   public SecurityProfile get(long id) throws IOException {
-    List results = template.query(PROFILE_SELECT_BY_ID, new Object[] { id }, new SecurityProfileMapper());
-    SecurityProfile sp = results.size() > 0 ? (SecurityProfile) results.get(0) : null;
+    List<SecurityProfile> results = template.query(PROFILE_SELECT_BY_ID, new Object[] { id }, new SecurityProfileMapper());
+    SecurityProfile sp = results.size() > 0 ? results.get(0) : null;
```
djcooke: This pattern is in a lot of places... Can we add `if (results.size() > 1) throw new IllegalStateException("...")` before this line?

sqlstore/src/test/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLSecurityProfileDAOTest.java
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  @Test
+  public void testCount() throws IOException {
+    assertTrue("No security profiles in DB", dao.count() > 0);
+  }
+
+  @Test
+  public void testGetByIdNone() throws IOException {
+    assertNull(dao.get(100L).getProfileId());
```
djcooke: Weird that the DAO's get method creates a security profile if it doesn't exist despite the method having an id parameter. I wonder if that is required by something... (observation, not a change request)

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLSecurityProfileDAO.java
apmasell: This code is about to be replaced next week and Hibernate will fix it.

apmasell: They can just use `uniqueResult` to do it.

PR #533, opened 2016-11-11
Infer LibraryDesignCode from alias when Source Template Type is null

callunity:
Old data is unreliable.


callunity:
PR for parsing the name : https://github.com/oicr-gsi/pinery-miso-migration/pull/76

djcooke:
The fixup cancels the actual changes that were made, and the exception message was clearer before (ValueTypeLookup doesn't and shouldn't care about the Library's alias). This PR should probably be closed

PR #535, opened 2016-11-14
GLT-1010 add PacBio dashboard link on Run page

callunity:


miso-web/src/main/webapp/scripts/run_ajax.js
```diff
@@ -155,6 +155,27 @@ var Run = Run || {
...
+    // today in format YYYYMMDD-000000
+    var todayString = "" + today.getFullYear() + zeroPad(today.getMonth() + 1, 2) + zeroPad(today.getDate(), 2) + "-" + zeroPad(0, 6);
+    var yesterday = new Date();
+    yesterday.setDate(yesterday.getDate() - 1);
+    var yesterdayString = "" + yesterday.getFullYear() + zeroPad(yesterday.getMonth() + 1, 2) + zeroPad(yesterday.getDate(), 2) + "-" + zeroPad(0, 6);
+    var url = pbDashboardUrl
+              + "?instrument=" + (instrumentName ? instrumentName : "")
+              + "&run=" + runName
+              + "&from=" + yesterdayString
+              + "&to=" + todayString;
```
djcooke: these url parameters result in not showing the correct run. I tried changing to the run's startDate for both `from` and `to` and it showed the correct run

callunity:
This somewhat works now, but either Karen or the Dashboard renames things such that MISO stores the run name containing the number of times the PacBio has run, and the Dashboard cuts off that part of the name. Closing this PR until we figure it out.

PR #539, opened 2016-11-15
*HOT* Auto Increment

frankie13:


miso-web/src/main/webapp/scripts/shared_hot.js
```diff
@@ -478,5 +478,127 @@ var Hot = {
...
       document.getElementById('saveSuccesses').classList.add('hidden');
     }
+  },
+  _newData : [],
+  incrementingAutofill : function(start, end, rows) {
+    var incrementingContentColumns = {};
+    var valid = true;
+    (function() {
+      // validation check
+      console.log(rows);
```
apmasell: Please delete debuggging output

miso-web/src/main/webapp/scripts/shared_hot.js
```diff
@@ -478,5 +478,127 @@ var Hot = {
...
+      var cols = rows[0];
+      for (var cols_i = 0; cols_i < cols.length; cols_i++) {
+        var cellContents = "" + cols[cols_i];
+        var numbersInCell = cellContents.match(/\d+/g);
+        var template = cellContents;
+        if (!numbersInCell) {
+          break;
+        }
+        for (var placeHolder_i = 0; placeHolder_i < numbersInCell.length; placeHolder_i++) {
+          template = template.replace(numbersInCell[placeHolder_i], "{"
```
apmasell: I'm very confused by this. `match` in this case should only return element, right? There are no capture groups in the regex. If there were, this code would be very fragile because if you named something: `A0001_0001`, it would get changed to `A{0}_{0}`, rather than `A{0}_{1}`.

apmasell:
Also, please don't merge develop into your branch. Rebase on top of develop.

miso-web/src/main/webapp/scripts/shared_hot.js
frankie13: The regex is designed to be iterative with /g, so it will return an array of matches.
The placeholder value is based on the array index of each value so you don't get repetition.

callunity:
Is this supposed to auto-increment numbers when the cell containing numbers is double-clicked or dragged down? I tried both of these methods and got the standard behaviour of just copying the cell contents (see Description and Matrix Barcode columns in attached image). 

Does this implementation allow turning off the autofill? For instance, if the description should contain a consistent number then autofilling would be undesirable.

autoinc

callunity:
(Sorry, overly-zealous comment deletion there.)

I'm still finding the default functionality even if my data is in the first row (see attached).

Additionally, it's going to need to be possible to toggle this on and off as there will be some cases where something like a description will contain a number that shouldn't be auto-incremented.

autoinc

callunity:
When auto-filling down 96 rows with numbers (no letters) it first auto-fills it with the default behaviour, and then takes nearly 30 seconds to complete the autofill and change the values. Can it be sped up?

miso-web/src/main/webapp/scripts/shared_hot.js
apmasell: Okay. Please document that. It's not obvious.

frankie13:
I believe I have addressed all your comments / bugs now.
Please take another look 🎉

frankie13:
ok, all comments addressed!

miso-web/src/main/webapp/pages/bulkEditDilutions.jsp
```diff
@@ -53,6 +53,8 @@
     <p>To fill all empty cells in a column with the value of your selected cell(s), <b>double-click</b> the square in the bottom right of your selected cell(s).
       <br/>To fill a variable number of columns with the value of your selected cell,  <b>click</b> the square in the bottom right of your 
       filled-in selected cell and <b>drag</b> up or down. All selected columns will be filled in.
+      <br/>To fill down a column with values following an incremental (+1) pattern, either select two adjacent cells in a column and then drag
```
callunity: @djcooke made a good point with the wording - could you please move the "either" a few words over, to "and then either drag down,..."? /neckbeard

PR #544, opened 2016-11-16
GLT-1168 add bulk actions after bulk save

callunity:


miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -1405,7 +1406,83 @@ Sample.hot = {
...
+    if (Hot.detailedSample) {
+      Sample.sampleClasses = Hot.sampleOptions.sampleClassesDtos;
+      Sample.validRelationships = Hot.sampleOptions.sampleValidRelationshipsDtos;
+      var uniqueParentScAliases = Sample.ui.getUniqueValues(
+          Hot.startData.map(function (sam) { return sam.sampleClassAlias; })
+      );
+      var uniqueParentSCs = Sample.sampleClasses.filter(function (sc) { return uniqueParentScAliases.indexOf(sc.alias) != -1; });
+      var validChildren = Sample.ui.getChildSampleClasses(uniqueParentSCs);
+      if (validChildren.length > 0) {
+        actions.push('<a style="cursor:pointer" href="javascript:void(0);">Propagate samples:</a>');
```
apmasell: Given we have `Propagate libraries` why not have these all as `Propagate to X`? Does it look terrible?

callunity:
Associated protocol: https://wiki.oicr.on.ca/display/MISO/LIMS+Entry%3A+Propagating+Aliquots+to+Library+Dilutions

miso-web/src/main/webapp/scripts/sample_hot.js
callunity: fixed

PR #549, opened 2016-11-18
GLT-1322 make TargetedResequencing user editable

callunity:


miso-web/src/main/webapp/scripts/library_ajax.js
```diff
@@ -247,69 +247,82 @@ Library.dilution = {
...
+            'url': ajaxurl,
+            'libraryPrepKitId': libraryPrepKitId
+          },
+          {
+            'doOnSuccess': function (json) {
+  	        	var selectElem = jQuery('#libraryDilutionTargetedResequencing');
+  	        	if (json.targetedResequencings.length == 0) {
+  	        		selectElem.append(new Option('NONE', 0));
+  	        	} else {
+  	        		selectElem.append(new Option('--- select'));
```
apmasell: I think it's reasonable to have a default of none, no matter what, right?

miso-web/src/main/webapp/scripts/library_ajax.js
```diff
@@ -247,69 +247,82 @@ Library.dilution = {
...
+  	        		selectElem.append(new Option('NONE', 0));
+                json.targetedResequencings.sort(function(a, b) {
+                   return (a.alias > b.alias) - (a.alias < b.alias);
```
apmasell: For strings, use `a.localeCompare(b)`.

PR #550, opened 2016-11-22
Add edit sample breadcrumbs back in.

frankie13:
Lab requested.


djcooke:
This code isn't missing from develop: https://github.com/TGAC/miso-lims/blob/develop/miso-web/src/main/webapp/pages/editSample.jsp#L77

If it's missing in your deployment, might it be a merge error with your config branch?

frankie13:
@djcooke sadly it doesn't appear on tab-2: https://github.com/TGAC/miso-lims/blob/develop/miso-web/src/main/webapp/pages/editSample.jsp#L738

djcooke:
Ohhhh... Is there a case where tab-2 is used that sample.project is not null on page load?

frankie13:
it seems so, if you go in from Project > Add Samples you get taken straight to tab-2

djcooke:
Ah, there it is :+1:

PR #553, opened 2016-11-22
GLT-1331 always display some platform tabs in Pools page

callunity:
display all platformTypes if no squencers are present in db (friendlier for new people trying out MISO)


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/ListPoolsController.java
```diff
@@ -74,7 +74,12 @@ public void setRequestManager(RequestManager requestManager) {
 
   @ModelAttribute("platformTypes")
   public Collection<String> populatePlatformTypes() throws IOException {
-    return PlatformType.platformTypeNames(requestManager.listActivePlatformTypes());
+    Collection<PlatformType> platforms = requestManager.listActivePlatformTypes();
+    if (platforms.size() > 0) {
+      return PlatformType.platformTypeNames(requestManager.listActivePlatformTypes());
```
djcooke: Can `return PlatformType.platformTypeNames(platforms);` instead of doing the same lookup twice

PR #554, opened 2016-11-23
GLT-1010 add PacBio dashboard link on Run page

callunity:


callunity:
You're telling me. It took @djcooke and myself an hour or two to figure out exactly which params it needed, and in what order. The best part was when the formats that let me load it on my computer threw a 404 when loading the same site on his computer. And vice versa.

PR #558, opened 2016-11-24
Speed up bulk add samples

frankie13:
This small change to using criteria speeds up bulk sample save from 25~30 secs per sample to <1 sec per sample.

It also removes some white space.


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSampleDao.java
```diff
@@ -358,25 +358,23 @@ public boolean aliasExists(String alias) throws IOException {
...
    *          projectId
    * @return Collection<Sample> samples
    */
   @Override
   public Collection<Sample> listByProjectId(long projectId) throws IOException {
-    Query query = currentSession().createQuery("from SampleImpl where project.id like :id");
-    query.setLong("id", projectId);
     @SuppressWarnings("unchecked")
-    List<Sample> records = query.list();
+    List<Sample> records = currentSession().createCriteria(SampleImpl.class).add(Restrictions.like("project.id", projectId)).list();
```
frankie13: now you mention it, using like is unnecessary - as evidenced by the lack of %'s.  I shall remove.

PR #559, opened 2016-11-24
GLT-1339 rename TargetedResequencing to TargetedSequencing

callunity:


sqlstore/src/main/resources/db/migration/V8200__tarReseq_to_tarSeq.sql
```diff
@@ -0,0 +1,14 @@
+--StartNoTest because FK was never named in the original migration
+SELECT CONSTRAINT_NAME INTO @constraint
+    FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE
+    WHERE TABLE_NAME = 'LibraryDilution' AND COLUMN_NAME = 'targetedResequencingId'
+    AND REFERENCED_TABLE_NAME = 'TargetedResequencing' AND REFERENCED_COLUMN_NAME = 'targetedResequencingId';
+SET @q = CONCAT('ALTER TABLE LibraryDilution DROP FOREIGN KEY ',@constraint);
```
apmasell: Nice.

callunity: @djcooke deserves all the credit on this one.

miso-web/src/main/resources/ehcache.xml
```diff
@@ -70,10 +70,10 @@
       retryAttempts="0" retryAttemptDelaySeconds="0"/> -->
   </cache>
   
-  <cache name="targetedResequencingCache" eternal="true" overflowToDisk="true" statistics="true" maxBytesLocalHeap="30M" maxBytesLocalDisk="100M">
+  <cache name="targetedSequencingCache" eternal="true" overflowToDisk="true" statistics="true" maxBytesLocalHeap="30M" maxBytesLocalDisk="100M">
```
djcooke: These caches don't appear to be used at all, and they shouldn't be... Right? If you want to remove them, there are also caches defined for Plates in here still too

PR #561, opened 2016-11-25
Fix Sample Navigation Arrows

frankie13:
Turns out the project was lazy loaded and so getSamples() just returned an empty collection.
This change checks to see if there are more samples in the project and, if not, tries to eagerly load the project to get the full object graph.

No adjacent samples means no navigation arrows.


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
```diff
@@ -157,6 +158,9 @@
   @Autowired
   private SampleValidRelationshipService sampleValidRelationshipService;
 
+  @Autowired
+  private ProjectStore projectStore;
```
apmasell: I think the controller is only supposed to access the service layer/RequestManager.

frankie13:
@apmasell d'oi.  Good point.  Fixed and tested.

PR #570, opened 2016-12-02
Fix library report

frankie13:
So creating the library report never seemed to end as it was trying to list every single library in the database.  We have ~26,000, so along with the mapping that was taking a while.
I changed it so it queries based on the created date of the library.  It always did this anyway but the logic was in the controller to sift though all 26K libraries, now it is in the DAO.


miso-web/src/main/webapp/WEB-INF/log4j.miso.properties
```diff
@@ -83,3 +83,5 @@ log4j.logger.notificationFileLogger=DEBUG, notificationFileAppender
 log4j.additivity.notificationFileLogger=false
 log4j.logger.limsFileLogger=DEBUG, limsFileAppender
 log4j.additivity.limsFileLogger=false
+
+log4j.logger.uk.ac.bbsrc.tgac.miso.webapp.controller.EditSampleController=ALL, stdout
```
frankie13: Opps :SBE:

PR #571, opened 2016-12-02
Glt 551 naming refactor

djcooke:
Note that this patch requires a change in miso.properties settings
* To maintain the default functionality from before this patch, you can either remove all naming configuration (anything beginning with `miso.naming`), or include only `miso.naming.scheme:default`. If you are using the miso.properties within the generated .war, it has been adjusted to use the defaults.
* To configure all of the OICR naming scheme, include `miso.naming.scheme:oicr` and remove all `miso.naming` settings

Documentation of all options in the Maintainer's Guide will follow


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/generation/DefaultLibraryAliasGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Override
+  public String generate(Library library) throws MisoNamingException {
+    if (library.getSample() != null) {
+      Pattern samplePattern = Pattern.compile("([A-z0-9]+)_S([A-z0-9]+)_(.*)");
+      Matcher m = samplePattern.matcher(library.getSample().getAlias());
+
+      if (m.matches()) {
+        try {
+          int numLibs = requestManager.listAllLibrariesBySampleId(library.getSample().getId()).size();
```
apmasell: Maybe file a ticket to add a count method to make this less expensive.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/generation/DefaultNameGenerator.java
```diff
@@ -0,0 +1,29 @@
...
+import uk.ac.bbsrc.tgac.miso.core.service.naming.DefaultMisoEntityPrefix;
+import uk.ac.bbsrc.tgac.miso.core.util.LimsUtils;
+
+public class DefaultNameGenerator implements NameGenerator<Nameable> {
+
+  private static final Logger log = Logger.getLogger(DefaultNameGenerator.class);
+
+  @Override
+  public String generate(Nameable nameable) throws MisoNamingException {
+    if (DefaultMisoEntityPrefix.get(nameable.getClass().getSimpleName()) == null) {
```
apmasell: Please file a ticket to replace this with a different interface and reference it here. This should be something like:
```
for (DefaultMisoEntity prefix : DefaultMisoEntity.values()) {
  if (prefix.getEntityClass().isAssignableFrom(nameable.getClass())) { return prefix.getEntityClass().name() + nameable.getId(); }
}
throw...
}
```

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/generation/OicrLibraryAliasGenerator.java
```diff
@@ -0,0 +1,51 @@
...
+  private static final Pattern samplePattern = Pattern.compile(sampleRegex);
+
+  @Override
+  public String generate(Library library) throws MisoNamingException {
+    if (library.getSample() != null) {
+      Matcher m = samplePattern.matcher(library.getSample().getAlias());
+
+      if (m.matches()) {
+        // try {
+        // int numLibs = requestManager.listAllLibrariesBySampleId(l.getSample().getId()).size();
```
callunity: Are these (and the catch below) kept here for a particular purpose?

djcooke: This is the library alias generation code more-or-less copy/pasted from what existed before. I did not fix the generator, which didn't work before. I'm not sure if the code that's commented out is useful. I will make sure there is a ticket open for getting Library alias generation working though

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/generation/DefaultNameGenerator.java
djcooke: GLT-1350 created

PR #580, opened 2016-12-07
Glt 551 docs

djcooke:
* updated docs for naming scheme
* fixup for OicrSampleAliasValidator - removes explicit Tissue Origins and Tissue Types


djcooke:
@frankie13: It is excessive, but the html files are generated. Modifying the admin manual is done through `2016-01-11-admin-manual.md`

PR #583, opened 2016-12-08
GLT-1353 fix bulk identity save

callunity:


miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -1198,7 +1205,7 @@ Sample.hot = {
         sample.concentration = obj.concentration;
       }
     } catch (e) {
-      console.log(e.error);
+      console.log(e);
```
callunity: Now we'll get useful error messages when things fail during the buildDtos step!

PR #584, opened 2016-12-08
fix validation for samples with nonStandardAlias

djcooke:


miso-web/src/main/webapp/scripts/sample_ajax.js
```diff
@@ -95,8 +95,12 @@ var Sample = Sample || {
     jQuery('#alias').attr('class', 'form-control');
     jQuery('#alias').attr('data-parsley-maxlength', '100');
     jQuery('#alias').attr('data-parsley-pattern', Utils.validation.sanitizeRegex);
-    jQuery('#alias').attr('data-parsley-sample-alias', '');
-    jQuery('#alias').attr('data-parsley-debounce', '500');
+    if (skipAliasValidation) {
+      jQuery('#alias').attr('data-parsley-required', 'true');
```
frankie13: cool, so forces the alias to be non-blank. 👍

PR #585, opened 2016-12-08
added project shortName validators

djcooke:
Project shortName validators:
* 'allowany': optional field; default option for 'default' naming scheme
* 'oicr': required field; must be 3-5 characters, mix of CAPS and numbers; default option for 'oicr' naming scheme


djcooke:
@callunity: fixed

PR #590, opened 2016-12-12
Convert Study to Hibernate and add StudyType model

apmasell:
JIRA: GLT-1283


miso-web/src/main/webapp/WEB-INF/db-config.xml
```diff
@@ -54,6 +55,7 @@
         <value>uk.ac.bbsrc.tgac.miso.core.data.impl.LibraryAdditionalInfoImpl</value>
         <value>uk.ac.bbsrc.tgac.miso.core.data.impl.PlatformImpl</value>
         <value>uk.ac.bbsrc.tgac.miso.core.data.impl.PoolOrderImpl</value>
+        <value>uk.ac.bbsrc.tgac.miso.core.data.PoolOrderCompletion</value>
```
callunity: Please move higher for alphabetical ordering.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/StudyType.java
```diff
@@ -0,0 +1,28 @@
+package uk.ac.bbsrc.tgac.miso.core.data;
+
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "StudyType")
+public class StudyType {
+  private long typeId;
```
callunity: Add `@Id` annotation

apmasell:
Fixed and added missing SQL code.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/StudyImpl.java
```diff
@@ -77,7 +122,229 @@ public StudyImpl(Project project, User user) {
...
+   * @param user
+   *          of type User
+   */
+  public StudyImpl(User user) {
+    setSecurityProfile(new SecurityProfile(user));
+  }
+
+  @Override
+  public void addExperiment(Experiment e) throws MalformedExperimentException {
+    // do experiment validation
```
djcooke: Or don't?

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLTgacSubmissionDAO.java
```diff
@@ -359,10 +359,6 @@ public SubmissionImpl mapRow(ResultSet rs, int rowNum) throws SQLException {
 
       try {
         // process submittables
-        for (Study study : studyDAO.listBySubmissionId(rs.getLong("submissionId"))) {
```
djcooke: reason for removing this?

sqlstore/src/main/resources/db/migration/V8900__hibernatize_everything.sql
```diff
@@ -1 +1,7 @@
-UPDATE Platform SET name = UPPER(name);
\ No newline at end of file
+UPDATE Platform SET name = UPPER(name);
+
+INSERT INTO StudyType(name) SELECT DISTINCT studyType AS name FROM Study WHERE studyType NOT IN (SELECT name FROM StudyType);
```
djcooke: Since StudyTypes are SRA-defined, it would be better to ensure that the full set is already in the DB, adding missing ones, and not assume that the existing values are good. Migration should fail if there is bad data that does not conform

apmasell: Okay. Fixed this and the comment.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLTgacSubmissionDAO.java
djcooke: This seems like a method that did and should still exist in the Study DAO?

PR #591, opened 2016-12-12
Use Hibernate for Sequencer Reference

callunity:


sqlstore/src/main/resources/db/migration/V8900__hibernatize_everything.sql
```diff
@@ -1,3 +1,13 @@
 UPDATE KitDescriptor SET kitType = UPPER(kitType), platformType = UPPER(platformType);
 
-UPDATE Platform SET name = UPPER(name);
\ No newline at end of file
+UPDATE Platform SET name = UPPER(name);
+
+ALTER TABLE SequencerReference ADD COLUMN ip VARCHAR(50) NOT NULL DEFAULT 'localhost';
+-- H2 doesn't have INET_NTOA function
+-- StartNoTest
+UPDATE SequencerReference tabl SET ip = (
```
apmasell: You should be able to do just `UPDATE SequencerReference SET ip = INET_NTOA(ipAddress);`

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
```diff
@@ -84,27 +114,12 @@ public Platform getPlatform() {
...
-  }
-
-  @Override
-  public void checkAvailability(int timeout) throws IOException {
-    this.available = getIpAddress() == null ? false : getIpAddress().isReachable(timeout);
-  }
-
-  @Override
-  public void setIpAddress(InetAddress ip) {
+  public void setIpAddress(String ip) {
```
djcooke: Maybe validate the IP address in the setter? Otherwise this can be anything and will need to be validated every time it's used, before getFQDN() is called, etc

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencerReference.java
```diff
@@ -31,14 +32,30 @@
 import org.codehaus.jackson.map.annotate.JsonSerialize;
 
 /**
- * A SequencerReference is a {@link HardwareReference} specifically designated with a {@link Platform}
+ * A SequencerReference is a {@link SequencerReference} specifically designated with a {@link Platform}
```
djcooke: This implies the possibility of a SequencerReference that is not a SequencerReference :grinning:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencerReference.java
```diff
@@ -54,10 +71,43 @@
...
+   */
+  void setIpAddress(String ip);
+
+  /**
+   * Returns the String ipAddress of this SequencerReference object.
+   * 
+   * @return String ipAddress.
+   * @throws UnknownHostException
+   */
+  String getIpAddress() throws UnknownHostException;
```
djcooke: Remove throws clause since this doesn't actually get thrown

miso-web/src/main/webapp/pages/viewStats.jsp
```diff
@@ -53,7 +53,6 @@
           <th>Health</th>
           <th>Start Date</th>
           <th>Last Updated</th>
-          <th class="fit">View Stats</th>
```
djcooke: I think the stuff removed here is something EI uses (statsdb configurable in miso.properties). Check with @frankie13 / @froggleston before removing

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSequencerReferenceDao.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+    return id;
+  }
+
+  @Override
+  public SequencerReference get(long id) throws IOException {
+    SequencerReference sr = (SequencerReference) currentSession().get(SequencerReferenceImpl.class, id);
+    SequencerReference preUpgradeSR = getByUpgradedReference(id);
+    if (preUpgradeSR != null) {
+      sr.setPreUpgradeSequencerReference(preUpgradeSR);
```
djcooke: Consider making this relationship unidirectional instead of having to look up the preUpgradeSequencerReference. It is usually not useful anyway (only on edit sequencer page I'd think)

sqlstore/src/test/resources/db/test_migration/V9000__miso_test_data.test.sql
```diff
@@ -418,9 +418,9 @@ VALUES (1,1),(1,2),(1,3),(1,4),(1,5),(1,6),(1,7),(1,8),(2,9),(2,10),(2,11),(2,12
 (3,20),(3,21),(3,22),(3,23),(3,24),(4,25),(4,26),(4,27),(4,28),(4,29),(4,30),(4,31),(4,32);
 
 DELETE FROM `SequencerReference`;
-INSERT INTO `SequencerReference`(`referenceId`, `name`, `ipAddress`, `platformId`, `available`) 
-VALUES (1,'h1179',X'0F000001',16,1),
-(2,'h1180',X'0F000001',16,1);
+INSERT INTO `SequencerReference`(`referenceId`, `name`, `ip`, `platformId`)
+VALUES (1,'h1179','localhost',16),
```
djcooke: Field named ip shouldn't contain a hostname. Change value or fieldname to match
ip=`127.0.0.1` or hostname=`localhost`

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
```diff
@@ -23,34 +23,64 @@
...
-  private Boolean available;
-  private InetAddress ip;
+
+  private String ip;
+
   private String serialNumber;
   private Date dateCommissioned;
   private Date dateDecommissioned = null;
+
+  @OneToOne(targetEntity = SequencerReferenceImpl.class)
```
djcooke: The current model/dao will cause an exception if this isn't 1:1, but it could be ManyToOne, and it wouldn't exactly be wrong. If the relationship is made unidirectional (by removing preUpgradSequencerReference), this should be ManyToOne; otherwise, OneToOne is probably more correct

miso-web/src/main/webapp/pages/viewStats.jsp
callunity: Hmm I seem to have been overzealous here. I'll re-add the stats in this file.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
callunity: Removing preUpgradedSequencerReference but leaving it as OneToOne because it is not currently possible to upgrade a sequencer and then upgrade it again (and upgrading should be the only the the name changes).

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
```diff
@@ -84,27 +112,22 @@ public Platform getPlatform() {
...
+  public void setIpAddress(String ip) {
+    if (ip == null) {
+      this.ip = null;
+    } else {
+      try {
+      InetAddress inet = InetAddress.getByName(ip);
+      this.ip = (inet != null ? inet.getHostAddress() : null);
+      } catch (IOException e) {
+        log.error("Error getting InetAddress from given ip " + ip, e);
+        this.ip = null;
```
djcooke: rethrow as IllegalArgumentException

miso-web/src/main/webapp/pages/viewStats.jsp
```diff
@@ -114,6 +114,7 @@
             <th>Health</th>
             <th>Start Date</th>
             <th>Last Updated</th>
+            <th>View Stats</th>
```
djcooke: re-added to wrong table

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerReference.java
djcooke: Please add a unique constraint to SequencerReference.upgradedSequencerReferenceId to enforce this

callunity:
Fixing up conflicting file: `sqlstore/src/test/resources/db-test-context.xml`

PR #600, opened 2016-12-15
Use Hibernate for Run, RunQC, Status

callunity:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractRun.java
```diff
@@ -81,39 +90,59 @@
...
+
+  @Column(nullable = false)
   private Boolean pairedEnd;
   private Integer cycles;
   private String filePath;
 
-  private Date lastUpdated;
-
   @OneToOne(targetEntity = StatusImpl.class, cascade = CascadeType.ALL)
   private Status status;
```
djcooke: Could make this a true OneToOne by adding `Status.runId` as its new primary key and dropping `Status.statusId` and `Run.status_statusId`. If you want. :grin:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/RunImpl.java
```diff
@@ -50,10 +52,12 @@
  * @author Rob Davey
  * @since 0.1.0
  */
+@Entity
+@Table(name = "Run")
 public class RunImpl extends AbstractRun implements Serializable {
   protected static final Logger log = LoggerFactory.getLogger(RunImpl.class);
 
-  @OneToMany(cascade = CascadeType.ALL)
+  @ManyToMany(targetEntity = SequencerPartitionContainerImpl.class, cascade = CascadeType.ALL)
```
djcooke: This works without specifying the join table?

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateRunDao.java
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  @Override
+  public long save(Run run) throws IOException {
+    updateWatchers(run);
+    long id;
+    if (run.getId() == AbstractRun.UNSAVED_ID) {
+      currentSession().save(run);
+      try {
+        String name = namingScheme.generateNameFor(run);
```
djcooke: Name generation and validation should be moved to service layer

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateRunDao.java
```diff
@@ -0,0 +1,386 @@
...
+    return records;
+  }
+
+  @Override
+  public Run getByAlias(String alias) throws IOException {
+    Criteria criteria = currentSession().createCriteria(RunImpl.class);
+    criteria.add(Restrictions.eq("alias", alias));
+    @SuppressWarnings("unchecked")
+    List<Run> records = criteria.list();
+    if (records.size() == 1) {
```
djcooke: Can return uniqueResult instead of checking

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateRunDao.java
```diff
@@ -0,0 +1,386 @@
...
+    }
+  }
+
+  @Override
+  @Deprecated
+  public List<Run> listByExperimentId(long experimentId) throws IOException {
+    return Collections.emptyList();
+  }
+
+  // TODO: confirm that this works as intended
```
djcooke: Make sure the methods marked with this have decent tests and then remove the TODO tags. Tests should fail if it doesn't work as intended

sqlstore/src/test/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLRunDAOTest.java
```diff
@@ -400,11 +400,11 @@ public void testSaveAll() throws MisoNamingException, IOException {
 
     CacheManager cacheManager = Mockito.mock(CacheManager.class);
     Mockito.when(cacheManager.getCache(Matchers.anyString())).thenReturn(null);
-    dao.setCacheManager(cacheManager);
+    // dao.setCacheManager(cacheManager);
```
djcooke: :fire:

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateRunDao.java
```diff
@@ -0,0 +1,386 @@
...
+    Criteria criteria = currentSession().createCriteria(RunImpl.class);
+    return ((Long) criteria.setProjection(Projections.rowCount()).uniqueResult()).intValue();
+  }
+
+  @Override
+  public boolean remove(Run run) throws IOException {
+    if (run.isDeletable()) {
+      currentSession().delete(run);
+
+      Run testIfExists = get(run.getId());
```
apmasell: This check is out of order. You would have to do the get before the delete.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateRunDao.java
```diff
@@ -0,0 +1,386 @@
...
+    criteria.createAlias("containers", "spc");
+    criteria.add(Restrictions.eq("spc.id", containerId));
+    criteria.addOrder(Order.desc("run.id"));
+    criteria.setMaxResults(1);
+    return (Run) criteria.uniqueResult();
+  }
+
+  private static final String[] searchProperties = new String[] { "name", "alias", "description" };
+
+  private Criterion searchRestrictions(String querystr) {
```
apmasell: You can use `criteria.add(DbUtils.searchRestrctions(querystr, "name", "alias", "description")`.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateRunDao.java
```diff
@@ -0,0 +1,386 @@
...
+    criteria.addOrder(Order.desc("id"));
+    criteria.setMaxResults((int) limit);
+    @SuppressWarnings("unchecked")
+    List<Run> records = criteria.list();
+    return records;
+  }
+
+  // TODO: confirm that this works as intended, and isn't horribly slow
+  @Override
+  public int[] saveAll(Collection<Run> runs) throws IOException {
```
apmasell: I think this is a bug to not return a `long[]`... Perhaps file a ticket.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateRunQcDao.java
```diff
@@ -0,0 +1,173 @@
...
+    Criteria criteria = currentSession().createCriteria(QcType.class);
+    criteria.add(Restrictions.eq("qcTarget", "Run"));
+    @SuppressWarnings("unchecked")
+    Collection<QcType> records = criteria.list();
+    return records;
+  }
+
+  public List<Partition> listPartitionSelectionsByRunQcId(long runQcId) {
+    Criteria criteria = currentSession().createCriteria(RunQCImpl.class);
+    criteria.add(Restrictions.eqOrIsNull("id", runQcId));
```
callunity: Hmm apparently my Eclipse autocompletes to `eqOrIsNull` instead of plain `eq`.

sqlstore/src/main/resources/db/migration/V8900__hibernatize_everything.sql
```diff
@@ -7,3 +7,26 @@ UPDATE Study SET studyTypeId = (SELECT typeId FROM StudyType WHERE name = studyT
 ALTER TABLE Study ADD CONSTRAINT study_studyTypeId FOREIGN KEY (studyTypeId) REFERENCES StudyType(typeId);
 ALTER TABLE Study DROP COLUMN studyType;
 ALTER TABLE Study ALTER COLUMN studyTypeId bigint(20) NOT NULL;
+
+ALTER TABLE Status ADD FOREIGN KEY (instrumentName) REFERENCES SequencerReference (name);
+ALTER TABLE Status ADD FOREIGN KEY (runName) REFERENCES Run (alias);
```
callunity: Not without refactoring the notification servers -- there are several instances of  `requestManager.getSequencerReferenceByName(r.getStatus().getInstrumentName());`

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateRunDao.java
callunity: Ticket filed: https://jira.oicr.on.ca/browse/GLT-1366

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractRun.java
callunity: Made a ticket for later: https://jira.oicr.on.ca/browse/GLT-1367 :)

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateRunQcDao.java
callunity: I don't think it's possible to query for the partitions directly -- RunQC to Partition is a unidirectional relationship.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/RunImpl.java
callunity: Probably not... fixed.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateRunQcDao.java
```diff
@@ -0,0 +1,161 @@
...
+  @Override
+  public Collection<QcType> listAllRunQcTypes() throws IOException {
+    Criteria criteria = currentSession().createCriteria(QcType.class);
+    criteria.add(Restrictions.eq("qcTarget", "Run"));
+    @SuppressWarnings("unchecked")
+    Collection<QcType> records = criteria.list();
+    return records;
+  }
+
+  public List<Partition> listPartitionSelectionsByRunQcId(long runQcId) {
```
djcooke: This method can actually be removed now. It was only used by the RowMapper

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
```diff
@@ -1520,7 +1530,31 @@ public long saveProjectOverviewNote(ProjectOverview overview, Note note) throws
...
+      long id;
+      if (run.getId() == AbstractRun.UNSAVED_ID) {
+        runStore.save(run);
+        try {
+          String name = namingScheme.generateNameFor(run);
+          run.setName(name);
+
+          validateNameOrThrow(run, namingScheme);
+          Number newId = (long) runStore.save(run);
+          if (newId.longValue() != run.getId()) {
```
djcooke: These will never be equal because `run.id` was never changed. The check is unnecessary now though. A similar check was used previously because the generated name was based on a predicted ID before the Run was actually saved. Since it is now saved first, the ID is solid.

Above (L1535) should be changed to `run.setId(runStore.save(run));` though

sqlstore/src/test/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLRunDAOTest.java
```diff
@@ -135,11 +133,7 @@ public void testListAllWithBiggerLimit() throws IOException {
     assertEquals(4, runs.size());
   }
 
-  @Test
-  public void testListAllWithZeroLimit() throws IOException {
-    List<Run> runs = dao.listAllWithLimit(0L);
-    assertEquals(0, runs.size());
-  }
+  // testListAllWithZeroLimit() deleted because our version of Hibernate (4.2.21.FINAL) treats limit 0 as "no limit" (not fixed until 4.3.2)
```
djcooke: Test should not be removed, especially since behaviour may change again if Hibernate version changes
* If that's the expected and acceptable behaviour, the test should be modified to pass under those conditions
* If the method should still return an empty list, the DAO method should be modified to include `if (limit == 0) return new ArrayList<Run>();` or similar

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
```diff
@@ -1500,7 +1511,24 @@ public long saveProjectOverviewNote(ProjectOverview overview, Note note) throws
   @Override
   public long saveRun(Run run) throws IOException {
     if (runStore != null) {
-      return runStore.save(run);
+      long id;
+      if (run.getId() == AbstractRun.UNSAVED_ID) {
+        run.setId(runStore.save(run));
```
djcooke: This will actually fail due to null name. Should generate a temporary name before initial save, like in `DefaultSampleService.create(Sample)`

PR #602, opened 2016-12-19
GLT-1365: stored procedures to add value-type data if missing

djcooke:


djcooke:
These should probably get moved to the afterMigrate script instead... Will update

djcooke:
Moved to afterMigrate and fixed error migrating

PR #603, opened 2016-12-20
Create procedures to delete sample or library

callunity:
These procedures will be added to production and run before the next release.


callunity:
changelogs deleted too

PR #604, opened 2016-12-20
removed unused things

djcooke:
Includes classes concerning Request, Activity, and Protocol


djcooke:
Will merge into hibernate instead since these changes are required for Hibernatizing Note

PR #605, opened 2016-12-21
Remove request, protocol, activity, workflow

djcooke:


djcooke:
Replaces https://github.com/TGAC/miso-lims/pull/604 - rebased onto hibernate, 2 commits added

PR #606, opened 2016-12-21
Use Hibernate for TargetedSequencing

callunity:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/TargetedSequencing.java
```diff
@@ -2,40 +2,64 @@
...
+@Table(name = "TargetedSequencing")
 public class TargetedSequencing {
 
   public static final Long UNSAVED_ID = 0L;
 
+  @Id
+  @GeneratedValue(strategy = GenerationType.AUTO)
   private Long targetedSequencingId;
 
   /** Name displayed in UI and associated with targeted sequencing bed file. */
```
djcooke: Out of scope, but turning these comments into Javadocs on the getters and setters would be nice :grin:

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateTargetedSequencingDao.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Autowired
+  private SessionFactory sessionFactory;
+
+  private Session currentSession() {
+    return getSessionFactory().getCurrentSession();
+  }
+
+  @Override
+  public long save(TargetedSequencing ts) throws IOException {
```
djcooke: Save method is never used, and with our policy to add via migrations, it should probably be removed

PR #607, opened 2016-12-21
GLT-1378: fix notification consumer config

djcooke:


miso-web/pom.xml
```diff
@@ -215,6 +215,11 @@
       <artifactId>miso-service</artifactId>
     </dependency>
     <dependency>
+      <!-- Do not remove; used via SPI -->
```
callunity: :+1:

PR #608, opened 2016-12-22
GLT-1285: Updated Watcher management for Hibernate DAOs

djcooke:


djcooke:
@apmasell: The problem with that is that Run manages the relationship. If the RunWatchers group members are added to the same collection as individual watchers for a specific Run, it's very easy to accidentally add all RunWatchers as individual watchers. Removing a person from the RunWatchers group would then not have the complete intended effect

PR #610, opened 2017-01-03
GLT-1380 allow users to delete barcodes

callunity:


callunity:
Fixed to remove the second mode of barcode deletion.

PR #611, opened 2017-01-03
Use Hibernate for SequencerPartitionContainer and SequencerPoolPartition

callunity:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SequencerPartitionContainerImpl.java
```diff
@@ -46,9 +53,15 @@
...
+@Entity
+@Table(name = "SequencerPartitionContainer")
 public class SequencerPartitionContainerImpl extends AbstractSequencerPartitionContainer<SequencerPoolPartition> implements Serializable {
   protected static final Logger log = LoggerFactory.getLogger(SequencerPartitionContainerImpl.class);
 
+  @OneToMany(targetEntity = PartitionImpl.class, cascade = CascadeType.ALL)
+  @JoinTable(name = "SequencerPartitionContainer_Partition", joinColumns = {
+      @JoinColumn(name = "container_containerId") }, inverseJoinColumns = {
+          @JoinColumn(name = "partitions_partitionId") })
   private List<SequencerPoolPartition> partitions = new AutoPopulatingList<SequencerPoolPartition>(PartitionImpl.class);
```
apmasell: I'm concerned about how the assumptions of having an `AutoPopulatingList` are going to work when this is reloaded from the database and now a Hibernate list. Maybe make a note in the testing bug?

callunity: Yes, me too. This seems to be in place for the Spring UI. I've left a comment on the [post-hibernatization testing ticket](https://jira.oicr.on.ca/browse/GLT-1387?focusedCommentId=125150&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-125150).

sqlstore/src/main/resources/db/migration/V8900__hibernatize_everything.sql
```diff
@@ -36,8 +36,7 @@ ALTER TABLE SequencerReference ADD CONSTRAINT upgraded_SR_UK UNIQUE (upgradedSeq
 DROP TABLE Request;
 DROP TABLE Request_Note;
 
-ALTER TABLE Status ADD FOREIGN KEY (instrumentName) REFERENCES SequencerReference (name);
-ALTER TABLE Status ADD FOREIGN KEY (runName) REFERENCES Run (alias);
+ALTER TABLE Status ADD CONSTRAINT Status_SequencerReference_FK FOREIGN KEY (instrumentName) REFERENCES SequencerReference (name);
```
djcooke: Using name instead of ID as foreign key is odd enough, but the foreign key isn't even required. The instrument is already linked through the run, which is also linked by name instead of ID! :worried: 

(this is not really a change request since there's a ticket to improve it already; I just couldn't not say anything)

sqlstore/src/main/resources/db/migration/V8900__hibernatize_everything.sql
```diff
@@ -53,10 +52,10 @@ UPDATE RunQC_Partition rqp SET partition_partitionId = (
   AND p.partitionNumber = rqp.partitionNumber
 );
 ALTER TABLE RunQC_Partition CHANGE COLUMN partition_partitionId partition_partitionId BIGINT(20) NOT NULL;
+ALTER TABLE RunQC_Partition DROP PRIMARY KEY;
+ALTER TABLE RunQC_Partition ADD PRIMARY KEY(`runQc_runQcId`, `partition_partitionId`);
```
djcooke: :100: Nice catch

sqlstore/src/test/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLSequencerPartitionContainerDAOTest.java
```diff
@@ -270,26 +233,18 @@ public void testRemoveContainerFromUnassociatedRun() throws IOException {
...
     pc.setLocationBarcode("location");
-    pc.setPlatform(platformDAO.get(1L));
+    Platform platform = new PlatformImpl();
+    platform.setId(1L);
+    pc.setPlatform(platform);
     pc.setLastModifier(user);
     return pc;
   }
 
   private void mockAutoIncrement(long value) {
```
djcooke: Remove method - no longer does anything

callunity:
@frankie13 , could you take a look at this? The commit for changing containers -> Runs to ManyToMany does make some changes to things like the EI file system path generator, and I can't easily test that.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerPartitionContainer.java
```diff
@@ -194,13 +183,32 @@ public void setPaired(Boolean paired) {
...
+  @Override
+  public void setRuns(Collection<Run> runs) {
+    this.runs = runs;
+  }
+
+  @Override
+  public Run getLastRun() {
+    Run lastRun = null;
+    for (Run thisRun : getRuns()) {
+      if (thisRun == null) continue;
```
djcooke: It seems like nulls make their way into a lot of collections in MISO... :worried:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencerPartitionContainer.java
```diff
@@ -46,30 +46,33 @@
...
   /**
-   * Sets the run of this Container object.
-   * 
-   * @param run The run of which this Container is a part.
+   * Returns the last run of this Container object.
    * 
+   * @return Run run
    */
-  void setRun(Run run);
+  Run getLastRun();
```
djcooke: Please Javadoc this to note what exactly it's expected to return (The run with the latest start date of those which have a known status?)

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerPartitionContainer.java
```diff
@@ -194,13 +183,32 @@ public void setPaired(Boolean paired) {
...
+      if (lastRun == null) {
+        lastRun = thisRun;
+      } else if (thisRun.getStatus() != null && thisRun.getStatus().getStartDate().after(lastRun.getStatus().getStartDate())) {
```
djcooke: If there are two runs and neither has a status, this will return whichever was examined first - likely the earliest run assuming the collection will be ordered by natural DB order

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/sqlstore/SQLTgacSubmissionDAO.java
```diff
@@ -381,7 +381,7 @@ public SubmissionImpl mapRow(ResultSet rs, int rowNum) throws SQLException {
               runDAO.listBySequencerPartitionContainerId(partition.getSequencerPartitionContainer().getId()));
           // if there is 1 run for the flowcell/container, sets the run for that container to the first on on the list
           if (runs.size() == 1) {
-            partition.getSequencerPartitionContainer().setRun(runs.get(0));
+            partition.getSequencerPartitionContainer().addRun(runs.get(0));
```
djcooke: I think this should be a set, rather than an add, since the container could already have runs... Not really sure though

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerPartitionContainer.java
callunity: Checking with EI to see if they have any runs without statuses, or multiple statuses for a single run.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencerPartitionContainer.java
callunity: Done

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerPartitionContainer.java
```diff
@@ -207,8 +206,13 @@ public Run getLastRun() {
   }
 
   @Override
-  public void addRun(Run run) {
-    runs.add(run);
+  public void setRun(Run run) {
+    if (runs.size() < 2) {
+      runs.add(run);
```
djcooke: This can result in NPE and also acts as an add instead of a set. After setRun(Run notNull) is called, `runs.size() == 1` should always be true. Change to something like:
```
if (runs != null && runs.size() > 1) {
  throw new IllegalStateException("Cannot set single run on container with multiple runs already linked!");
} else {
  runs = new ArrayList<>();
  if (run != null) runs.add(run);
}
```

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractSequencerPartitionContainer.java
```diff
@@ -196,7 +196,6 @@ public void setRuns(Collection<Run> runs) {
   public Run getLastRun() {
     Run lastRun = null;
     for (Run thisRun : getRuns()) {
-      if (thisRun == null) continue;
       if (lastRun == null) {
         lastRun = thisRun;
       } else if (thisRun.getStatus() != null && thisRun.getStatus().getStartDate().after(lastRun.getStatus().getStartDate())) {
```
djcooke: Conditions for determining lastRun are still unbalanced. Both conditions should include the null check on `run.getStatus()`, or both should skip the null check. Status should never be null, and should be 1:1 with Run, so I'd say skip the null checks.

djcooke: startDate may also be null on `thisRun` and/or `lastRun`, which is actually valid and should not result in NPE's

callunity: Addressed.

PR #614, opened 2017-01-05
Glt 1280 hibernate notes

djcooke:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/FormUtils.java
```diff
@@ -2997,164 +2997,6 @@ public static JSONArray preProcessSampleSheetImport(File inPath, User u, Request
     }
   }
 
-  public static JSONArray processSampleSheetImport(File inPath, User u, RequestManager manager) throws Exception {
```
callunity: Have you confirmed with EI that they don't use this?

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/MigrationAuthorizationManager.java
```diff
@@ -83,4 +83,9 @@ public void throwIfNotWritable(SecurableByProfile resource) throws IOException,
     return new ArrayList<>(unfiltered);
   }
 
+  @Override
+  public void throwIfNonAdminOrMatchingOwner(User owner) throws IOException, AuthorizationException {
+
+  }
```
callunity: Is this method supposed to be blank?

core/src/main/java/com/eaglegenomics/simlims/core/Note.java
```diff
@@ -31,10 +35,19 @@
...
+
+  @Column(nullable = false, updatable = false)
   private Date creationDate = new Date();
+
+  @Column(nullable = false, updatable = false)
   private String text = "";
+
+  @Column(nullable = false, updatable = false)
   @ManyToOne
+  @JoinColumn(name = "owner_userId")
```
callunity: (You'll need to open the file to see the additional @Column above @ManyToOne

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/MigrationAuthorizationManager.java
djcooke: Yes, MigrationAuthorizationManager is used to bypass authorization and force use of a configured user when using the MISO Service layer outside of Spring. There is no reason to perform auth checks during migration, and the DefaultAuthorizationManager makes use of a security context, which doesn't exist in migration either

djcooke: This is noted in the class-level Javadoc if you look at the entire file

core/src/main/java/com/eaglegenomics/simlims/core/Note.java
djcooke: True. Will move the properties onto JoinColumn

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/MigrationAuthorizationManager.java
callunity: Yes, I did look at the Javadoc; perhaps it's that I'm still learning the Java conventions as to when it's ok to leave methods blank.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/FormUtils.java
djcooke: Not used from anywhere on develop branch. Confirmed now with @frankie13 that their internal config branch doesn't use it either. I had previously assumed that the "config" branch doesn't do forms processing stuff, but better safe I suppose!

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/MigrationAuthorizationManager.java
djcooke: A method body can be blank whenever it has void return type and you don't want it to actually do anything. It's nicer to leave a comment line explaining why it's empty though, so I will do that

djcooke:
Added fixups to address @callunity's comments

PR #630, opened 2017-01-17
Remove cache manager front end

apmasell:


spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/BarcodeHelperService.java
```diff
@@ -0,0 +1,100 @@
...
+import uk.ac.bbsrc.tgac.miso.core.data.impl.LibraryDilution;
+import uk.ac.bbsrc.tgac.miso.core.manager.RequestManager;
+
+/**
+ * uk.ac.bbsrc.tgac.miso.spring.ajax
+ * <p/>
+ * Info
+ * 
+ * @author Rob Davey
+ * @since 0.0.2
```
callunity: `@author` and `@since`

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/SampleControllerHelperService.java
```diff
@@ -119,7 +119,7 @@
   @Autowired
   private NamingScheme namingScheme;
   @Autowired
-  private CacheHelperService cacheHelperService;
+  private BarcodeHelperService cacheHelperService;
```
callunity: Please rename local variable to `barcodeHelperService`

apmasell: No. I will delete it instead.

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/BarcodeHelperService.java
apmasell: Fixed.

PR #632, opened 2017-01-18
Use Hibernate for Libraries and Dilutions

callunity:


core/src/test/java/uk/ac/bbsrc/tgac/miso/core/service/naming/generation/OicrLibraryAliasGeneratorTest.java
```diff
@@ -40,9 +38,9 @@ public void test_alias_generation_from_sample() throws Exception {
     library.setLibraryStrategyType(libraryStrategyType);
     LibraryDesignCode code = new LibraryDesignCode();
     code.setCode("WG");
-    library.getLibraryAdditionalInfo().setLibraryDesignCode(code);
+    library.setLibraryDesignCode(code);
 
-    assertThat(sut.generate(library), is("BART_1273_Br_P_PE_300_WG"));
+    assertThat(sut.generate((Library) library), is("BART_1273_Br_P_PE_300_WG"));
```
apmasell: This cast seems unnecessary or there's an inheritance problem somewhere.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractLibrary.java
```diff
@@ -77,78 +84,90 @@
...
+  @Column(nullable = false)
+  private PlatformType platformName;
+
   private String alias;
   private Boolean qcPassed;
-  private boolean lowQuality;
 
-  @Transient
-  private List<Index> indices = new ArrayList<>();
+  @Column(nullable = false)
```
callunity: Good point. I've made those changes.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateLibraryDilutionDao.java
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  private Criterion searchRestrictions(String querystr) {
+    return DbUtils.searchRestrictions(querystr, "name", "identificationBarcode", "library.name", "library.alias", "library.description");
+  }
+
+  @Override
+  public long save(LibraryDilution dilution) throws IOException {
+    long id;
+    if (dilution.getId() == AbstractLibrary.UNSAVED_ID) {
```
djcooke: It's probably the same, but shouldn't be comparing dilution's ID to AbstractLibrary's

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/changelog/SequencerPartitionContainerChangeLog.java
```diff
@@ -27,7 +27,7 @@
   private Long sequencerPartitionContainerChangeLogId;
 
   @ManyToOne(fetch = FetchType.LAZY, targetEntity = SequencerPartitionContainerImpl.class)
-  @JoinColumn(name = "containerId", nullable = false, updatable = false)
+  @JoinColumn(name = "containerId", referencedColumnName = "containerId", nullable = false, updatable = false)
```
callunity: I'll remove referencedColumnName (it has targetEntity)

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/RequestManager.java
```diff
@@ -604,4 +495,28 @@
...
+  public LibraryStrategyType getLibraryStrategyTypeByName(String name) throws IOException;
+
+  @Deprecated
+  public long saveLibraryDilution(LibraryDilution libraryDilution) throws IOException;
+
+  @Deprecated
+  public Collection<Library> listAllLibrariesBySampleId(long sampleId) throws IOException;
+
+  @Deprecated
+  public Collection<LibraryDilution> listAllLibraryDilutions() throws IOException;
```
callunity: Because the `FormUtils` class uses them, and its methods take a RequestManager parameter :((

I had assumed that also passing in the necessary services would not be desirable, but I didn't look into it beyond that. Do you have ideas as to how we might refactor that class?

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -429,11 +424,11 @@ public void saveLibraryDilutions(final Collection<LibraryDilution> libraryDiluti
...
-      
-      if (ldi.getLibrary().getId() == AbstractDilution.UNSAVED_ID && ldi.getLibrary().getLibraryAdditionalInfo() != null
-          && ldi.getLibrary().getLibraryAdditionalInfo().getPreMigrationId() != null) {
-        Long preMigrationId = ldi.getLibrary().getLibraryAdditionalInfo().getPreMigrationId();
-        ldi.setLibrary(serviceManager.getLibraryDao().getByPreMigrationId(preMigrationId));
+
+      if (ldi.getLibrary().getId() == LibraryDilution.UNSAVED_ID && ((DetailedLibrary) ldi.getLibrary()).getPreMigrationId() != null) {
+        Long preMigrationId = ((DetailedLibrary) ldi.getLibrary()).getPreMigrationId();
+        ldi.setLibrary(serviceManager.getLibraryDao()
+            .get(serviceManager.getLibraryDao().getByPreMigrationId(preMigrationId).getId()));
```
callunity: Oops, bad merge

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryDilutionService.java
```diff
@@ -0,0 +1,246 @@
...
+
+  @Override
+  public LibraryDilution get(long dilutionId) throws IOException {
+    LibraryDilution dilution = dilutionDao.get(dilutionId);
+    authorizationManager.throwIfNotReadable(dilution);
+    return dilution;
+  }
+
+  @Override
+  public LibraryDilution save(LibraryDilution dilution) throws IOException {
```
callunity: That's probably not necessary.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryDilutionService.java
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  @Override
+  public List<LibraryDilution> getAllByLibraryId(Long libraryId) throws IOException {
+    Collection<LibraryDilution> allDilutions = dilutionDao.listByLibraryId(libraryId);
+    return authorizationManager.filterUnreadable(allDilutions);
+  }
+
+  @Override
+  public List<LibraryDilution> getAllWithLimit(long limit) throws IOException {
```
callunity: It's used for the front page widget, which doesn't paginate.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryDilutionService.java
```diff
@@ -0,0 +1,246 @@
...
+   * @throws IOException
+   */
+  private void loadChildEntities(LibraryDilution dilution) throws IOException {
+    if (dilution.getLibrary() != null) {
+      dilution.setLibrary(libraryDao.get(dilution.getLibrary().getId()));
+    }
+    if (dilution.getTargetedSequencing() != null) {
+      dilution.setTargetedSequencing(targetedSequencingDao.get(dilution.getTargetedSequencing().getId()));
+    }
+    dilution.setDilutionCreator(authorizationManager.getCurrentUsername());
```
callunity: ...good point.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryDilutionService.java
```diff
@@ -0,0 +1,246 @@
...
+  /**
+   * Copies modifiable fields from the source LibraryDilution into the target LibraryDilution to be persisted
+   * 
+   * @param target the persisted LibraryDilution to modify
+   * @param source the modified LibraryDilution to copy modifiable fields from
+   * @throws IOException
+   */
+  private void applyChanges(LibraryDilution target, LibraryDilution source) {
+    target.setConcentration(source.getConcentration());
+    if (!autoGenerateIdBarcodes) target.setIdentificationBarcode(source.getIdentificationBarcode());
```
callunity: I had assumed that since there's no UI for the user to enter a barcode when idBarcode generation is enabled, there would be no reason to set it. Is that thinking too narrowly?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
```diff
@@ -0,0 +1,550 @@
...
+    authorizationManager.throwIfNotWritable(managed);
+    LibraryQC deleteQc = null;
+    for (LibraryQC qc : managed.getLibraryQCs()) {
+      if (qc.getId() == qcId) {
+        deleteQc = qc;
+        break;
+      }
+    }
+    if (deleteQc == null) throw new IOException("QC " + qcId + " not found for Library " + library.getId());
+    authorizationManager.throwIfNonAdminOrMatchingOwner(securityManager.getUserByLoginName(deleteQc.getQcCreator()));
```
callunity: Is there a disadvantage to doing a check for null qcCreator in this class, and calling the throwIfNonAdminOrMatchingOwner if user is not null?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/RequestManager.java
djcooke: The FormUtils methods that use RequestManager could go into a different, autowirable class which has the appropriate Services wired into it. It'd be fair to call that out of scope of this ticket though

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
```diff
@@ -0,0 +1,550 @@
...
+  }
+
+  /**
+   * Copies modifiable fields from the source Library into the target Library to be persisted
+   * 
+   * @param target the persisted Library to modify
+   * @param source the modified Library to copy modifiable fields from
+   * @throws IOException
+   */
+  private void applyChanges(Library target, Library source) throws IOException {
```
callunity: Aha, I had missed the part where the things from `loadChildEntities` first need to have (possibly) changed IDs set here.

callunity: (volume and qcPassed are already set in this method)

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryDilutionService.java
djcooke: Ah, I didn't realize there was no way to change it, but in that case there's no harm in accepting a change that would be considered fine if it were possible. Either way is fine

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/RequestManager.java
callunity: I'll make a new ticket for it.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
djcooke: It's not a problem with qcCreator being null. It's a problem with qcCreator being a String that is not an existing MISO username. If `securityManager.getUserByLoginName(deleteQc.getQcCreator())` returns null, then `authorizationManager.throwIfNonAdminOrMatchingOwner(securityManager.getUserByLoginName(deleteQc.getQcCreator()))` will throw a NPE. The correct behaviour would be for `throwIfNonAdminOrMatchingOwner` to only check for admin status if the "MatchingOwner" parameter is null

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -266,16 +267,12 @@ private Sample save(Sample sample) throws IOException {
         created.setAlias(generatedAlias);
         needsUpdate = true;
       }
-      if (isStringBlankOrNull(sample.getAlias())) {
-        sample.setAlias(sample.getName());
+      if (isStringBlankOrNull(created.getAlias())) {
+        created.setAlias(created.getName());
```
callunity: Probably just setting it to something so it wouldn't have a blank alias?

djcooke: There shouldn't be a fail-safe here. The alias should either be validated as given or generated using the name generator. Above, this method also does validation *after* saving instead of before

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryDilutionService.java
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  /**
+   * Copies modifiable fields from the source LibraryDilution into the target LibraryDilution to be persisted
+   * 
+   * @param target the persisted LibraryDilution to modify
+   * @param source the modified LibraryDilution to copy modifiable fields from
+   * @throws IOException
+   */
+  private void applyChanges(LibraryDilution target, LibraryDilution source) {
```
callunity: Fixed

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
```diff
@@ -0,0 +1,550 @@
...
+        throw new IOException("Error generating alias for library", e);
+      }
+    } else {
+      namingScheme.validateLibraryAlias(library.getAlias());
+    }
+    return save(library).getId();
+  }
+
+  @Override
+  public void update(Library library) throws IOException {
```
callunity: Updated.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractLibraryQC.java
```diff
@@ -46,9 +45,9 @@
   private Double results;
 
   @Column(nullable = false)
-  private int insertSize;
+  private Integer insertSize;
```
djcooke: Getter and setter should match this. Getter will throw NPE if not set currently

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
callunity: Point.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
```diff
@@ -0,0 +1,550 @@
...
+  }
+
+  /**
+   * Confirms that when a Library Design is selected (detailed library only), the corresponding
+   * LibraryDesignCode, LibrarySelection and LibraryStrategy all match the values for the selected LibraryDesign.
+   *
+   * @param library the Library with desired selection, strategy, and libraryDesignCode set.
+   * @throws IOException when the library's values don't match the values of the LibraryDesign
+   */
+  private void validateLibraryDesignValues(Library library) throws IOException {
```
callunity: Fixed

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryDilutionService.java
djcooke: :+1:

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryDilutionService.java
djcooke: :+1:

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -271,6 +270,11 @@ private Sample save(Sample sample) throws IOException {
         created.setAlias(created.getName());
         needsUpdate = true;
       }
+      if (autoGenerateIdBarcodes && isStringEmptyOrNull(created.getIdentificationBarcode())) {
```
djcooke: Can probably happen before initial save, so it doesn't cause a resave to be required

callunity: It can't happen before create because the name (generated after create) is used in the idBarcode

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/security/DefaultAuthorizationManager.java
```diff
@@ -130,9 +130,13 @@ public void throwIfNotWritable(SecurableByProfile resource) throws IOException,
     return filtered;
   }
 
+  /**
+   * Does not throw if owner is null (editable by anyone).
+   */
   @Override
   public void throwIfNonAdminOrMatchingOwner(User owner) throws IOException, AuthorizationException {
     User currentUser = getCurrentUser();
+    if (owner == null) return;
```
djcooke: This is an unintended grant of permissions to all users. If the owner doesn't exist, this should instead call throwIfNonAdmin

callunity: I think that if the owner doesn't exist because they're no longer there, it should be editable by other lab members (who aren't admins) because dilutions should be editable by all lab members in the same way that samples and libraries are. Perhaps it could be restricted to editing if the currentUser has permissions on the dilution, but isn't necessarily the owner?

PR #633, opened 2017-01-19
Remove non-SQLStore references to DataObjectFactory

apmasell:
In preparation for its demise

### DIFF
--- a/spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/RunControllerHelperService.java
+++ b/spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/RunControllerHelperService.java
@@ -67,13 +67,14 @@ import uk.ac.bbsrc.tgac.miso.core.data.SequencerPoolPartition;
 import uk.ac.bbsrc.tgac.miso.core.data.SequencerReference;
 import uk.ac.bbsrc.tgac.miso.core.data.Study;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.RunImpl;
+import uk.ac.bbsrc.tgac.miso.core.data.impl.RunQCImpl;
+import uk.ac.bbsrc.tgac.miso.core.data.impl.SequencerPartitionContainerImpl;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.illumina.IlluminaRun;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.ls454.LS454Run;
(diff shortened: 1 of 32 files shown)


spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/RunControllerHelperService.java
```diff
@@ -124,7 +123,7 @@ public JSONObject changePlatformType(HttpSession session, JSONObject json) {
         if (newPt != null) {
           log.info("STORED: " + newRuntype + " :: " + storedPlatformType);
           if (!newRuntype.equals(storedPlatformType)) {
-            run = dataObjectFactory.getRunOfType(newPt, user);
+            run = new RunImpl(user);
```
djcooke: Should create a subclass (IlluminaRun, SolidRun, etc)

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/RunControllerHelperService.java
```diff
@@ -185,7 +184,7 @@ public JSONObject populateRunOptions(HttpSession session, JSONObject json) {
       Map<String, Object> responseMap = new HashMap<>();
       User user = securityManager.getUserByLoginName(SecurityContextHolder.getContext().getAuthentication().getName());
 
-      Run run = dataObjectFactory.getRunOfType(pt, user);
+      Run run = new RunImpl(user);
```
djcooke: subclass

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/RunControllerHelperService.java
apmasell: Are the subclasses actually needed? The only difference between those classes is the constructor when passed the XML data from the client, which none of these methods do.

djcooke: Difficult to say for sure since this makes a pass through the UI and we can't deploy it. Minimally, the platformtype needs set and a new SecurityProfile and Status should be set on the Run

apmasell: Okay. I've fixed this.

PR #643, opened 2017-01-24
Make more unit tests pass

callunity:


spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/PoolWizardControllerHelperService.java
```diff
@@ -235,7 +236,9 @@ public JSONObject addStudy(HttpSession session, JSONObject json) {
       s.setAlias(p.getAlias());
       s.setDescription(studyDescription);
       s.setSecurityProfile(p.getSecurityProfile());
-      s.setStudyType(studyType);
+      StudyType sType = new StudyType();
```
apmasell: This is not going to work, but I have a patch that fixes this. So, proceed.

djcooke:
:+1:

sqlstore/src/test/resources/db/test_migration/V9000__miso_test_data.test.sql
```diff
@@ -225,7 +225,7 @@ VALUES (1,'admin','2015-08-27',1,5),
 (12,'admin','2015-08-27',1,5),
 (13,'admin','2015-08-27',1,5),
 (14,'admin','2015-08-27',1,5),
-(15,'admin','2015-08-27',1,5);
+(14,'admin','2015-08-27',1,5);
```
tonydebat: Duplicate line.

PR #644, opened 2017-01-25
Remove dead Lane object and update annotations and tests

apmasell:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/ExperimentImpl.java
```diff
@@ -115,6 +115,7 @@
   private Run run;
 
   @ManyToOne(cascade = CascadeType.ALL)
+  @JoinColumn
```
callunity: `(name = "securityProfile_profileId")`

djcooke:
:+1:

PR #649, opened 2017-01-26
GLT-1471 Excluded javassist from hibernate and reflections. Added newer version.

tonydebat:


apmasell:
Actually, we might be able to ditch `org.reflections`. I think it was only used by the printer code I've since deleted.

tonydebat:
I removed org.reflections.

PR #652, opened 2017-01-27
Change interceptor design to avoid multiple Hibernate sesions

apmasell:
The original interceptor design to log changes to pools and boxes created
multiple Hibernate sessions. This attaches the interceptor to the Hibernate
session factory.

### DIFF
--- a/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/CollectionLoggingInterceptor.java
+++ b/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/CollectionLoggingInterceptor.java
@@ -1,0 +1,182 @@
+package uk.ac.bbsrc.tgac.miso.persistence.impl;
+
+import java.io.Serializable;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Queue;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+import org.hibernate.EmptyInterceptor;
+import org.hibernate.type.Type;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.BeanFactory;
+import org.springframework.beans.factory.BeanFactoryAware;
+
+import uk.ac.bbsrc.tgac.miso.core.data.Box;
+import uk.ac.bbsrc.tgac.miso.core.data.Boxable;
+import uk.ac.bbsrc.tgac.miso.core.data.ChangeLog;
+import uk.ac.bbsrc.tgac.miso.core.data.Dilution;
+import uk.ac.bbsrc.tgac.miso.core.data.Pool;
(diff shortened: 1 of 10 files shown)


djcooke:
What is the advantage to using an interceptor rather than just writing the changelogs during save in the Box and Pool Services?

apmasell:
This was the solution Chris used in the original Pool conversion to Hibernate and it seems pretty standard when you look at ways of doing audit logging for Hibernate.

djcooke:
It makes sense when abstraction is possible, but that's not the case here. Is there any other advantage? The weirdness that has to happen to wire in the DAO should be justified

apmasell:
We use that DAO wiring pattern twice for `ApplicationContext` and autowire much of the notification processing system the same way. If you want something changed, name it.

djcooke:
My real problem with this is adding a bean that should not really be used for anything to the IoC container in order to autowire something into a static field just to use it in another instance of the object. Having multiple instances of something that should be a singleton tends to cause issues that are a pain to debug. I see two solutions:

My preferred solution: Create the changelogs in the Box and Pool Services instead of using an interceptor since they are specific to those classes.

Other option: instead of setting the interceptor *class* in hibernateProperties, set the interceptor as a reference to the actual *bean* via `entityInterceptor` property on the `sessionFactory` bean. Get rid of the static/context stuff and just autowire the DAO in the interceptor class

apmasell:
I used the second solution. Fixed.

apmasell:
No, sorry, this doesn't work because there is a cycle in that the interceptor needs the session factory (transitively by the change log store).

djcooke:
You could make the interceptor implement BeanFactoryAware and use that to get the "sessionFactory".

Whatever way you work around it, the circular dependency is another point against doing it this way though. Yes, an interceptor makes sense for audit logging when all entities are to be handled identically, but this is for two specific entity types only, and they even are handled differently. I still fail to see the benefit of an interceptor here.

apmasell:
Okay. That deploys.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/CollectionLoggingInterceptor.java
```diff
@@ -0,0 +1,185 @@
...
+      if (names.length() > 0) {
+        names.append(", ");
+      }
+      names.append(dilution.getName());
+    }
+    return names.toString();
+  }
+
+  @Override
+  public boolean onFlushDirty(Object entity, Serializable id, Object[] currentState, Object[] previousState, String[] propertyNames,
```
djcooke: Could you add tests for this? The specific part I'm concerned about is how `(HashSet) x.removeAll(y)` is used to determine changes. While this should be fine, we've had several problems relating to hashcode methods and they're still not consistent across our models.

apmasell: I switched this to use strings as box does so the implementation of `equals` and `hashCode`  don't matter.

apmasell:
I've rebased for the new change log interface and heavily de-duplicated the copy-pasta.

PR #654, opened 2017-01-27
Project, security, and experiment fixes

apmasell:


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateProjectDao.java
```diff
@@ -67,8 +67,18 @@
...
+  @Autowired
   private JdbcTemplate template;
 
+  @Autowired
+  private WatchManager watchManager;
+
+  @Override
+  public void addWatcher(Project project, User watcher) {
+    watchManager.watch(project, watcher);
+    currentSession().update(project);
```
callunity: I've got a fix in a PR that [removes the watchManager](https://github.com/TGAC/miso-lims/pull/650/commits/20a0485dc9e29148fd458d1b4ab2e4809ddae3b8), so let's just make sure that doesn't get squashed by this :)

PR #657, opened 2017-01-27
use managed entities in requestManager

callunity:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/RequestManager.java
```diff
@@ -501,4 +497,10 @@
    */
   @Deprecated
   public Collection<LibraryDilution> listAllLibraryDilutions() throws IOException;
+
+  /**
+   * @Deprecated in v0.28. Use sampleService.save instead
+   */
+  @Deprecated
+  public long saveSample(Sample sample) throws IOException;
```
djcooke: Should just be removed, but that'll happen with switch-over to Services anyway

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
```diff
@@ -1130,18 +1148,28 @@ public long saveRun(Run run) throws IOException {
...
+        saveRun(run);
+      }
     } else {
       throw new IOException("No runStore available. Check that it has been declared in the Spring config.");
     }
   }
 
   @Override
-  public long saveRunQC(RunQC runQC) throws IOException {
+  public void saveRunQC(RunQC runQC) throws IOException {
```
apmasell: The caller never consumed the output.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
```diff
@@ -1130,18 +1148,28 @@ public long saveRun(Run run) throws IOException {
...
     if (runQcStore != null) {
-      long id = runQcStore.save(runQC);
-      if (runAlertManager != null) runAlertManager.update(runQC.getRun());
```
apmasell: Yes. Please revert.

callunity: Reverted.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
djcooke: It's easily accessible, can be useful, and all of the other save methods return the inserted ID. I don't see a reason to remove it, and would prefer consistency

PR #660, opened 2017-01-30
GLT-1452: removed Library things from RequestManager

djcooke:
Includes removing some dead methods from FormUtils, and moving FormUtils from core to the spring module


spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/util/FormUtils.java
```diff
@@ -0,0 +1,1291 @@
...
+import uk.ac.bbsrc.tgac.miso.service.SampleService;
+
+/**
+ * uk.ac.bbsrc.tgac.miso.core.util
+ * <p/>
+ * Info
+ * 
+ * @author Rob Davey
+ * @date 06-Sep-2011
+ * @since 0.1.1
```
callunity: I'm not sure the @since is accurate anymore, and the comment above about this being in `core.util` definitely isn't.

djcooke: Changing the real value of `@since` would require time travel, but I will remove the useless/inaccurate parts of the doc (package comment and `@date`)

PR #662, opened 2017-01-31
GLT-1451: moved Kit things from RequestManager to Service

djcooke:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditKitDescriptorController.java
```diff
@@ -95,14 +95,13 @@ public ModelAndView setupForm(@PathVariable Long kitDescriptorId, ModelMap model
         kitDescriptor = new KitDescriptor();
         model.put("title", "New Kit Descriptor");
       } else {
-        kitDescriptor = requestManager.getKitDescriptorById(kitDescriptorId);
+        kitDescriptor = kitService.getKitDescriptorById(kitDescriptorId);
+        if (kitDescriptor == null) {
+          throw new SecurityException("No such Kit Descriptor");
```
callunity: Why SecurityException?

djcooke: I left as it was. What should happen here is a nice 404 page, but we don't have exception handlers that do that, so it'll just show a stack trace whatever the exception is

PR #665, opened 2017-02-01
Show/hide columns on propagate samples page

frankie13:
Don't forget, you need new things in the properties file for this to work!

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/DefaultNamingScheme.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/DefaultNamingScheme.java
@@ -1,5 +1,7 @@
 package uk.ac.bbsrc.tgac.miso.core.service.naming;
 
+import org.springframework.web.context.support.SpringBeanAutowiringSupport;
+
 import uk.ac.bbsrc.tgac.miso.core.data.Library;
 import uk.ac.bbsrc.tgac.miso.core.data.Nameable;
 import uk.ac.bbsrc.tgac.miso.core.data.Sample;
@@ -24,6 +26,13 @@ public class DefaultNamingScheme extends AbstractNamingScheme {
   private NameValidator libraryAliasValidator = new DefaultLibraryAliasValidator();
   private NameGenerator<Library> libraryAliasGenerator = new DefaultLibraryAliasGenerator();
   private NameValidator projectShortNameValidator = new AllowAnythingValidator();
+
+  /**
+   * Creates a new DefaultNamingScheme and attempts to autowire all of its validators' and generators' dependencies.
+   */
+  public DefaultNamingScheme() {
+    SpringBeanAutowiringSupport.processInjectionBasedOnCurrentContext(libraryAliasGenerator);
+  }
 
   @Override
   public void setNameGenerator(NameGenerator<Nameable> generator) {
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
@@ -92,9 +92,11 @@ import uk.ac.bbsrc.tgac.miso.core.data.type.LibraryStrategyType;
 import uk.ac.bbsrc.tgac.miso.core.data.type.LibraryType;
 import uk.ac.bbsrc.tgac.miso.core.data.type.PlatformType;
 import uk.ac.bbsrc.tgac.miso.core.exception.MalformedLibraryException;
+import uk.ac.bbsrc.tgac.miso.core.exception.MisoNamingException;
 import uk.ac.bbsrc.tgac.miso.core.manager.RequestManager;
 import uk.ac.bbsrc.tgac.miso.core.security.util.LimsSecurityUtils;
 import uk.ac.bbsrc.tgac.miso.core.service.IndexService;
+import uk.ac.bbsrc.tgac.miso.core.service.naming.NamingScheme;
 import uk.ac.bbsrc.tgac.miso.core.util.AliasComparator;
 import uk.ac.bbsrc.tgac.miso.core.util.LimsUtils;
 import uk.ac.bbsrc.tgac.miso.dto.DetailedLibraryDto;
@@ -145,6 +147,17 @@ public class EditLibraryController {
   @Autowired
   private KitService kitService;
 
+  @Autowired
+  private NamingScheme namingScheme;
+
+  public NamingScheme getNamingScheme() {
+    return namingScheme;
+  }
+
+  public void setNamingScheme(NamingScheme namingScheme) {
+    this.namingScheme = namingScheme;
+  }
+
   public void setRequestManager(RequestManager requestManager) {
     this.requestManager = requestManager;
   }
@@ -175,7 +188,12 @@ public class EditLibraryController {
   private Boolean autoGenerateIdBarcodes;
   @Value("${miso.detailed.sample.enabled}")
   private Boolean detailedSample;
-
+  @Value("${miso.display.library.bulk.libraryalias}")
+  private Boolean showLibraryAlias;
+  @Value("${miso.display.library.bulk.description}")
+  private Boolean showDescription;
+  @Value("${miso.display.library.bulk.volume}")
+  private Boolean showVolume;
   @ModelAttribute("metrixEnabled")
   public Boolean isMetrixEnabled() {
     return metrixEnabled;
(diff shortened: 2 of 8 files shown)


morgantaschuk:
If building with Java 8, something needs to be changed in the POM.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -859,16 +884,61 @@ public ModelAndView editPropagateSamples(@PathVariable String sampleIds, ModelMa
...
+          List<String> alias2Nums = new ArrayList<>(Arrays.asList(alias2.split(regex)));
+
+          alias1Nums.removeAll(Arrays.asList(""));
+          alias2Nums.removeAll(Arrays.asList(""));
+          if (alias1Nums.size() != alias2Nums.size()) {
+            // give up, just compare by default.
+            rtn = alias1.compareTo(alias2);
+          }
+          for (int i = 0; i < alias1Nums.size(); i++) {
+            if (Integer.valueOf(alias1Nums.get(i)) > Integer.valueOf(alias2Nums.get(i))) {
```
apmasell: If you prefer:
```
for (int i = 0; i < alias1Nums.size() && rtn == 0; i++) {
￼ rtn = Integer.compare(Integer.valueOf(alias1Nums.get(i)), Integer.valueOf(alias2Nums.get(i)));
}
```

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/DefaultNamingScheme.java
```diff
@@ -47,7 +57,7 @@ public void setSampleAliasValidator(NameValidator validator) {
 
   @Override
   public void setLibraryAliasGenerator(NameGenerator<Library> generator) {
-    this.libraryAliasGenerator = generator;
+    this.libraryAliasGenerator = (DefaultLibraryAliasGenerator) generator;
```
djcooke: Remove cast

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/DefaultNamingScheme.java
```diff
@@ -22,9 +24,17 @@
   private NameValidator sampleAliasValidator = new DefaultSampleAliasValidator();
   private NameGenerator<Sample> sampleAliasGenerator = null;
   private NameValidator libraryAliasValidator = new DefaultLibraryAliasValidator();
-  private NameGenerator<Library> libraryAliasGenerator = new DefaultLibraryAliasGenerator();
+  private DefaultLibraryAliasGenerator libraryAliasGenerator = new DefaultLibraryAliasGenerator();
```
djcooke: `NameGenerator` must remain the declared type - configuration may cause this to be replaced by a different generator

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -845,6 +866,10 @@ public ModelAndView editPropagateSamples(@PathVariable String sampleIds, ModelMa
           hasPlain = true;
         }
         LibraryDto library = new LibraryDto();
+        Library tempLibrary = new LibraryImpl();
+        tempLibrary.setSample(sample);
+        final String libraryAlias = namingScheme.generateLibraryAlias(tempLibrary);
```
djcooke: Alias generation should be handled in the Service layer unless it's necessary to display the alias before the library is saved?

frankie13: it is I'm afraid.  Without an alias the form is invalid so we need to know it upfront.

frankie13:
@djcooke @morgantaschuk I believe I have addressed all the things!

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
djcooke: That should be changed to match how the sample forms work: If sample alias generation is enabled, it is not a required field. It may still be provided if you'd like to use a custom alias, but leaving it blank will result in one being generated.

Generating the alias during save is also better for generating unique names. If it's done during creation and two people are creating libraries at the same time, it can result in a collision.

miso-web/src/main/webapp/pages/bulkEditLibraries.jsp
```diff
@@ -78,15 +78,15 @@
...
- 
+
  </div>
-  
+
   <script type="text/javascript">
     jQuery(document).ready(function () {
       Library.hot.librariesJSON = ${librariesJSON};
-      Library.hot.librariesJSON = Hot.sortByProperty(Library.hot.librariesJSON, 'parentSampleAlias');
+      Library.hot.hideCols = ${hideCols};
```
callunity: `${hideCols}` isn't set for bulk editing, which causes critical JS errors.

miso-web/src/main/webapp/pages/bulkEditLibraries.jsp
```diff
@@ -78,15 +78,15 @@
...
    <div id="hotContainer"></div>
- 
+
  </div>
-  
+
   <script type="text/javascript">
     jQuery(document).ready(function () {
       Library.hot.librariesJSON = ${librariesJSON};
-      Library.hot.librariesJSON = Hot.sortByProperty(Library.hot.librariesJSON, 'parentSampleAlias');
```
callunity: Please wrap the `Hot.sortByProperty` line if an `if Hot.detailedSample` check (and move the `Hot.detailedSample` assignment up above this line) instead of deleting it, as we will still want to use this sorting in our lab.

frankie13: :white_check_mark:

miso-web/src/main/webapp/pages/bulkEditLibraries.jsp
frankie13: :white_check_mark:

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
frankie13: I agree with you @djcooke - this approach would be cleaner.
We do however have something that works, and works well for now and given that this task has already been dragging on for sometime I suggest we go with this code as it stands with a view to changing the approach in the future.
I am concerned with getting the develop branch deployed to our production server soon as I had hoped to do it around new year, since this issue isn't blocking / breaking anything I recommend we press on!

miso-web/src/main/webapp/pages/bulkEditLibraries.jsp
```diff
@@ -78,20 +78,32 @@
...
+
+      if (Hot.detailedSample) {
+        Library.hot.librariesJSON = Hot.sortByProperty(Library.hot.librariesJSON, 'parentSampleAlias');
+      }
+      <c:choose>
+	    <c:when test="${not empty hideCols}">
+	      Library.hot.hideCols = ${hideCols};
+	    </c:when>    
+	    <c:otherwise>
+	      Library.hot.hideCols = {};
```
callunity: Since you're initializing `${hideCols}` as a JSONArray, you could also just make the assignment instead of wrapping it in the test; if you keep it this way, I think this needs to be an empty array instead of empty object since you're doing checks for `indexOf`.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -775,6 +793,10 @@ public ModelAndView editPropagateSamples(@PathVariable String sampleIds, ModelMa
           hasPlain = true;
         }
         LibraryDto library = isDetailedSampleEnabled() ? new DetailedLibraryDto() : new LibraryDto();
+        Library tempLibrary = new LibraryImpl();
+        tempLibrary.setSample(sample);
+        final String libraryAlias = namingScheme.generateLibraryAlias(tempLibrary);
```
callunity: Please wrap 798 and 799 in `if (namingScheme.hasLibraryAliasGenerator())`

miso-web/src/main/resources/external/miso.properties
```diff
@@ -39,6 +39,10 @@ miso.alerting.enabled:true
 # toggle mirroring of all alerts to system alerts (user id 0)
 miso.alerting.saveSystemAlerts:true
 
+## configs for caching
+# precache objects at MISO startup. This will greatly increase startup time, but will improve MISO performance for users at the outset
+miso.db.caching.precache.enabled:false
```
callunity: This won't do much since Hibernate takes care of the caching, but that means it can be deleted later :)

PR #676, opened 2017-02-03
Box and pool saving

apmasell:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/BoxImpl.java
```diff
@@ -42,12 +42,20 @@
...
+  @JoinTable(name = "BoxPosition", joinColumns = { @JoinColumn(name = "boxId") }, inverseJoinColumns = {
+      @JoinColumn(name = "targetId") })
   @AnyMetaDef(idType = "long", metaType = "string", metaValues = {
       @MetaValue(targetEntity = LibraryImpl.class, value = "L"),
+      @MetaValue(targetEntity = DetailedLibraryImpl.class, value = "LD"),
       @MetaValue(targetEntity = PoolImpl.class, value = "P"),
-      @MetaValue(targetEntity = SampleImpl.class, value = "S")
+      @MetaValue(targetEntity = SampleImpl.class, value = "S"),
+      @MetaValue(targetEntity = SampleAliquotImpl.class, value = "SA"),
+      @MetaValue(targetEntity = IdentityImpl.class, value = "SI"),
```
apmasell: True, but we don't have any logic that prevents them from going in a box if they have a barcode. :-(
http://cdn2.stylecraze.com/wp-content/uploads/2013/11/Human-Inscription-with-Barcode-Tattoo-Design.jpg

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/BoxImpl.java
```diff
@@ -42,12 +42,20 @@
...
+      @MetaValue(targetEntity = SampleStockImpl.class, value = "SS"),
+      @MetaValue(targetEntity = SampleTissueImpl.class, value = "ST"),
+      @MetaValue(targetEntity = SampleTissueProcessingImpl.class, value = "SP"),
+      @MetaValue(targetEntity = SampleCVSlideImpl.class, value = "SC"),
+      @MetaValue(targetEntity = SampleLCMTubeImpl.class, value = "SL"),
```
djcooke: Can these be changed to something more descriptive like "Sample" and "SampleTissue" instead of "S" and "ST"?

apmasell: Fixed.

PR #679, opened 2017-02-06
Fix user saving and fix password handling

### LINKED ISSUE #631: Cannot change admin password or create user on local instance
Hello,

the manager of our in-house sequencing facility wants to use MISO, so I have set up a local instance as described in docs/_posts/2016-01-11-admin-manual.md.

However, when I change the password of the default admin user, it has no effect. Changing the email address works. I also cannot create a new user, the error message I get is "Cannot create user - user isn't an admin.". Unfortunately the stack trace isn't helpful to me.

Thanks in advance for your help!

apmasell:
Closes #631 and GLT-462


docs/_posts/2016-01-11-admin-manual.md
```diff
@@ -173,10 +173,7 @@ The configuration files are:
 | `submission.properties`   | properties to set the submission environment.              |
 
 ## Security Environment
-MISO can use either LDAP or JDBC as an authentication mechanism. The mechanism
-is set in both `/storage/miso/security.properties` and the
-`$CATALINA_HOME/bin/setenv.sh` or `/etc/default/tomcat8` files and both must be
-the same.
+MISO can use either LDAP or JDBC as an authentication mechanism.
```
callunity: Please indicate in which file the authentication property must be set, as the lines below reference setting the security method.

apmasell: Expanded.

docs/_posts/2016-01-11-admin-manual.md
```diff
@@ -173,7 +173,9 @@ The configuration files are:
 | `submission.properties`   | properties to set the submission environment.              |
 
 ## Security Environment
-MISO can use either LDAP or JDBC as an authentication mechanism.
+MISO can use either LDAP (`ldap`), Active Directory LDAP (`ad`), or JDBC
+(`jdbc1) as an authentication mechanism. This is set by the `-Dsecurity.method`
```
callunity: "`jdbc1" -> "`jdbc`"

docs/_posts/2016-01-11-admin-manual.md
callunity: If this is the case, then it would make more sense to move the entire `-DsecurityMethod` part down here rather than having the user add it and then have to go looking upwards and re-open the file.

docs/_posts/2016-01-11-admin-manual.md
apmasell: Fixed.

tonydebat:
Tested ad and jdbc. With active directory set the 'edit account' page worked correctly, allowing me to only update my email address.

In jdbc mode I was prevented from logging in with a null password, which is good. 

Once I did set a password, I was able to login, but I was not able to change my password via the UI. The following message was still being displayed "Password change support only available for the 'jdbc' security method. If using LDAP, please change the user password in your LDAP server."


image

apmasell:
That is the intended behaviour. There was code to change the password in LDAP, but there was also code to not change it. Since the code surrounding how the administrator changes the password didn't work with the above code, I set everything to not allow changing it.

tonydebat:
I'm saying that I was using the `jdbc` security method and was prevented from updating my password. The error message indicates that updating a password should be available for `jdbc`.

If changing the password has been completely disabled then I would suggest either completely removing the password field and message, or provide a message that indicates users will need to ask an admin to update their password.

apmasell:
Fixed. I missed a variable rename.

miso-web/src/main/webapp/pages/editUser.jsp
```diff
@@ -189,8 +188,7 @@
               <c:otherwise>
                 <tr>
                   <td>Password:</td>
-                  <td><i>Password change support only available for the 'jdbc' security method. If using LDAP, please change
-                    the user password in your LDAP server.</i></td>
+                  <td><i>Change password using your the method specified by IT.</i></td>
```
callunity: "To change your password for your LDAP account you must use the methods specified by IT." ?

apmasell: Our users don't really know what LDAP is, so I'd rather not mention it.

callunity: Ok. Then could you please remove the extraneous "your" (4th word, and also in the changed line above)?

apmasell: Sorry. Fixed.

PR #689, opened 2017-02-08
Remove dead binding initialisers

apmasell:


callunity:
Changing a LibraryDesign on a library results in the following:
`identifier of an instance of uk.ac.bbsrc.tgac.miso.core.data.LibraryDesign was altered from 4 to 1; nested exception is org.hibernate.HibernateException: identifier of an instance of uk.ac.bbsrc.tgac.miso.core.data.LibraryDesign was altered from 4 to 1`

apmasell:
There was a number of unrelated problems with the library service that I have fixed.

tonydebat:
Andre, how did you identify the dead bindings?

apmasell:
The dead bindings were bindings for types we don't use any more (emPCRs and such). The actual bound fields have been relocated to more appropriate binders.

PR #695, opened 2017-02-13
GLT-1509: added option to correct UI timezone

djcooke:
Nothing is changed unless you add the relevant options to your `miso.properties` file. Explanation of problem and solution in Javadoc and test comments


callunity:
Relevant: https://xkcd.com/1799

PR #696, opened 2017-02-14
GLT-1216 Targeted sequencing values from GSLE

tonydebat:
This does not include values which may be used by the pacbio.


tonydebat:
Thanks, I'll move this and look into the ValueTypeLookup.

PR #699, opened 2017-02-15
Fix bug causing pool order completions to be incorrect

apmasell:
They id keys were wrong in Hibernate and it was throwing away data

JIRA: GLT-1522


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/PoolOrderCompletion.java
```diff
@@ -22,23 +23,57 @@
...
+      this.pool = pool;
+    }
+
+  }
+
   private static final long serialVersionUID = 1L;
 
-  @Enumerated(EnumType.STRING)
-  private HealthType health;
+  @Id
```
djcooke: Should this be `@EmbeddedId`?

apmasell: Seems to work either way, but sure.

PR #700, opened 2017-02-15
Delete deprecated classes and most deprecated methods

callunity:
🔥


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/RunInfoUtils.java
```diff
@@ -1,109 +0,0 @@
...
-   * 
-   * @param runName
-   *          of type String
-   * @return Map<String, String>
-   * @throws IOException
-   *           when
-   */
-  public static Map<String, String> checkIlluminaStatus(String runName) throws IOException {
-    ProcessBuilder pb = new ProcessBuilder();
-    pb.command("ssh", "davey@149.155.209.208", "./glob_illumina_status.pl", "-p", runName);
```
djcooke: 😆

PR #702, opened 2017-02-16
GLT-1511 more fixes for migration

callunity:


migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -553,6 +553,7 @@ public void saveRuns(final Collection<Run> runs) throws IOException {
       Run run = serviceManager.getRunDao().getByAlias(newRun.getAlias());
       if (run == null) {
         run = newRun;
+        run.getStatus().setLastUpdated(new Date());
```
djcooke: `Status` is about to disappear, so it doesn't really matter, but should use `timeStamp` member var of DefaultMigrationTarget instead of new `Date`s so everything with the migration's date has the exact same value

migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -571,6 +572,10 @@ private void updateRun(Run from, Run to) throws IOException {
     log.debug("Updating run " + to.getId());
     to.getStatus().setCompletionDate(to.getStatus().getCompletionDate());
     to.getStatus().setHealth(to.getStatus().getHealth());
+
+    // slated for demolition
+    to.getStatus().setLastUpdated(new Date());
```
djcooke: Again should be `timeStamp` but doesn't really matter

PR #704, opened 2017-02-16
Fix discriminator values in Hibernate migration for boxes

apmasell:


callunity:
Do the `BoxImpl` discriminators need to be updated as well?

djcooke:
It looks like the BoxImpl MetaValues match what's going into the DB now.

The DiscriminatorValues on SampleImpl, LibraryImpl, and PoolImpl aren't actually used for anything now, right? If not, please remove

callunity:
Aha, I missed the `CONCAT('Sample', ...)`

PR #705, opened 2017-02-16
GLT-1518 Copy values from Status table xml to new fields in the Run table.

tonydebat:
Configure flyway to run Java migrations as well as SQL migrations.

### DIFF
--- a/sqlstore/pom.xml
+++ b/sqlstore/pom.xml
@@ -219,6 +219,10 @@
           <driver>com.mysql.jdbc.Driver</driver>
           <url>jdbc:mysql://localhost:3306/lims?characterEncoding=UTF-8</url>
           <outOfOrder>true</outOfOrder>
+          <locations>
+            <location>classpath:db.migration</location>
+            <location>classpath:uk.ac.bbsrc.tgac.miso.db.migration</location>
+          </locations>
         </configuration>
       </plugin>
     </plugins>
--- a/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/db/migration/V0168__StatusXmlToRunTable.java
+++ b/sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/db/migration/V0168__StatusXmlToRunTable.java
@@ -1,0 +1,138 @@
+package uk.ac.bbsrc.tgac.miso.db.migration;
+
+import java.io.StringReader;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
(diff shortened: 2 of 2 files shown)


tonydebat:
Heather I'm not able to reproduce this problem, but I did update the location configuration.

callunity:
I'll give this a try and report back. Hopefully someone else will be able to try this out as well, to see if it's just me!

callunity:
```
[INFO] Migrating schema `lims` to version 0169 - StatusXmlToRunTable (out of order)
[ERROR] Migration of schema `lims` to version 0169 failed! Please restore backups and roll back database and code!
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 3.556 s
[INFO] Finished at: 2017-02-17T10:59:00-05:00
[INFO] Final Memory: 18M/740M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.flywaydb:flyway-maven-plugin:3.2.1:migrate (default-cli) on project sqlstore: org.flywaydb.core.api.FlywayException: Migration failed ! Table 'lims.STATUS' doesn't exist -> [Help 1]
```
At the time of failure, the `Run` table has the extra columns added and the `Status` table does indeed exist (but maybe the difference in capitalization as compared to the STATUS in the error is significant?).

tonydebat:
I've corrected the case of the SQL statements.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/db/migration/V0169__StatusXmlToRunTable.java
```diff
@@ -0,0 +1,138 @@
...
+      String key = rs.getString("runName");
+      String value = rs.getString("xml");
+      result.put(key, value);
+    }
+    return result;
+  }
+
+  private void addStatusColumnsToRun(Connection connection) throws SQLException {
+    List<String> newColumnSql = Lists.newArrayList();
+    newColumnSql.add("ALTER TABLE Run ADD COLUMN numCycles int(11) DEFAULT NULL;");
```
apmasell: These should go in a separate table `RunIllumina`. We could leave it like this and follow it up with a regular migration to massage the tables. In which case, let's call this 168 and that one 169.

callunity: That should probably be 178 and 179, as Tuesday's release created V170.

apmasell: Yeah. Actually, we're going to have to fix that when we integrate the branch...

PR #708, opened 2017-02-17
Add table changes for new run model

apmasell:
JIRA: GLT-1515


sqlstore/src/main/resources/db/migration/afterMigrate.sql
```diff
@@ -242,34 +242,87 @@ FOR EACH ROW
...
         CASE WHEN NEW.description <> OLD.description THEN 'description' END,
         CASE WHEN (NEW.filePath IS NULL) <> (OLD.filePath IS NULL) OR NEW.filePath <> OLD.filePath THEN 'filePath' END,
         CASE WHEN NEW.name <> OLD.name THEN 'name' END,
         CASE WHEN NEW.pairedend <> OLD.pairedend THEN 'pairedend' END,
-        CASE WHEN (NEW.platformRunId IS NULL) <> (OLD.platformRunId IS NULL) OR NEW.platformRunId <> OLD.platformRunId THEN 'platformRunId' END,
-        CASE WHEN (NEW.sequencerReference_sequencerReferenceId IS NULL) <> (OLD.sequencerReference_sequencerReferenceId IS NULL) OR NEW.sequencerReference_sequencerReferenceId <> OLD.sequencerReference_sequencerReferenceId THEN 'sequencerReference_sequencerReferenceId' END,
-        CASE WHEN (NEW.status_statusId IS NULL) <> (OLD.status_statusId IS NULL) OR NEW.status_statusId <> OLD.status_statusId THEN 'status_statusId' END), ''),
+        CASE WHEN (NEW.sequencerReference_sequencerReferenceId IS NULL) <> (OLD.sequencerReference_sequencerReferenceId IS NULL) OR NEW.sequencerReference_sequencerReferenceId <> OLD.sequencerReference_sequencerReferenceId THEN 'sequencerReference_sequencerReferenceId' END), ''),
```
callunity: `health` shoudl be added to the log_message as well, and I think `completionDate` should be added here since it's possible to change the completion date when changing the status.

apmasell: Fixed. I think I've gotten everything, but I'm going to exclude metrics since that will be updated constantly while the sequencer is running.

tonydebat:
After deploying with `prod` db:

- The run list works, which is correct.
- Clicking on a sequencer reveals the error `Property 'status' not found on type uk.ac.bbsrc.tgac.miso.core.data.IlluminaRun`. The same error appears when clicking on a sample that has participated in a run.
  - http://localhost:8080/miso/stats/sequencer/6
  - http://localhost:8080/miso/sample/401
- Clicking on run reveals a `NullPointerException`
  - http://localhost:8080/miso/run/1774

apmasell:
I've fixed the `status` problem on the sequencer and sample pages (it also occurred on the library page). Still figuring out the NPE.

PR #711, opened 2017-02-21
GLT-1526 add pools to lanes

callunity:


callunity:
I will cherry-pick these changes onto the notification server branch and make the necessary changes there.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SequencerPartitionContainerImpl.java
```diff
@@ -59,15 +71,56 @@
...
+  private String identificationBarcode;
+  private String locationBarcode;
+
+  @ManyToMany(targetEntity = RunImpl.class)
+  @JoinTable(name = "Run_SequencerPartitionContainer", joinColumns = {
+      @JoinColumn(name = "containers_containerId") }, inverseJoinColumns = {
+          @JoinColumn(name = "Run_runId") })
+  private Collection<Run> runs = null;
+
+  @ManyToOne(targetEntity = SecurityProfile.class, cascade = CascadeType.ALL)
```
djcooke: SecurityProfile on PartitionImpl has CascadeType.PERSIST, and here is CascadeType.ALL. It should be the same everywhere

miso-web/pom.xml
```diff
@@ -252,95 +252,95 @@
...
-              </artifactItems>
-              <excludes>META-INF</excludes>
-              <outputDirectory>target/ROOT/scripts/fluxion-ajax/</outputDirectory>
-              <overWriteReleases>true</overWriteReleases>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
-    </plugins>
+      <plugins>
```
djcooke: Did the formatter do this? Changes in this file are indentation-only, and the new indentation is incorrect

callunity: Hmm yes -- I had originally wrapped the plugins in a `<pluginManagement>` tag but that caused all the tests to fail and the JS/CSS concatenation to fail so I took it out, but apparently the indentation remains. I'll revert the indentation.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SequencerPartitionContainerImpl.java
callunity: Hmm. I'll change that.

PR #712, opened 2017-02-21
GLT-1502 make changelog for changing indices

callunity:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
```diff
@@ -388,6 +395,53 @@ public void deleteQc(Library library, Long qcId) throws IOException {
...
+   * 
+   * @param originalIndices
+   * @param updatedIndices
+   * @param target
+   * @throws IOException
+   */
+  private void makeChangeLogForIndices(List<Index> originalIndices, List<Index> updatedIndices, Library target) throws IOException {
+    
+    Set<String> added = new TreeSet<>(stringifyIndices(updatedIndices));
+    added.removeAll(originalIndices);
```
apmasell: You're removing `Index` objects from a set containing strings. You probably want to do:
```
Set<String> original  = stringifyIndices(originalIndices);
Set<String> updated  = stringifyIndices(updatedIndices);
Set<String> added = new TreeSet<>(updated);
added.removeAll(original);
Set<String> removed = new TreeSet<>(original);
removed.removeAll(updated);
```

callunity: Ah, I could see how that wouldn't work well. I'll make those changes.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
```diff
@@ -388,6 +395,55 @@ public void deleteQc(Library library, Long qcId) throws IOException {
...
+    Set<String> added = new TreeSet<>(updated);
+    added.removeAll(originalIndices);
+    Set<String> removed = new TreeSet<>(original);
+    removed.removeAll(updatedIndices);
+
+    if (!added.isEmpty() || !removed.isEmpty()) {
+      StringBuilder message = new StringBuilder();
+      message.append("Indices");
+      LimsUtils.appendSet(message, removed, "removed");
+      LimsUtils.appendSet(message, added, " Added");
```
djcooke: This will result in
* when removing only:
`Indices removed: Family - Index 01 (AAAAAA), Family - Index 02 (CCCCCC)`
* when adding only:
`Indices  Added: Family - Index 01 (AAAAAA), Family - Index 02 (CCCCCC)`
* when adding and removing:
`Indices removed: Family - Index 01 (AAAAAA), Family - Index 02 (CCCCCC) Added: Family - Index 03 (GGGGGG), Family - Index 04 (TTTTTT)`

It's not visible here, but there's an extra space before "Added" in both cases since `LimsUtils#appendSet` also adds a space. The space here should be removed. "Added" should match "removed" (no capital), and a semicolon would be nice to separate removals from additions in cases where both happen

PR #714, opened 2017-02-22
GLT-1184: fix partition numbers and ordering

djcooke:
Also
* fixes errors occurring when trying to save a new container
* removed SecurityProfiles from partitions (they were never set separately from the container)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SequencerPartitionContainerImpl.java
```diff
@@ -405,4 +371,14 @@ public ChangeLog createChangeLog(String summary, String columnsChanged, User use
     changeLog.setUser(user);
     return changeLog;
   }
+
+  @Transient
+  private final Comparator<Partition> partitionNumberComparator = new Comparator<Partition>() {
```
apmasell: This can be static instead of transient.

djcooke: It's funny that I went "why is Hibernate trying to save this?" and didn't think of that

PR #716, opened 2017-02-23
GLT-1528: require aliquot parent for detailed libraries

djcooke:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
```diff
@@ -534,6 +533,21 @@ private void validateAliasUniqueness(Library library) throws IOException {
...
+
+    if (!isAliquotSample(library.getSample())) {
+      String sc = null;
+      if (isDetailedSample(library.getSample())) {
+        DetailedSample sample = (DetailedSample) library.getSample();
+        sc = sample.getSampleClass() == null ? "not set" : sample.getSampleClass().getAlias();
+      } else {
+        sc = "Plain Sample";
+      }
+      throw new IllegalArgumentException(String.format("Sample Class '%s' is not valid for Library parent. Must be an aliquot", sc));
```
callunity: The current wording is ambiguous as to whether the sample class is the parent or the library is the parent. "...is not a valid parent for Library"?

PR #718, opened 2017-02-24
GLT-1530 move concentration to DetailedSample

callunity:


callunity:
Not intentional, but I'll be changing it with the bulk QC changes.

PR #722, opened 2017-02-24
Fix dilution and library bulk saving

callunity:


callunity:
I approve the changes I didn't write.

PR #726, opened 2017-02-27
UI updates

callunity:


miso-web/src/main/webapp/pages/editRun.jsp
```diff
@@ -487,50 +487,18 @@
...
+            <td>Location:</td>
+            <td><span id="locationBarcode">${container.locationBarcode}</span></td>
+          </tr>
+        </c:if>
+        <c:if  test="${not empty container.validationBarcode}">
+          <tr>
+            <td>Validation:</td>
+            <td><span id="validationBarcode">${container.validationBarcode}</span></td>
+          </tr>
+        </c:if>
```
callunity: Changing these on the Run page doesn't get saved.

djcooke: I think the most convenient way to fill in all container data is to wait for the Run and Container to be created via notification and then add pools and location from the Run page.

This change prevents setting location on the Run page and forces the user to the Edit Container page. Fix the save instead, unless there's a reason not to?

Same applies for validationBarcode

callunity: Hmm. I suspect the non-saving is related to the weird container behaviour, so I'm going to take this code out and apply it during those fixes instead.

PR #728, opened 2017-02-28
Fix add project overview

apmasell:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
```diff
@@ -1065,6 +1065,7 @@ public long saveProjectOverview(ProjectOverview overview) throws IOException {
         original.setLastUpdated(new Date());
         overview = original;
       }
+      overview.setLastUpdated(new Date());
```
callunity: Please remove the same `set` on 1065.

tonydebat:
Works with local db. Never ending query problem is still present when used against the production db.

PR #730, opened 2017-03-03
GLT-1561 Convert relationship between targeted sequencing and library

tonydebat:
prep kits to a many-to-many relationship.

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/kit/KitDescriptor.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/kit/KitDescriptor.java
@@ -83,7 +83,7 @@ public class KitDescriptor implements Serializable, ChangeLoggable {
   @OneToMany(targetEntity = KitDescriptorChangeLog.class, mappedBy = "kitDescriptor")
   private final Collection<ChangeLog> changelog = new ArrayList<>();
 
-  @ManyToMany(targetEntity = TargetedSequencing.class)
+  @ManyToMany
   @JoinTable(name = "TargetedSequencing_KitDescriptor", inverseJoinColumns = {
       @JoinColumn(name = "targetedSequencingId") }, joinColumns = {
           @JoinColumn(name = "kitDescriptorId") })
@@ -281,6 +281,10 @@ public class KitDescriptor implements Serializable, ChangeLoggable {
 
   public void setLastModifier(User lastModifier) {
     this.lastModifier = lastModifier;
+  }
+
+  public Collection<TargetedSequencing> getTargetedSequencing() {
+    return targetedSequencing;
   }
 
   /**
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/kit/KitDescriptor.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/kit/KitDescriptor.java
@@ -26,6 +26,7 @@ package uk.ac.bbsrc.tgac.miso.core.data.impl.kit;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 
 import javax.persistence.Entity;
 import javax.persistence.EnumType;
@@ -34,6 +35,8 @@ import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.persistence.JoinColumn;
+import javax.persistence.JoinTable;
+import javax.persistence.ManyToMany;
 import javax.persistence.ManyToOne;
 import javax.persistence.OneToMany;
 import javax.persistence.Table;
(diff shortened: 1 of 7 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/kit/KitDescriptor.java
```diff
@@ -79,6 +83,12 @@
   @OneToMany(targetEntity = KitDescriptorChangeLog.class, mappedBy = "kitDescriptor")
   private final Collection<ChangeLog> changelog = new ArrayList<>();
 
+  @ManyToMany(targetEntity = TargetedSequencing.class)
+  @JoinTable(name = "TargetedSequencing_KitDescriptor", inverseJoinColumns = {
```
djcooke: The JoinTable should only be defined on one end of the relationship (here or in TargetedSequencing). Since this is a bidirectional many-to-many, you have to choose which side is the "owner" of the relationship.

The owner keeps the `@JoinTable` definition and modifying its collection will cause relationship updates. The non-owner uses the `mappedBy` property of `@ManyToMany` and modifying its collection has no effect.

The other option is to make this a unidirectional relationship. It's useful to get TargetedSequencing options for a KitDescriptor, but is it useful to get KitDescriptors for a TargetedSequencing? If not, just remove the kitDescriptors field from TargetedSequencing.

Unidirectional relationships are simpler and in my opinion should be preferred unless bidirectional is really required

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/LibraryControllerHelperService.java
```diff
@@ -588,10 +589,12 @@ public JSONArray getTargetedSequencingTypes(Long libraryPrepKitId) throws IOExce
...
-      targetedSequencingMap.put("kitDescriptorId", targetedSequencing.getKitDescriptor().getId());
-      fullTargetedSequencingCollection.add(targetedSequencingMap);
-      if (libraryPrepKitId != null && libraryPrepKitId.equals(targetedSequencing.getKitDescriptor().getId())) {
-        targetedSequencingByKit.add(targetedSequencingMap);
+      for (KitDescriptor kitDescriptor : targetedSequencing.getKitDescriptors()) {
+        targetedSequencingMap.put("kitDescriptorId", kitDescriptor.getId());
+        fullTargetedSequencingCollection.add(targetedSequencingMap);
+        if (libraryPrepKitId != null && libraryPrepKitId.equals(kitDescriptor.getId())) {
+          targetedSequencingByKit.add(targetedSequencingMap);
+        }
```
djcooke: This method can be greatly simplified - just fetch the kitDescriptor by id and get its targetedSequencings instead of all this

sqlstore/src/main/resources/db/migration/V8000__ts_many_to_many.sql
```diff
@@ -0,0 +1,64 @@
+CREATE TABLE `TargetedSequencingTemp` (
+  `targetedSequencingId` bigint(20) NOT NULL AUTO_INCREMENT,
+  `alias` varchar(255) NOT NULL,
+  `description` varchar(255) NOT NULL,
+  `kitDescriptorId` bigint(20) NOT NULL,
```
djcooke: This column isn't needed even temporarily, so you could just not create it. Doesn't really matter though

sqlstore/src/main/resources/db/migration/V8000__ts_many_to_many.sql
```diff
@@ -0,0 +1,64 @@
...
+-- StartNoTest
+
+-- Delete foreign keys if they exist.
+
+set @var=if((SELECT true FROM information_schema.TABLE_CONSTRAINTS WHERE
+            CONSTRAINT_SCHEMA = DATABASE() AND
+            TABLE_NAME        = 'LibraryDilution' AND
+            CONSTRAINT_NAME   = 'FK_ld_targetedSequencing_targetedSequencingId' AND
+            CONSTRAINT_TYPE   = 'FOREIGN KEY') = true,'ALTER TABLE LibraryDilution
+            DROP FOREIGN KEY FK_ld_targetedSequencing_targetedSequencingId','select 1');
```
djcooke: Haha. I like it

callunity:
I think this PR or one shortly after it will need to include changes to `ValueTypeLookup`, as it's still expecting a unique key for the TS aliases (so I can't test-run a migration).

djcooke:
@callunity: the root purpose of this PR is to prevent the duplication that is a problem for ValueTypeLookup. Now that you mention it, there should be a `UNIQUE` key on `TargetedSequencing.alias` to enforce that though

tonydebat:
@djcooke I've made the changes. Also, `TargetedSequencing.alias` is unique.

tonydebat:
@callunity As Dillan mentioned, this PR ensures that TargetedSequences are unique. This needs to be applied before the `oicr` PR that contains all the GSLE TargetedSequencing values. https://github.com/oicr-gsi/miso-lims/pull/148

PR #731, opened 2017-03-03
Initial changes for QCs

callunity:


sqlstore/src/main/resources/db/migration/V8201__archived_qcTypes.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE QCType ADD COLUMN archived BOOLEAN NOT NULL DEFAULT FALSE;
+ALTER TABEL QCType ADD COLUMN precision INT NOT NULL DEFAULT 0;
```
djcooke: TABEL!

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
```diff
@@ -363,15 +361,13 @@ public void addQc(Library library, LibraryQC qc) throws IOException {
...
-    try {
-      qc.setLibrary(managed);
-      managed.addQc(qc);
-    } catch (MalformedLibraryException e) {
-      throw new IOException("Malformed Library " + managed.getName() + " in the database...", e);
-    } catch (MalformedLibraryQcException e) {
-      throw new IOException("Malformed Library QC");
+
+    // update concentration if QC is of relevant type
+    if ("QuBit".equals(qc.getQcType().getName())) {
```
djcooke: Would be better if there were a field in the QcType table to indicate purpose=concentration rather than hard-coding something that may or may not exist in the DB. Maybe this is okay temporarily

callunity: Yes, @apmasell proposed making the same change. How about I delete this for now and we can re-add it later if it's necessary.

PR #734, opened 2017-03-06
Add size to library

apmasell:


sqlstore/src/main/resources/db/migration/V8801__library_size.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE Library ADD COLUMN `size` bigint(20);
+
+UPDATE Library SET size = (SELECT insertSize FROM LibraryQC WHERE LibraryQC.library_libraryId = Library.libraryId ORDER BY qcDate LIMIT 1);
```
djcooke: We're okay with assuming that the latest QC has the correct size? This affects EI more than us

apmasell: That's the ERA submission code assumes.

djcooke: So you're agreeing that it's probably not a good idea? :grin:

apmasell: The insert sizes are preserved in the migration. This is just picking “the best” one and it can be changed.

djcooke: Did not realize that a previous migration preserves old insert sizes separately, so the data will never be removed automatically during flyway migration, so this assumption is okay

apmasell:
Renamed!

PR #739, opened 2017-03-07
Screen out bad insert sizes

apmasell:


sqlstore/src/main/resources/db/migration/V8801__library_size.sql
```diff
@@ -1,3 +1,3 @@
 ALTER TABLE Library ADD COLUMN dnaSize bigint(20);
 
-UPDATE Library SET dnaSize = (SELECT insertSize FROM LibraryQC WHERE LibraryQC.library_libraryId = Library.libraryId ORDER BY qcDate LIMIT 1);
+UPDATE Library SET dnaSize = (SELECT insertSize FROM LibraryQC WHERE LibraryQC.library_libraryId = Library.libraryId AND (insertSize IS NOT NULL OR insertSize != 0) ORDER BY qcDate LIMIT 1);
```
djcooke: Should this be `(insertSize IS NOT NULL AND insertSize != 0)` (`AND` instead of `OR`)?

apmasell: Yes. Facepalm.

PR #743, opened 2017-03-08
Ignore swap files, logs, WARs, and generated migrations

apmasell:


.gitignore
```diff
@@ -11,6 +15,10 @@ target/
 # Flyway database migration credentials file.
 sqlstore/flyway.properties
 
+# Generated migrations
+sqlstore/src/main/resources/db/migration/afterMigrate.sql
+sqlstore/src/main/resources/db/migration/beforeMigrate.sql
```
djcooke: I didn't notice this before. Shouldn't they generate into `target` rather than `src`?

apmasell: Fixed.

apmasell:
It seems the `main/resources` directory doesn't get packaged. I've moved it to `classes`.

PR #744, opened 2017-03-08
GLT-850: attempt to generate non-standard migration aliases

djcooke:


migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/util/OicrMigrationNamingScheme.java
```diff
@@ -44,4 +56,39 @@ private ValidationResult validateAdditionalPatterns(String value) {
 
   }
 
+  private static class OicrMigrationSampleAliasGenerator extends OicrSampleAliasGenerator {
+
+    @Override
+    public String generate(Sample sample) throws MisoNamingException {
+      // attempt to ignore up to two levels of non-standard aliases
```
apmasell: Why two?

djcooke: The second level is needed for MDT migration. Depending on child class and hierarchy, the child portion of the alias is appended to either the parent's or the grandparent's alias.

For example, a Stock's alias uses the Tissue alias as a base, but there may be a TissueProcessing in the middle. Aliquot aliases also use the Tissue alias as a base, and that should always be their grandparent. A third level could potentially be needed (multiple TissueProcessing levels), but it's unlikely, so not bothering for now

apmasell: Okay. Please paste that in a comment.

djcooke: Updated comment

djcooke:
Added check to throw exception if sample somehow still saved with null alias

PR #745, opened 2017-03-09
Bulk create sample QCs

callunity:


miso-web/src/main/webapp/scripts/sample_hot.js
```diff
@@ -615,6 +615,19 @@ Sample.hot = {
...
+   * Checks if the first sample in the sample being created/updated is an RNA sample
+   */
+  isFirstSampleRNA: function () {
+    var firstSample;
+    if (Sample.hot.createOrEdit == 'Create') {
+      firstSample = Sample.hot.newSamplesJSON[0];
+    } else if (Sample.hot.createOrEdit == 'Edit') {
+      firstSample = Sample.hot.samplesJSON[0];
+    }
+    return firstSample.sampleClassAlias.indexOf("RNA") != -1;
```
apmasell: We might want to create a table of SampleClass to QC relations. Not for this request though.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SampleQCImpl.java
```diff
@@ -135,4 +136,22 @@ public int hashCode() {
...
+      if (getId() > s.getId()) return 1;
+    } else if (getQcType() != null && s.getQcType() != null && getResults() != null && s.getResults() != null && getQcDate() != null
+        && s.getQcDate() != null) {
+      int type = getQcType().compareTo(s.getQcType());
+      if (type != 0) return type;
+      int results = getResults().compareTo(s.getResults());
+      if (results != 0) return results;
+      int creator = getQcDate().compareTo(s.getQcDate());
+      if (creator != 0) return creator;
+    }
```
djcooke: These compareTo(QC) could be abstracted a bit: If it's okay to compare date before data, just make the impl classes do `int compare = super.compareTo(t); if (compare == 0) {// impl-specific compare} else {return 0}`

If date really should be compared after data, make AbstractQc do the generic stuff, then call an abstract "compareQcData" method, which is overridden in all of the impl classes to only compare implementation-specific things. If that's still 0, do the date comparison.

May want to move null checks around too (it doesn't matter if qcDate is null if it ends up basing the response on qcType)

Just thoughts, no change necessarily needed. As is, it's better than before

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -1327,13 +1321,7 @@ private static QcTypeDto asDto(QcType from) {
 
   private static QcType to(QcTypeDto from) {
     QcType to = new QcType();
-    to.setQcTypeId(from.getId());
-    to.setName(from.getName());
-    to.setDescription(from.getDescription());
-    to.setQcTarget(from.getQcTarget());
-    to.setUnits(from.getUnits());
-    to.setPrecisionAfterDecimal(from.getPrecisionAfterDecimal());
-    to.setArchived(from.isArchived());
```
djcooke: All fields on the DTO should still be included in the conversion method for consistency and predictability

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryDilutionService.java
```diff
@@ -204,7 +204,7 @@ private void loadChildEntities(LibraryDilution dilution) throws IOException {
     if (dilution.getLibrary() != null) {
       dilution.setLibrary(libraryDao.get(dilution.getLibrary().getId()));
     }
-    if (dilution.getTargetedSequencing() != null) {
+    if (dilution.getTargetedSequencing() != null && dilution.getTargetedSequencing().getId() != null) {
```
djcooke: Is there a reason the targeted sequencing is set and its id isn't? Seems like something an exception should be thrown for

callunity: I'll fix the root problem instead, which is that the DilutionDto->Dilution method was always creating a targeted sequencing and setting it on the dilution, even when the DilutionDto.targetedSequencingId was empty.

PR #748, opened 2017-03-10
Kill scriptaculous

apmasell:


apmasell:
I've fixed the `setBaseUrl` and `doOnSuccess` problem.
The used pools stuff doesn't seem to actually use Fluxion, so I'm not sure what's going on there.
Making the list pages slower is outside the scope of this PR.

apmasell:
The ready to run page is legitimately empty on the prod DB. If you add a pool to a run, it shows up there.

callunity:
Huh, that's interesting about the ready to run page. I wonder when that broke.

miso-web/src/main/webapp/pages/external/externalHeader.jsp
```diff
@@ -118,7 +118,6 @@
 </table>
 
 <script type="text/javascript">
-    setBaseUrl("<c:url value="/miso/external/"/>");
```
callunity: This is important! Keep this one!

callunity: Or rather, this does need to be distinct from the base MISO URL.

callunity: (That being said, the login page external webapp doesn't seem to be complaining about this line being deleted so I will approve it)

apmasell: Looking at the deleted code, this only matters for the `activityInput.js` URLs, which we never used.

PR #755, opened 2017-03-13
Fix project name sorting

apmasell:
JIRA: GLT-1538


miso-web/src/main/webapp/scripts/project_ajax.js
```diff
@@ -193,7 +192,7 @@ Project.ui = {
             "bJQueryUI": true,
             "iDisplayLength": 25,
             "aaSorting": [
-              [0, "desc"]
+              [0, "asc"]
```
callunity: Why the change to `asc`? I would think that the lab would want to see more recent projects first.

apmasell: Fair. It looked really weird the other way, but I guess we sort everything else by date.

PR #756, opened 2017-03-14
Remove insert size from Library QC

apmasell:


spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/ImportExportControllerHelperService.java
```diff
@@ -359,15 +359,23 @@ public JSONObject confirmLibrariesPoolsUpload(HttpSession session, JSONObject js
...
                   lqc.setQcCreator(user.getLoginName());
                   lqc.setQcDate(new Date());
                   lqc.setQcType(requestManager.getLibraryQcTypeByName("Qubit"));
 
+                  LibraryQC lis = new LibraryQCImpl();
+                  lis.setLibrary(library);
+                  lis.setResults((double) insertSize);
+                  lis.setQcCreator(user.getLoginName());
+                  lis.setQcDate(new Date());
+                  lis.setQcType(requestManager.getLibraryQcTypeByName("Insert Size"));
```
callunity: Our lab isn't using this, but this would fail if we tried -- our insert size QCType is called "Tape Station" because that's the name of the instrument, and we haven't got an "Insert Size" QCType in our database. I'm less fond of the option to rename Tape Station to Insert Size because if we get a new instrument to measure insert size, it doesn't capture the difference.

apmasell: I thought the migration created an `Insert Size` QCType for the existing data where we don't know the QC method.

callunity: It does, `WHERE NOT EXISTS [a qcType with target Library and units 'bp']`. Our Tape Station does have both of those so it wouldn't be created there.

apmasell: After discussion, Heather and I decided the reasonable workflow is to put the imported insert size on the library itself and not write a QC record for it. If the import changes it, then it will be recorded in the changelog.

PR #766, opened 2017-03-16
Update qcPassed based on detailedQcStatus changes

apmasell:
JIRA: GLT-1604


apmasell:
Okay. I'm worried about the case where we transition from known to unknown. Suppose this happens:
1. Set detailed QC to STR: failed (QC passed is now false).
2. Set detailed QC to waiting for consent.

At this point, the detailed QC says the QC state is unknown but the previous state is failed. I think it should remain failed. Should it flip to unknown?

apmasell:
Changed to a stateless copy.

PR #767, opened 2017-03-17
Update run's changelog when a status is changed

apmasell:
JIRA: GLT-1187

### DIFF
--- a/sqlstore/src/main/resources/db/migration_afterMigrate/05_standard.sql
+++ b/sqlstore/src/main/resources/db/migration_afterMigrate/05_standard.sql
@@ -312,23 +312,22 @@ CREATE TRIGGER StatusChange BEFORE UPDATE ON Status
 FOR EACH ROW
   BEGIN
     DECLARE log_message varchar(500) CHARACTER SET utf8;
-    DECLARE runId bigint(20);
     SET log_message = CONCAT_WS(', ',
       CASE WHEN NEW.health <> OLD.health THEN CONCAT('health: ', COALESCE(OLD.health, 'n/a'), ' → ', COALESCE(NEW.health, 'n/a')) END,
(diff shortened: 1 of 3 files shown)


sqlstore/src/main/resources/db/migration_afterMigrate/05_standard.sql
```diff
@@ -312,23 +312,22 @@ CREATE TRIGGER StatusChange BEFORE UPDATE ON Status
...
+      INSERT INTO RunChangeLog(runId, columnsChanged, userId, message) SELECT
         runId,
         COALESCE(CONCAT_WS(',',
           CASE WHEN NEW.health <> OLD.health THEN 'health' END,
           CASE WHEN (NEW.completionDate IS NULL) <> (OLD.completionDate IS NULL) OR NEW.completionDate <> OLD.completionDate THEN 'completionDate' END,
           CASE WHEN (NEW.startDate IS NULL) <> (OLD.startDate IS NULL) OR NEW.startDate <> OLD.startDate THEN 'startDate' END,
           CASE WHEN NEW.runName <> OLD.runName THEN 'runName' END), ''),
-        (SELECT lastModifier FROM Run WHERE alias = NEW.runName),
-        log_message);
+        lastModifier,
```
callunity: Where does `lastModifier` come from?

apmasell: The run itself.

callunity: Ah, I missed that it had changed from `INSERT VALUES` to `INSERT SELECT`.

callunity:
statusupdate

The `SELECT [...] lastModifier` doesn't seem to be behaving as expected.

apmasell:
I'm not sure there's a sensible way to solve this without adding a last modifier field to the status or rolling status into run. Hibernate is updating the status before the run and I can't change that.

callunity:
Are you sure that's what it is? I have updated the description and saved, and then updated the description and saved again, and it's still showing both changes as coming from the notification server. This indicates to me that the `run.setLastModifier` call in the `EditRunController` is not working properly.

apmasell:
Oh, well, that's even better.

apmasell:
Fixed.

PR #774, opened 2017-03-21
Fix container add

apmasell:
This changes the operations of the run page. Rather than trying to “replace” a container by providing a new barcode, it adds and removes containers or allows editing of a run and its containers. It also logs that to the database.


miso-web/src/main/webapp/scripts/run_ajax.js
```diff
@@ -571,6 +558,64 @@ Run.ui = {
...
+            'runControllerHelperService',
+            'addContainerByBarcode',
+            {
+              'runId': runId,
+              'barcode': jQuery('#containerBarcode').val(),
+              'url': ajaxurl
+            },
+            {
+              'doOnSuccess': function(json) {
+                if (json.err) {
```
djcooke: This can be true when `doOnSuccess` is called? I think there are other JSONUtil methods that will properly mark the response as error and invoke a different callback. Doesn't really matter, just an observation

apmasell: Sadly, it is the case. I originally had it without, but all the other calls have it this way. It is actually succeeding and returning an object that contains error information. Because Fluxion.

apmasell: Okay, different sad: there is a way but we don't use it a fair amount of the time.

tonydebat:
I get this error when attempting to save a new sequencing container. Let me know if you'ed like be to create a separate issue.
```object references an unsaved transient instance - save the transient instance before flushing: com.eaglegenomics.simlims.core.SecurityProfile; nested exception is org.hibernate.TransientObjectException: object references an unsaved transient instance - save the transient instance before flushing: com.eaglegenomics.simlims.core.SecurityProfile```

tonydebat:
I created a ticket for the stacktrace:
https://jira.oicr.on.ca/browse/GLT-1613

djcooke:
The TransientObjectException will be fixed in GLT-1548

PR #775, opened 2017-03-21
GLT-1611 Notification server uses RunImpl instead of subclasses

tonydebat:
Notification server failed because hibernate does not know about the
RunImpl subclasses. Modified to use RunImpl rather than subclasses. Will switch
to subclasses when the Notification Server is revisited.

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
@@ -1106,6 +1106,7 @@ public class MisoRequestManager implements RequestManager {
       run.getStatus().setInstrumentName(run.getSequencerReference().getName());
 
       if (run.getId() == AbstractRun.UNSAVED_ID) {
+        run.setSecurityProfile(securityProfileStore.get(securityProfileStore.save(run.getSecurityProfile())));
         run.setLastModifier(getCurrentUser());
         run.getStatus().setLastUpdated(new Date());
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
```diff
@@ -1185,6 +1186,31 @@ public void saveRuns(Collection<Run> runs) throws IOException {
     }
   }
 
+  private void setValuesBeforeSavingNewRun(Run run) throws IOException {
```
djcooke: This method does the same as `saveRun` when the run is unsaved. Can just call `saveRun` for all runs in `saveRuns` instead of separating them

notification-consumer-services/src/main/java/uk/ac/bbsrc/tgac/miso/notification/consumer/service/mechanism/IlluminaNotificationMessageConsumerMechanism.java
```diff
@@ -147,18 +149,23 @@ public void setAttemptRunPopulation(boolean attemptRunPopulation) {
...
               r.setAlias(runName);
               r.setFilePath(runName);
               r.setDescription(m.group(3));
               r.setPairedEnd(false);
               is.setHealth(ht);
               r.setStatus(is);
+              r.setPlatformType(PlatformType.ILLUMINA);
             } else {
               String xml = run.getString(IlluminaTransformer.JSON_STATUS);
-              is = new IlluminaStatus(xml);
```
djcooke: parsing from IlluminaStatus needs to be done somewhere still

tonydebat: It's done in ```((RunImpl) r).illuminaRunConfig(xml, null);```.   That's on line189 of ```RunImpl.java```.

I think a reference of this new instance of IlluminaStatus will need be exposed because it's referenced later in ```IlluminaNotificationMessageConsumerMechanism.java```.

tonydebat:
@djcooke Fixup for both your comments.

tonydebat:
@apmasell Yes, ```RunImpl.illuminaRunConfig``` runs ``IlluminaStatus.parseStatusXml``` when it creates an ```IlluminaStatus``` on line 189.

apmasell:
Wait, if it creates `IlluminaStatus` object, won't it fail to save?

tonydebat:
Will fix

tonydebat:
Fixed.

PR #791, opened 2017-03-31
Targeted sequencing fixes

apmasell:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/kit/KitDescriptor.java
```diff
@@ -81,13 +82,33 @@
...
+
+  public void setChangelog(Collection<ChangeLog> changelog) {
+    this.changelog = changelog;
+  }
+
+  public void setTargetedSequencing(Set<TargetedSequencing> targetedSequencing) {
+    this.targetedSequencing = targetedSequencing;
+  }
+
+  public long getKitDescriptorId() {
```
djcooke: There are already `getId()` and `setId(Long)` methods in this class

tonydebat:
When I use a single library form to generate a library dilution the number of available targeted sequencing values is restricted by the library kit:
image
When I propagate the same library into a library dilution I'm shown a large list of targeted sequencing values that are not restricted by the library kit. Also, the list contains duplicates.
image

tonydebat:
Uggg... and then I cleared my cache. 👍

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/kit/KitDescriptor.java
apmasell: Sorry. Over eager autogeneration.

PR #794, opened 2017-03-31
Kill expensive columns

apmasell:


miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -251,6 +251,44 @@ public static DetailedQcStatus to(DetailedQcStatusDto from) {
     return to;
   }
 
+  public static SampleDto asMinimalDto(Sample from) {
+    DetailedSampleDto dto = new DetailedSampleDto();
+
+    if (isDetailedSample(from)) {
+      dto.setSampleClassId(((DetailedSample) from).getSampleClass().getId());
+    }
+    dto.setId(from.getId());
```
djcooke: Should put all the plain sample setters into a separate method and call that method from both here and `asDto(Sample)` to avoid duplication

PR #796, opened 2017-04-04
Create a resuable interface for generating jQuery tables

apmasell:
This goes all the way down to Hibernate and handle searches and project restrictions

JIRA: GLT-983


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/security/AuthorizedPaginatedDataSource.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Override
+  public long count() throws IOException {
+    return backingSource.count();
+  }
+
+  @Override
+  public List<T> list(int offset, int limit, boolean sortDir, String sortCol) throws IOException {
+    return authorizationManager.filterUnreadable(backingSource.list(offset, limit, sortDir, sortCol));
```
tonydebat: If all the items in the returned window are unreadable for the user then an empty list will be returned. There could be more readable items in the collection though. Is this scenario handled?

djcooke:
Had a quick look and I like the design except for the transaction scope. What would you think about creating a 'PagableQueryOptions' object to hold request-specifics such as limit, offset, queryString, etc, and making the PageableDataSource a static component of the DAO? You could get rid of the authorization layer and let Services handle auth as normal.

The JQueryTableBackend would then pass the options to Service methods instead of talking directly to the PageableDataSource. It would look something like:

```
// Controller
private static final JQueryTableBackend<?> jQueryBackend = new JQueryTableBackend() {
  @autowired XService service;

  getAll(request, response, uriBuilder) {
    PageableQueryOptions opts = new PageableQueryOptions(...); // or could use builder pattern
    ...
    List<?> items = service.list(opts);
    int count = service.count(opts);
    ...
  }
}

// Service
list(opts) {return authorizationManager.filterUnreadable(dao.list(opts))}
count(opts) {return dao.count(opts)}

// DAO
private static final PaginatedDataSource paginatedSource = ...

list(opts) {return paginatedSource.list(opts)}
list(opts) {return paginatedSource.count(opts)}
```

Another problem is that the count is not limited to authorized entities, but it's a problem that already exists, and I don't think there's a good solution since the point of pagination is to avoid loading all entities, which would be necessary for auth checks.

apmasell:
Okay. I'd be fine with that. I'll create a interface that can be used for both the service and the DAO. It will require creating at least stub services for the things that are in `RequestManager`.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/security/AuthorizedPaginatedDataSource.java
apmasell: Nope. It's also not handled now on the list pages. Without putting authorization checks into Hibernate, I don't know who we'd do it.

apmasell:
I've updated the interfaces to avoid transactions in the controller layer. I've also paginated the libraries list.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateLibraryDao.java
```diff
@@ -305,58 +306,15 @@ public LibraryStrategyType getLibraryStrategyTypeByName(String name) throws IOEx
...
-    // want to sort most List X pages
-    criteria.createAlias("derivedInfo", "derivedInfo");
-    criteria.setFirstResult(offset);
-    criteria.setMaxResults(limit);
-    criteria.addOrder("asc".equalsIgnoreCase(sortDir.toLowerCase()) ? Order.asc(sortCol) : Order.desc(sortCol));
-    @SuppressWarnings("unchecked")
-    List<Library> records = criteria.list();
-    return records;
-  }
+  private final static List<String> STANDARD_ALISESES = Arrays.asList("derivedInfo", "sample");
```
apmasell: Fixed.

apmasell:
Yeah. The library dilutions also have to fixed. I'm working on those.

PR #799, opened 2017-04-06
Paginate samples list on edit project page

apmasell:
JIRA: GLT-983


tonydebat:
Deploys and Project page is visible, but sections do not load. In the image below the sample section is empty.

image

tonydebat:
And then I cleared the cache...

PR #800, opened 2017-04-07
GLT-1637 Modify how hibernate loads collections

tonydebat:
Marked smaller collections as subselect so they will be completely
loaded on the first reference. Marked larger collections as batch so items
will be retrieved in chunks instead of one at a time.


tonydebat:
Dillan, I've made changes.

PR #804, opened 2017-04-13
Display duplicated barcodes to the users on the main page

apmasell:


apmasell:
This is much too slow with the big database, so I'm going to close it and decide what to do later.

PR #807, opened 2017-04-18
Paginate orders

apmasell:


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/BaseHibernatePaginatedDataSource.java
```diff
@@ -115,10 +110,30 @@ default Criteria createPaginationCriteria() throws IOException {
...
   Criterion searchRestrictions(String query);
 
-  /**
-   * Add additional restrictions to the {@link Criteria} object. Note, during listing, a collection of IDs is fetched and then these are
-   * used to perform a second query. In that instance, this method is <b>not</b> called for the ID-driven fetch.
-   */
-  default void setAdditionalPaginationCriteria(Filter filter, Criteria criteria) {
+  @Override
+  default void setFulfilled(Criteria item, boolean isFulfilled) {
+    throw new IllegalArgumentException();
```
djcooke: I didn't really see the problem with your previous design of filters, but this seems worse to me. This filter only makes sense for orders, but could be set on anything and will be a runtime error. This will get much more confusing when we add advanced search and many more fields need to be filterable

apmasell: Some of the options in the basic `PaginationFilter` wasn't going to work for dilutions. It basically means that I'd have to have totally separate pagination filters for many things and the common-but-not-universal elements of each one would have to be reimplemented independently. I'd rather not have a separate filter class for each model.

There's also the case that some objects require special considerations for different filters. For instance, runs filtered by project require a bunch of aliases that prevent empty runs from showing up in other contexts, so there was a general method before that became “here's a filter and a criteria, patch any mistakes.” In this design, that gets handled by inheritance of the `Sink` interface with puts all that logic in one places rather than spooky action at a distance.

There really isn't a way to make this type-safe as we want to add more conditions (which is already happening). Moreover, `Criteria` already prefers runtime time errors for things we know at compile time. Actually, in the general design of MISO, we've traded compile time errors for runtime errors.

djcooke: I see some of your points, but they can be addressed while still having a separate filter class for each entity type which have their unique fields separate from the rest.

I'm not sure what you mean about trading compile time errors for runtime since the former should always be preferable. If you mean checked vs. unchecked Exceptions, that's not what I'm talking about. What I mean is that it should not be possible to set a "fulfilled" property on a filter for Samples since that property is only applicable to Orders.

apmasell:
Okay. I've created a ticket for the follow up. https://jira.oicr.on.ca/browse/GLT-1665

PR #808, opened 2017-04-18
Evade Hibernate proxies when doing heirarchy checks

apmasell:
JIRA: GLT-1657

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractLibrary.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractLibrary.java
@@ -12,11 +12,11 @@
  *
  * MISO is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
- * along with MISO.  If not, see <http://www.gnu.org/licenses/>.
+ * along with MISO. If not, see <http://www.gnu.org/licenses/>.
  *
  * *********************************************************************
  */
@@ -506,7 +506,6 @@ public abstract class AbstractLibrary extends AbstractBoxable implements Library
     this.lastModifier = lastModifier;
   }
 
-
   @Override
   public Collection<ChangeLog> getChangeLog() {
     return changeLog;
@@ -523,17 +522,6 @@ public abstract class AbstractLibrary extends AbstractBoxable implements Library
       }
     }
     return IndexFamily.NULL;
-  }
-
-  @Override
-  public SampleTissue getSampleTissue() {
-    if (this.getSample() instanceof DetailedSample) {
-
-      for (DetailedSample parent = (DetailedSample) this.getSample(); parent != null; parent = parent.getParent()) {
-        if (parent instanceof SampleTissue) return (SampleTissue) parent;
-      }
-    }
-    return null;
   }
 
   @Override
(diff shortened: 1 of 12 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/PartitionImpl.java
```diff
@@ -140,8 +140,8 @@ public void setPool(Pool pool) {
   public boolean equals(Object obj) {
     if (obj == null) return false;
     if (obj == this) return true;
-    if (!(obj instanceof PartitionImpl)) return false;
-    PartitionImpl them = (PartitionImpl) obj;
+    if (!(obj instanceof Partition)) return false;
```
djcooke: All of these should continue to use the implementation class to increase the likelihood that `one.equals(two) == two.equals(one)`

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractLibrary.java
```diff
@@ -539,7 +538,7 @@ public SampleTissue getSampleTissue() {
     if (this.getSample() instanceof DetailedSample) {
 
       for (DetailedSample parent = (DetailedSample) this.getSample(); parent != null; parent = parent.getParent()) {
-        if (parent instanceof SampleTissue) return (SampleTissue) parent;
+        if (parent.getSampleClass().getSampleCategory().equals(SampleTissue.CATEGORY_NAME)) return (SampleTissue) parent;
```
djcooke: This method doesn't seem to be used for anything, so should be removed; otherwise, null checks are needed

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/LimsUtils.java
```diff
@@ -650,25 +651,29 @@ public static boolean isPlainSample(Sample sample) {
 
   public static boolean isIdentitySample(Sample sample) {
     if (!isDetailedSample(sample)) return false;
-    return sample instanceof Identity;
+    return sample instanceof Identity || ((DetailedSample) sample).getSampleClass().getSampleCategory().equals(Identity.CATEGORY_NAME);
```
djcooke: Needs null checks. Same in below methods

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/PartitionImpl.java
apmasell: Yes, but one might be a proxy. So, either it has to be deproxied (and the models know about Hibernate), or they should just check using the interface.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/AbstractLibrary.java
apmasell: Deleted.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/LimsUtils.java
apmasell: What's the correct behaviour if null? I think it's neither true nor false, so I'm happy with an NPE.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/PartitionImpl.java
djcooke: Oh, right...

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/LimsUtils.java
djcooke: It should return false if there's a null I'd think. Just because it's not yet defined as an Identity doesn't mean it has to be defined at all

PR #813, opened 2017-04-19
Improved box and position lookup for boxables

djcooke:


sqlstore/src/main/resources/db/migration_afterMigrate/05_standard.sql
```diff
@@ -882,3 +882,30 @@ CREATE OR REPLACE VIEW LibraryDerivedInfo AS
 CREATE OR REPLACE VIEW BoxDerivedInfo AS
   SELECT boxId, MAX(changeTime) AS lastModified FROM BoxChangeLog GROUP BY boxId;
 
+CREATE OR REPLACE VIEW SampleBoxPosition
```
apmasell: Neat!

PR #814, opened 2017-04-19
Indicate when a dilution is being added/removed from a pool

apmasell:
JIRA: GLT-1661


apmasell:
Fixed.

apmasell:
I'll add a fix for that. I thought disabled was...you know...disabled. :-/

PR #819, opened 2017-04-20
Glt 1662 dilution sort

apmasell:


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernatePoolableElementViewDao.java
```diff
@@ -0,0 +1,71 @@
...
+    return Collections.emptyList();
+  }
+
+  @Override
+  public String propertyForSortColumn(String original) {
+    return original;
+  }
+
+  @Override
+  public void setPlatformType(Criteria criteria, PlatformType platformType) {
```
djcooke: I realize this is for the PaginatedDataSource, but `dao.setPlatformType` and similar are confusing method names for what they do. Nothing to fix in this PR; just wanted to mention it

sqlstore/src/main/resources/db/migration_afterMigrate/05_standard.sql
```diff
@@ -934,9 +934,32 @@ JOIN BoxPosition bp
...
+    JOIN Library l ON l.libraryId = d.library_libraryId
+    JOIN Sample s ON s.sampleId = l.sample_sampleId
+    JOIN Project p ON p.projectId = s.project_projectId;
+
+CREATE OR REPLACE VIEW PoolableElementView_Index
+AS SELECT
+    d.dilutionId AS dilutionId,
+    i.index_indexId AS index_indexId
+  FROM LibraryDilution d
+    JOIN Library_Index i ON i.library_libraryId = d.library_libraryId;
```
djcooke: The libraryId is included in PoolableElementView, so this additional view is unnecessary. The existing Library_Index table can bridge PoolableElementView to Indices the same way it bridges Library to Indices

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernatePoolableElementViewDao.java
apmasell: Agreed. How about if they all became `restrictPaginationByX`?

sqlstore/src/main/resources/db/migration_afterMigrate/05_standard.sql
apmasell: Fixed.

PR #821, opened 2017-04-21
Improve handling of runs without set parameters

apmasell:


miso-web/src/main/webapp/scripts/run_ajax.js
```diff
@@ -70,6 +70,12 @@ var Run = Run || {
     jQuery('#sequencerReference').attr('data-parsley-error-message', 'You must select a Sequencer.');
     jQuery('#sequencerReference').attr('data-parsley-errors-container', '#sequencerReferenceError');
 
+    jQuery('#sequencingParameters').attr('class', 'form-control');
+    jQuery('#sequencingParameters').attr('required', 'true');
```
djcooke: Prod:
```
mysql> SELECT COUNT(*) FROM Run WHERE sequencingParameters_parametersId IS NULL;
+----------+
| COUNT(*) |
+----------+
|      364 |
+----------+
1 row in set (0.00 sec)
```

apmasell: I know. We're going to have to get those incrementally fixed.

PR #825, opened 2017-04-24
GLT-1667: improved load time of New/Edit Run page

djcooke:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditRunController.java
```diff
@@ -267,7 +267,6 @@ public ModelAndView setupForm(Run run, ModelMap model) throws IOException {
       } else if (run.getId() == AbstractRun.UNSAVED_ID) {
         run = new RunImpl(user);
         model.put("title", "New Run");
-        model.put("availablePools", populateAvailablePools(user));
```
apmasell: You can also delete the called method. It's not used elsewhere.

apmasell: Actually, `populateAvailablePools(PlatformType platformType, User user)`
 `populateAvailableExperiments(User user)` and `populateAvailableExperiments(PlatformType platformType, User user)` are all dead code.

djcooke: :fire:

PR #826, opened 2017-04-24
Glt 1416 advanced search

apmasell:


sqlstore/src/main/resources/db/migration_afterMigrate/05_standard.sql
```diff
@@ -940,6 +940,9 @@ AS SELECT
     d.concentration AS dilutionConcentration,
     d.identificationBarcode AS dilutionBarcode,
     d.lastUpdated AS lastModified,
+    d.creationDate AS created,
+    d.dilutionUserName AS creatorName,
+    (SELECT loginName FROM User WHERE userId = d.lastModifier) AS lastModifierName,
```
djcooke: Any reason for the subquery instead of a join?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/BoxDerivedInfo.java
```diff
@@ -26,11 +31,28 @@
   @Temporal(TemporalType.TIMESTAMP)
   private Date lastModified;
 
+  @Temporal(TemporalType.TIMESTAMP)
+  private Date created;
+
+  @OneToOne(targetEntity = UserImpl.class)
+  @JoinColumn(name = "creator", nullable = false)
+  @JsonBackReference
+  private User creator;
```
apmasell: A bunch of other user fields are annotated as such and I don't have a good testing strategy to figure out what's serialising these things. In other news, I'm now a fan of magical thinking.

sqlstore/src/main/resources/db/migration_afterMigrate/05_standard.sql
apmasell: None. The query analyser will convert it. Do we have a preference in general?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/BoxDerivedInfo.java
djcooke: The BackReference here only makes sense if `User` has a "boxDerivedInfos" collection

sqlstore/src/main/resources/db/migration_afterMigrate/05_standard.sql
djcooke: No, it's probably just me :grin:

djcooke: Looks like we should prefer a join. Ready to run pools request was taking 7.5 seconds with the subquery, but is down to 1.5 seconds when I changed it to a join. Will include the change in GLT-1682

PR #830, opened 2017-04-26
Try to make derived info snappier while still providing creator

apmasell:


tonydebat:
I get sample list down from 4.5s to 3.0s. And library list down from 3.0s to 1.5s.

PR #831, opened 2017-05-01
Allow seaches to operate on “ago” ranges

apmasell:
JIRA: GLT-1416

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/AgoMatcher.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/AgoMatcher.java
@@ -1,0 +1,27 @@
+package uk.ac.bbsrc.tgac.miso.core.util;
+
+import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/AgoMatcher.java
```diff
@@ -0,0 +1,22 @@
...
+public class AgoMatcher {
+  private final Pattern pattern;
+  private final int seconds;
+
+  AgoMatcher(String unit, int seconds) {
+    pattern = Pattern.compile("^([0-9]+)" + unit + "s?$", Pattern.CASE_INSENSITIVE);
+    this.seconds = seconds;
+  }
+
+  public int matches(String text) {
```
djcooke: Please rename to something that conveys that this does more than check for a match. Preferably Javadoc to make it more clear (also gets number of seconds, which can't be 0 because that means no match)

PR #832, opened 2017-05-01
GLT-1684: allow editing (single) sample subproject

djcooke:


tonydebat:
Hi Dillan, this change allows the sub-project of any sample to be modified, provided that the project already has sub-projects in the database. I would like to know more about sub-projects, such as their intended use and what affect changing a sub-project should have.

djcooke:
@tonydebat: This change is important right now because it is currently impossible to set Subproject on an Identity, which the lab needs to do. Once it is possible to create Identities and set Subproject from the bulk create page, making this unmodifiable again is an option

PR #835, opened 2017-05-04
Create JavaScript constants file

apmasell:


tonydebat:
http://localhost:8080/miso/constants.js

Generates a NPE on line 337 of MenuController.java:
  `  createArray(node, "indexFamilies", indexFamilies, Fetch.forLong("id", IndexFamily::getId),`


```
 java.lang.NullPointerException
        at uk.ac.bbsrc.tgac.miso.webapp.controller.MenuController$Fetch$1.write(MenuController.java:207)
        at uk.ac.bbsrc.tgac.miso.webapp.controller.MenuController.createArray(MenuController.java:279)
        at uk.ac.bbsrc.tgac.miso.webapp.controller.MenuController.constantsScript(MenuController.java:337)
```

apmasell:
Fixed.

djcooke:
Using the timestamp on constants.js as is prevents the client cache from being used. I realize it's a small file, but if the timestamp used for url param were only date+hour instead of milliseconds, then the client would use its own cached version instead of requesting a new copy of the server's cached version on every page load

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/MenuController.java
```diff
@@ -154,4 +192,166 @@ public ModelAndView tissueOptions(ModelMap model) {
...
+      ObjectNode dto = array.addObject();
+      for (Fetch<T> writer : writers) {
+        writer.write(dto, item);
+      }
+    }
+  }
+
+  @RequestMapping(path = "/constants.js")
+  @ResponseBody
+  public String constantsScript(HttpServletResponse response) throws IOException {
```
djcooke: Why serialize into a new form instead of using the existing DTOs?

apmasell: There are a bunch of things that didn't have DTOs and were serialised manually. If there are ones that have DTOs that I missed, I can change them.

apmasell:
We probably still want the Tomcat thing for images and such. I'll remove the JS config though.

apmasell:
I let it live in the client cache for a minute. Let me know if you want a different time.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/MenuController.java
djcooke: Oh, I didn't notice DTOs were used for some. It would be nicer to create the missing DTOs. If you'd rather merge as is, can we call this temporary and make a ticket for creating and using DTOs for consistency?

djcooke:
The stuff you have in MenuController controls the server-side caching to refresh it once every 15 minutes at most, which should be a good duration.

I don't see where you've attempted to make it use the client-side cache for a minute, but it's not working. The URL param used `${timestamp.time}` in `header.jsp` is the current timestamp in milliseconds, so it will never use the client-cached version. I have it deployed and can confirm that it never uses the cached version

apmasell:
Sorry, the change didn'tget pushed. Check again.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/MenuController.java
apmasell: Fixed.

PR #836, opened 2017-05-05
Ignore unknown health types sent by notification server

apmasell:
JIRA: GLT-1718


djcooke:
I realize what you mean about not being able to test since it depends on current run state. Unit tests would be nice if this were actually in reasonably-sized units.

Are we sure the notification server should never be allowed to transition a run from Running > Unknown? This might actually indicate a change in state that we'd want to be aware of.

:+1: if we prefer to ignore that possibility until we've fixed related issues with the notification server

apmasell:
I agree. I think we should really track whether a run's status was last set by a human or NS and if a human changed it, NS should be barred from doing so.

PR #849, opened 2017-05-12
GLT-1727: fixed migration of box positions

djcooke:


tonydebat:
This has a negative impact on performance.
On my machine:
The time to list libraries increased 14s (from 36s to 50s).
The time to list samples increased 38s (from 52s to 90s).

PR #850, opened 2017-05-12
Add relations table for samples

apmasell:
JIRA: GLT-1416

### DIFF
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
@@ -741,6 +741,15 @@ public class EditSampleController {
         model.put("poolSampleMap", poolSampleMap);
         model.put("samplePools", pools);
         model.put("sampleRuns", getRunsBySamplePools(pools));
+        List<SampleDto> relations = new ArrayList<>();
+        if (LimsUtils.isDetailedSample(sample)) {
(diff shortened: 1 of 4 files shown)


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
```diff
@@ -741,6 +741,15 @@ public ModelAndView setupForm(@PathVariable Long sampleId, @PathVariable Long pr
         model.put("poolSampleMap", poolSampleMap);
         model.put("samplePools", pools);
         model.put("sampleRuns", getRunsBySamplePools(pools));
+        List<SampleDto> relations = new ArrayList<>();
+        if (sample instanceof DetailedSample) {
```
apmasell: Fixed.

apmasell:
It probably should include libraries and dilutions, but we don't have a consistent DTO for those unless they all get converted to boxables or pool views, which strips out the sample class. I guess we could make a new DTO type for this. Thoughts?

The library error is unrelated, but I'll fix it.

djcooke:
Yeah, a "SampleRelativeDto" makes sense

apmasell:
Okay. Let's go with this as a stop-gap and I'll solicit what the lab wants in such a table.

apmasell:
Mentioned in https://jira.oicr.on.ca/browse/GLT-253

tonydebat:
This sample takes 24 seconds to load. (Has a huge number of children, not typical.)
http://localhost:8080/miso/sample/40229

PR #851, opened 2017-05-15
Receive LCM Tubes and CV slides

apmasell:


apmasell:
Fixed the add single page.

miso-web/src/main/webapp/scripts/sample_ajax.js
```diff
@@ -725,13 +723,31 @@ Sample.ui = {
...
-    var selectedId = jQuery('#sampleClass option:selected').val();
-    var sampleCategory = Sample.options.getSampleCategoryByClassId(selectedId);
-    jQuery('#sampleCategory').val(sampleCategory);
-    switch (sampleCategory) {
+    var sampleClassId = parseInt(jQuery('#sampleClass option:selected').val());
+    var sampleClass = Constants.sampleClasses.filter(function (sampleClass) {
+      return sampleClass.id == sampleClassId;
+    })[0];
+    if (!sampleClass) {
+      return;
```
djcooke: Should call `Sample.ui.setUpForTissue();` as in default case below. This removes from view all fields that may not be relevant to whatever is selected next (since `(!sampleClass)` should indicate that nothing is currently selected)

apmasell: Fixed.

PR #856, opened 2017-05-17
GLT-1658: Show both location and boxLocation for Samples and Libraries

djcooke:


tonydebat:
Seems a little repetitive (from the sample list):
	Foyer644_s_F10, SANTIAGO - F10 (Foyer644_s)

PR #858, opened 2017-05-18
Avoid duplicate criteria aliases

apmasell:


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernatePaginatedDataSource.java
```diff
@@ -31,6 +44,244 @@
 @Transactional(rollbackFor = Exception.class)
 public interface HibernatePaginatedDataSource<T> extends PaginatedDataSource<T>, PaginationFilterSink<Criteria> {
 
+  class AliasTrackingCriteria implements Criteria {
```
djcooke: Should be static or extracted into another file. I'd go with the latter

apmasell: Done.

PR #859, opened 2017-05-19
bug fixes

djcooke:


tonydebat:
Works for Sample select all. Library select all has the same problem. Could fix or create a new ticket.

djcooke:
@tonydebat: created ticket already: https://jira.oicr.on.ca/browse/GLT-1753

PR #861, opened 2017-05-19
Allow saving a box that has scan errors

apmasell:
JIRA: GLT-1635

### DIFF
--- a/miso-web/src/main/webapp/scripts/box_visualization.js
+++ b/miso-web/src/main/webapp/scripts/box_visualization.js
@@ -483,122 +483,44 @@ Box.ScanDialog = function() {
   return self;
 };
 
-Box.ScanErrors = function() {
-  var self = new BoxVisual();
-
-  self.show = function(opts) {
-    self.scan = opts.scan;
-    self.size = opts.size;
-    self.data = opts.data;
-
-    jQuery('#dialogInfoAbove').html('<h1 class="warning">Scan Failed!</h1>'+
-                                    '<p>'+self.scan.errors.message +
-                                    '<br>Read error positions:</p><br>');
-    jQuery('#dialogInfoBelow').html('');
-    jQuery('#dialogDialog').dialog({
-      autoOpen: false,
-      width: Box.dialogWidth,
-      modal: true,
-      resizable: false,
-      position: [jQuery(window).width()/2 - Box.dialogWidth/2, 50],
-      buttons: {
-        "Rescan": function () {
-          Box.initScan();
-        },
-        "Cancel": function () {
-          jQuery('#dialogDialog').dialog('close');
-        }
-      }
-    });
-    self.create({
-      div: '#dialogVisual',
-      size: {
-        rows: self.size.rows,
-        cols: self.size.cols
-      },
-      data: self.data
-    });
-    jQuery('#updateSelected, #removeSelected, #emptySelected').prop('disabled', true).addClass('disabled');
-    jQuery('#dialogDialog').dialog('open');
-  };
-
-  self.getBoxItem = function(row, col) {
-    var pos = Box.utils.getPositionString(row, col);
-    var img = '/styles/images/tube_error.png';
-    if (self.scan.errors.type == 'Unknown Barcode') {
-      img = '/styles/images/tube_duplicate_barcode_error.png';
-    } 
-
-    if (jQuery.inArray(pos, self.scan.errors.errorPositions) !== -1) {
-      return new BoxItem({
-        row: row,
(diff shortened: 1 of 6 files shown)


miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -483,122 +483,44 @@ Box.ScanDialog = function() {
...
-  };
 
-  self.getBoxItem = function(row, col) {
-    var pos = Box.utils.getPositionString(row, col);
-    var img = '/styles/images/tube_error.png';
-    if (self.scan.errors.type == 'Unknown Barcode') {
-      img = '/styles/images/tube_duplicate_barcode_error.png';
-    } 
+    var message = (results.errors.length == 0 ? "<h1>Scan Success! </h1>" : '<h1 class="warning">Scan Failed!</h1>') +
+        results.errors.map(function(err) { return "<p>Well " + err.coordinates + ": " + err.message + "</p>"; }).join("") +
```
callunity: Perhaps `Position` instead of `Well`, to also cover tubes in boxes as well as wells in plates?

apmasell: Sure. I think the scanning software calls them wells, but that's immaterial

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/BoxControllerHelperService.java
```diff
@@ -690,148 +751,107 @@ public JSONObject prepareBoxScanner(HttpSession session, JSONObject json) {
...
+      box.getSize().positionStream().map(position -> {
+        BoxableView originalItem = box.getBoxables().containsKey(position) ? box.getBoxables().get(position) : null;
+        BoxableView newItem = barcodesByPosition.containsKey(position) && boxablesByBarcode.containsKey(barcodesByPosition.get(position))
+            ? boxablesByBarcode.get(barcodesByPosition.get(position)) : null;
+        if (originalItem != null && newItem != null &&
+            !newItem.getIdentificationBarcode().equals(originalItem.getIdentificationBarcode())) {
+          DiffMessage dto = new DiffMessage();
+          dto.action = "changed";
+          dto.modified = Dtos.asDto(newItem);
+          dto.modified = Dtos.asDto(originalItem);
```
apmasell: Fixed.

spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/BoxControllerHelperService.java
```diff
@@ -676,6 +682,61 @@ public JSONObject prepareBoxScanner(HttpSession session, JSONObject json) {
     }
   }
 
+  private static boolean isRealPosition(BoxScan scan, String position) {
+    return !position.equals(scan.getNoReadLabel()) && !position.equals(scan.getNoReadLabel());
```
djcooke: Checks same condition twice. Is one supposed to be `getNoTubeLabel`? Either of these conditions is still a real position, so maybe "isFilledPosition" or something would be a better method name

apmasell: Yes. Fixed.

spring/src/test/java/uk/ac/bbsrc/tgac/miso/spring/ajax/test/BoxControllerHelperServiceTest.java
```diff
@@ -138,12 +138,16 @@ public final void testGetBoxScanWithReadErrorsReturnsError() throws Exception {
...
+    when(scan.getRowCount()).thenReturn(box.getSize().getRows());
+    when(scan.getColumnCount()).thenReturn(box.getSize().getColumns());
     when(boxScanner.getScan()).thenReturn(scan);
-    when(scan.hasReadErrors()).thenReturn(true);
+    when(scan.getReadErrorPositions()).thenReturn(Arrays.asList("A01"));
+    when(requestManager.getBoxById(1L)).thenReturn(box);
 
     JSONObject json = new JSONObject();
     json.put("boxId", id);
-    assertEquals(error, boxControllerHelperService.getBoxScan(null, json));
```
apmasell: Done.

apmasell:
The history of us having row/col/column/pos/position/coordinates is a rather frustrating one made more complicated by the former `BoxPosition` object. My last attempt to fix that broke the scanner, so if we want to fix it, we need something planned.

PR #862, opened 2017-05-19
Unify all CV and HE slides into a single class

apmasell:
JIRA: GLT-1715

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Stain.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Stain.java
@@ -1,0 +1,46 @@
+package uk.ac.bbsrc.tgac.miso.core.data;
+
+import java.io.Serializable;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "Stain")
+
+public class Stain implements Serializable {
+  private static final long serialVersionUID = 1L;
+  @ManyToOne
+  @JoinColumn(name = "stainCategoryId")
+  private StainCategory category;
+  private String name;
+  @Id
+  private Long stainId;
+
+  public StainCategory getCategory() {
+    return category;
+  }
+
+  public Long getId() {
+    return stainId;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public void setCategory(StainCategory category) {
+    this.category = category;
+  }
+
+  public void setId(Long id) {
+    this.stainId = id;
+  }
+
+  public void setName(String name) {
+    this.name = name;
+  }
+}
--- a/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
+++ b/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
@@ -31,14 +31,15 @@ import uk.ac.bbsrc.tgac.miso.core.data.DetailedSample;
 import uk.ac.bbsrc.tgac.miso.core.data.Identity;
 import uk.ac.bbsrc.tgac.miso.core.data.Sample;
 import uk.ac.bbsrc.tgac.miso.core.data.SampleAliquot;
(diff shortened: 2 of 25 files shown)


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -505,6 +510,16 @@ private void loadChildEntities(Sample sample) throws IOException {
       if (sai.getSubproject() != null && sai.getSubproject().getId() != null) {
         sai.setSubproject(subProjectDao.getSubproject(sai.getSubproject().getId()));
       }
+      if (isTissueProcessingSample(sai) && sai instanceof SampleSlide) {
+        Stain originalStain = ((SampleSlide) sai).getStain();
+        Stain stain;
+        if (originalStain == null) {
+          stain = null;
+        } else {
+          stain = stainDao.get(originalStain.getId());
```
callunity: Why the StainDAO and not the StainService?

apmasell: I don't know. We seem to have no set pattern as to whether a service should access a service or a DAO.

callunity: I like the pattern where services have access to their own DAO and other services because it means we only need to implement things like filters for access in one place, instead of everywhere that might access that DAO (though that's certainly more relevant for samples than lists of stains). I think that pattern got messed up though when we were creating new services and didn't necessarily go back to pre-existing services to replace the DAO access with service access.

djcooke: Access via Service/RequestManager should always be preferred unless there's a good reason to bypass it

apmasell: Fixed.

miso-web/src/main/webapp/pages/editSample.jsp
```diff
@@ -562,6 +562,12 @@
               <td class="h">Thickness (&#181;m):</td>
               <td><form:input id="thickness" path="thickness"/></td>
             </tr>
+            <tr>
+              <td class="h">Stain:</td>
```
djcooke: Above needs fixed: `<c:when test="${sampleClass eq 'CV Slide' or sampleCategory eq 'new'}">`

apmasell: Fixed.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Stain.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "Stain")
+
+public class Stain implements Serializable {
+  private static final long serialVersionUID = 1L;
+  private String stainGroup;
```
apmasell: Probably neither. We are going to get a whole bunch of stains that come in different group (especially antibodies), so the code sorts them into groups based on these labels. Maybe we can do submenus.

djcooke: StainGroup seems to be singular both on the model and in the DB. Is this wrong? If a Stain belongs to only one StainGroup, then there are less StainGroups than Stains. If we're constraining Stains to a DB table, shouldn't StainGroups be constrained too? Especially if submenus may be necessary?

apmasell: Yes, but since we either retrieve all the stains or a stain by ID, creating a separate table for stain groups doesn't serve any purpose. The groups are going to be institute specific, so creating an enum is definitely unhelpful. The idea is that for all the antibody stains, we can set the group to be IHC so that they either appear on the same submenu or at least sort together in the drop down.

apmasell: Moved into a separate table.

PR #864, opened 2017-05-23
GLT-1756 Add logging to migration script.

tonydebat:


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/db/migration/V0168__StatusXmlToRunTable.java
```diff
@@ -27,6 +31,9 @@
  */
 public class V0168__StatusXmlToRunTable implements JdbcMigration {
 
+  private static final Logger logger = Logger.getLogger(ClassName.class.getName());
```
apmasell: Wait, isn't this the wrong class name?

PR #865, opened 2017-05-24
GLT-1652 fix dialog title text

callunity:


miso-web/src/main/webapp/pages/editRun.jsp
```diff
@@ -358,7 +358,7 @@
         </c:forEach>
       </div>
     </c:if>
-    <div id="addRunNoteDialog" title="Create new Note"></div>
+    <div id="addRunNoteDialog" title="Add Container"></div>
```
apmasell: Wait...are add container and add note sharing the same dialog? Will this made add note wrong?

callunity: Hmm, good point. It will be more helpful if I instead add a new dialog for the add container. Thanks!

miso-web/src/main/webapp/scripts/run_ajax.js
```diff
@@ -511,14 +511,15 @@ Run.ui = {
   },
 
   addContainerByBarcode: function (runId) {
-    jQuery('#addRunNoteDialog')
-      .html("<form><p>Adding a container will lose any unsaved changes!</p>" +
+    jQuery('#addContainerDialog')
+      .html("<form><p>Adding a container now will lose any unsaved changes to the run information above!</p>" +
+            "<p>If there are no unsaved changes or you wish to continue anyway, enter the container's barcode below.</p>" +
```
tonydebat: Concerned that this is too wordy and may serve to frighten the user into believing they can/are making a mistake. The previous warning is concise and says it all.

callunity: I will revert it to the previous warning.

PR #866, opened 2017-05-25
Update OicrLibraryAliasValidator regex

callunity:


djcooke:
We decided to change the Sample alias regex to be more flexible a while back. Rather than including the specific letters, it uses `[A-Zn]`. It should probably be the same here, and the LibraryType portion should likely be similar since these can be DB additions that shouldn't need a MISO update for regex purposes

callunity:
I can certainly update the Tissue Type portion to be more general. The LibraryType portion affects how bcl2fastq is run for Illumina samples, so I'd prefer to keep that restricted.

PR #868, opened 2017-05-25
Force a tissue type and origin, but allow editing

apmasell:
JIRA: GLT-1612


miso-web/src/main/webapp/scripts/sample_ajax.js
```diff
@@ -206,6 +206,7 @@ var Sample = Sample || {
         sampleClassAlias = jQuery('#sampleClassAlias').text();
       }
       switch (sampleCategory) {
+      case 'Tissue':
```
callunity: Is this intended as a fall-though to Tissue Processing?

apmasell: This is a mistake. I'll revert it.

PR #872, opened 2017-05-26
Use mouse events to access shift-key state in multi-selects

apmasell:
JIRA: GLT-1728


tonydebat:
Works for sample and library. Could not test with pools due to:
 https://jira.oicr.on.ca/browse/GLT-1778

PR #877, opened 2017-05-29
GLT-1751 santization/validation fixes

djcooke:
Only `<>&` should be always-restricted characters now. Some fields maintain stricter validation of course


djcooke:
@callunity: added validator to identificationBarcode on the bulk dilution page

PR #880, opened 2017-05-30
GLT-1782 fix Interop Metrics display

callunity:
Reverts some changes that weren't quite ready for prime time.

### DIFF
--- a/miso-web/src/main/webapp/pages/editRun.jsp
+++ b/miso-web/src/main/webapp/pages/editRun.jsp
@@ -303,7 +303,7 @@
     </div>
   </c:if>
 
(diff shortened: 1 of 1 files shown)


miso-web/src/main/webapp/pages/editRun.jsp
```diff
@@ -303,7 +303,7 @@
     </div>
   </c:if>
 
-  <c:if test="${not empty run.metrics}">
+  <c:if test="${run.health.key ne 'Failed' and run.health.key ne 'Stopped' and metrixEnabled and miso:instanceOf(run, 'uk.ac.bbsrc.tgac.miso.core.data.IlluminaRun')}">
```
djcooke: Aren't there some metrics available for failed/stopped runs?

callunity: I would think that there are, and have made a ticket to investigate that. This fix, however, reverts to display the Interop Metrics that were available before.

PR #881, opened 2017-05-30
Simplify how Handsontables and apply to libraries

apmasell:
The goal of this is to make a more uniform system for handling Handsontables so
fewer field copying mistakes will be made for Handsontables results.
Additionally, the code to load and store column values is together to make
inspecting it easier.

JIRA: GLT-1419


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -800,18 +727,33 @@ public ModelAndView setupForm(@PathVariable Long libraryId, @PathVariable Long s
...
-      List<LibraryDto> libraryDtos = new ArrayList<>();
+  private final BulkTableBackend<Library, LibraryDto, Sample> libraryBulkBackend = new BulkTableBackend<Library, LibraryDto, Sample>(
+      "Libraries", "Samples", "library") {
+
+    @Override
+    protected LibraryDto packageDto(Sample sample) {
+      LibraryDto dto;
+      if (sample instanceof DetailedSample) {
+        DetailedLibraryDto detailedDto = new DetailedLibraryDto();
+        detailedDto.setParentSampleClassId(sample.getId());
```
callunity: Should this be `sample.getSampleClassId()`?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -822,82 +764,19 @@ public ModelAndView editPropagateSamples(@PathVariable String sampleIds, ModelMa
         } else {
           hasPlain = true;
         }
-        LibraryDto library = isDetailedSampleEnabled() ? new DetailedLibraryDto() : new LibraryDto();
-        if (namingScheme.hasLibraryAliasGenerator()) {
```
callunity: Some of this needs to be kept, as EI wants to see the bulk propagate page with the library aliases already filled in.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -906,30 +785,75 @@ public ModelAndView editBulkLibraries(@PathVariable String libraryIds, ModelMap
...
+    @Override
+    protected void writeConfiguration(ObjectMapper mapper, ObjectNode config) {
+      config.put("showDescription", showDescription);
+      config.put("showVolume", showVolume);
+      config.put("showLibraryAlias", showLibraryAlias);
+    }
+  };
+
+  /**
+   * used to edit samples with ids from given {sampleIds} sends Dtos objects which will then be used for editing in grid
```
callunity: I'm a bit confused about this documentation -- are these indeed samples being edited? And is this method able to handle both edit and propagate?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -906,30 +785,75 @@ public ModelAndView editBulkLibraries(@PathVariable String libraryIds, ModelMap
...
+    @Override
+    protected DilutionDto packageDto(Library item) {
+      DilutionDto dto = new DilutionDto();
+      dto.setLibrary(Dtos.asDto(item));
+      return dto;
+    }
+
+    @Override
+    protected Iterable<LibraryDilution> load(List<Long> ids) throws IOException {
+      // TODO implement
```
callunity: Stray `TODO`?

miso-web/src/main/webapp/scripts/aa.js
```diff
@@ -0,0 +1 @@
+HotTarget = {}
```
callunity: What was your thinking behind making this its own file?

miso-web/src/main/webapp/scripts/hot.js
```diff
@@ -0,0 +1,573 @@
...
+      }
+    });
+    
+    var save = document.getElementById('save');
+    save
+        .addEventListener(
+            'click',
+            function() {
+              // We are now saving the contents of the table. This can be called
+              // multiple times if the save was
```
callunity: if the save was...unsuccessful?

callunity:
Bulk propagate with OICR config: 
- library alias is not marked as 'required', so saving without filling out out results in the following error: `nested exception is uk.ac.bbsrc.tgac.miso.webapp.controller.rest.RestException: Invalid library alias: 'null' - alias cannot be empty`
- `Index Kit` is not marked as required and `Index 1` is pre-filled out with "No Index"; neither of these have the "invalid" class when the table is loaded. Leaving these as is and then clicking "Save" results in the libraries saving but the `Index 1` field has the "invalid" (red) class added to it, which is visually confusing.

apmasell:
Okay. The validation problem is fixed and you're now forced to pick an index family.

callunity:
The library alias is still not rendering as required during propagation.

Part of the platform change trigger clears out the Index Family field. This is probably why this field is still displaying as empty, and is still not marked as required.

apmasell:
I added more calls to validate cells and that seems to mark things as required.

miso-web/src/main/webapp/scripts/library_hot.js
```diff
@@ -2,834 +2,402 @@
...
-        // if any members are null, fill them with empty objects otherwise things go poorly
+    };
+  };
+  
+  var makeDesignUpdate = function(id, nameProperty, defaultValue, source) {
+    return function(lib, flat, value, setReadOnly, setOptions, setCell) {
+      var design = Utils.array.findFirstOrNull(
+          Utils.array.namePredicate(value), Constants.libraryDesigns);
+      if (design) {
+        setCell(Utils.array.maybeGetProperty(Utils.array.findFirstOrNull(
```
callunity: What does `setCell` do? It's not a Handsontable API method (that would be `setCellMeta`) and I can't see where it's declared.

apmasell: It sets the cell contents: https://github.com/TGAC/miso-lims/pull/881/files#diff-06f0aa9b486ef785656b9d1f0d7afd6fR272

callunity:
It propagates! It updates! So, it's UI time... could you please style the "Edit" and "Propagate dilutions" buttons which appear in the green box? (smaller text, cursor image on hover, and have the link address in the `<a>` tag instead of using an event handler, maybe reword to "Edit these libraries" and "Propagate dilutions")

apmasell:
Something about the styling in that `div` is funky, so I put them else where. I changed to “Edit these libraries”, but since I'd like to use the same bulk actions for the list pages, we need to remember to make them usable in both contexts (so no “selected”).

miso-web/src/main/webapp/scripts/hot_library.js
```diff
@@ -0,0 +1,408 @@
...
+                      'doOnError' : function(json) {
+                        return callback(false);
+                      }
+                    });
+              } else {
+                return callback(false);
+              }
+            },
+            type : 'text',
+            validator : Constants.automaticLibraryAlias
```
djcooke: Should probably be combined with the above validator?

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateLibraryDesignDao.java
```diff
@@ -45,54 +44,27 @@ public void setSessionFactory(SessionFactory sessionFactory) {
...
   @Override
   public List<LibraryDesign> getLibraryDesigns() throws IOException {
-    Query query = currentSession().createQuery("from LibraryDesign");
     @SuppressWarnings("unchecked")
-    List<LibraryDesign> libraryDesigns = query.list();
-    fetchSqlStore(libraryDesigns);
-    return libraryDesigns;
-  }
-
-  private LibraryDesign fetchSqlStore(LibraryDesign libraryDesign) throws IOException {
```
djcooke: :fire:

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/util/BulkTableBackend.java
```diff
@@ -0,0 +1,76 @@
...
+  public ModelAndView propagate(String idString, ModelMap model) throws IOException {
+    List<Long> ids = parseIds(idString);
+    List<Dto> dtos = new ArrayList<>();
+    for (ParentModel item : loadParents(ids)) {
+      dtos.add(packageDto(item));
+    }
+    return prepare(idString, model, true, "Create " + name + " from " + parentName, dtos);
+  }
+
+  protected abstract void writeConfiguration(ObjectMapper mapper, ObjectNode config);
```
apmasell: Yes. It's going to be needed in the sample version where we want to ship around the sample class.

apmasell:
Fixed. I think `BulkTableBackend` is going to have to change. It was meant for 1:1 propagation, but for pools, were also going to need n:1 and I'm not even sure about boxes, so it might be easiest to split the `packageDto` out as something completely different.

PR #884, opened 2017-05-31
Set sample QC passed for new detailed samples

apmasell:
JIRA: GLT-1752


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -266,6 +266,9 @@ private Sample save(Sample sample) throws IOException {
       }
       if (newId == Sample.UNSAVED_ID) {
         newId = sampleDao.addSample(sample);
+        if (sample instanceof DetailedSample && ((DetailedSample) sample).getDetailedQcStatus() != null) {
+          ((DetailedSample) sample).setQcPassed(((DetailedSample) sample).getDetailedQcStatus().getStatus());
+        }
```
djcooke: Should this be moved above (with discarded/volume matching) to affect both inserts and updates?

apmasell: Fixed.

PR #886, opened 2017-05-31
Allow filling a box by user-defined barcodes

apmasell:
JIRA: GLT-788

### DIFF
--- a/miso-web/src/main/webapp/pages/editBox.jsp
+++ b/miso-web/src/main/webapp/pages/editBox.jsp
@@ -213,6 +213,7 @@
         <a onclick="Box.ui.exportBox(${box.id});" href="javascript:void(0);" class="add">Export Box to Excel</a>
         <a onclick="Box.ui.discardEntireBox(${box.id});" href="javascript:void(0);" class="add">Discard All Tubes</a>
         <c:if test="${(scannerEnabled) && (box.size.scannable)}"><a onclick="Box.initScan();" href="javascript:void(0);">Scan Box</a></c:if>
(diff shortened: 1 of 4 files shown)


apmasell:
Fixed the NPE. This is based on how Danielle and Illinca already do things and all of the `GECCO_PLATE` boxes are set up this way and the OCTANE boxes are similar. I think we can do better, but this is good enough for now.

I'm not sure how to design the dialog. If you have better text or layout, I'm happy to change it.

djcooke:
Clearer wording:
"Fill by Special Barcodes" -> "Fill by Barcode Pattern"
Dialog title: "Fill by Barcode Pattern"
Additional text at top of dialog: "Existing tubes with barcodes matching this pattern will be added to the box in (whatever order this happens in)."
Bonus points: picture with zig-zagging arrow to show fill direction

djcooke:
I'm still having trouble getting this to work. I assigned these barcodes to libraries:

TESTBAR001
TESTBAR002
TESTBAR003

I entered 'TESTBAR' as the prefix, selected the 'numeric' suffix, and hit 'Fill.' The previous box contents were removed, but nothing was added. I tried again after the box was empty and still nothing was added.

apmasell:
Fixed the numeric problem. I was reading a UI element I had deleted.

djcooke:
Thoughts on wording?

apmasell:
I thought I made those changes. That wording is fine.

djcooke:
Oh, I missed that. My suggestion of "Existing tubes with barcodes matching this pattern will be added to the box in (whatever order this happens in)" was meant to have the parenthesized part replaced to explain what ordering is used when the pattern is numeric

PR #888, opened 2017-05-31
GLT-1517 DTOs for notification server messages.

tonydebat:


notification-server/src/main/java/uk/ac/bbsrc/tgac/miso/notification/dto/NotificationDto.java
```diff
@@ -0,0 +1,169 @@
...
+    this.sequencerName = sequencerName;
+  }
+
+  public DateTime getCompletionDate() {
+    return completionDate;
+  }
+
+  public void setCompletionDate(DateTime completionDate) {
+    if (completionDate != null) {
+      this.completionDate = new DateTime(completionDate, DateTimeZone.UTC);
```
callunity: `startDate` and `completionDate` are dates in the Run table, and we don't store the associated time. Will this DateTime conversion make assumptions about the time (like 00:00:00.00) that could cause a date change during conversion to UTC?

djcooke: Using `java.util.Date` would work better since that's what they would be converted to for use in MISO Runs anyway. The less conversions we need to do the better.

Hard-coding the timezone to UTC might help when the DB is running on UTC, but it could cause inconsistencies with a DB not running on UTC.

tonydebat: Replaced JodaTime DateTime with LocalDate (new in Java8). LocalDate is a date that has no time or timezone so it avoids UTC conversion.

tonydebat:
Moved to dto module.

djcooke:
I'd argue that these DTOs belong specifically to notification-server. Notification-consumer-services depends on notification-server, and notification-consumer-service should be the only module that consumes the DTOs.

The notification server is also meant to serve clients other than MISO. I'm not sure if EI has uses for this, but either way, I don't see the point to adding another dependency?

djcooke:
I understand wanting to use a better date class than java.util.Date, but the rest of MISO still uses java.util.Date. I don't think you can convert from from LocalDate to java.util.Date without specifying a time zone, so again the conversion can cause an offset. Using the same date class throughout would make things simpler

apmasell:
Since the `notification-server` module is going away, it has to go elsewhere. The new run server is going to depend on `core` for `PlatformType`.

djcooke:
Sorry, closed PR by accident.

Is the new Run Server/Scanner also going to serve other clients, or only MISO? Those other clients would depend on MISO too just for
* DTOs that were intended only to be created by the NotificationServer/RunServer
* A single enum containing maybe 3 strings we care about?

Having the DTOs belong to MISO is comparable to having MISO or GSLE own the Pinery DTOs. Am I missing something?

apmasell:
For now, it's going to be a part of MISO.

djcooke:
Discussed with @apmasell. Since the run-scanner is going to be a new module developed while notification-server still exists, having the DTOs in the separate DTO module for now does make more sense.

The date type is still concerning. Three options:

1. Avoid timezone issues entirely (Works for Dates only, not time): use java-util.Date, which is already handled fine in MISO. No extra conversion necessary
2. Notification Server provides timezone info: NS would use date/time types that include timezone info. MISO is responsible for converting that to it's "UI timezone," and then Hibernate/JDBC/MySQL will convert it to DB time on save
3. MISO is responsible for knowing NS timezone: NS would use date/time types that do not include timezone info. miso.properties would have a setting for notification-server-timezone. MISO is responsible for converting from "notification timezone" to "UI timezone," and then Hibernate/JDBC/MySQL will convert it to DB time on save

I'd vote 1 for dates to keep it simple, and 2 if time is involved. LocalDate would support option 3, which is fine if that's how we want to handle it?

tonydebat:
Regarding the location of the code: The DTOs are likely the only code that will be shared between the RunScanner and MISO. If we're using Maven modules to minimize dependencies we could make this explicit by create a new 'runscanner-dto' module that will be shared between MISO and RunScanner.

Regarding the dates: In the Run table both the `startDate` and the `completionDate` are stored as MySql `date` types, which are dates without time.  The use of LocalDate in the RunScanner dtos accurately reflects this choice. 

A time zone is needed when converting from LocalDate to Date because Date contains a time and therefore needs a time zone. This isn't a failing of LocalDate and the timezone requirement is true everywhere we use Date.

Related Information:
On the RunScanner side the `startDate` for Illumina is also a date without time. The Illumina `completeDate` does contain a time. I'm not sure what the situation is for other sequencers.

miso-dto/src/test/java/uk/ac/bbsrc/tgac/miso/dto/IlluminaNotificationDtoTest.java
```diff
@@ -0,0 +1,107 @@
...
+    mapper.registerModule(new JavaTimeModule())
+        .setDateFormat(new ISO8601DateFormat())
+        .enable(SerializationFeature.INDENT_OUTPUT);
+    String serialized = mapper.writeValueAsString(notificationDto);
+    NotificationDto deSerialized = mapper.readValue(serialized, NotificationDto.class);
+    
+    ZonedDateTime runScannerTorontoStartDate = notificationDto.getStartDate().withZoneSameInstant(ZoneId.of("America/Toronto"));
+    ZonedDateTime misoWinnipegStartDate = deSerialized.getStartDate().withZoneSameInstant(ZoneId.of("America/Winnipeg"));
+
+    assertThat(runScannerTorontoStartDate.toLocalDate(), is(misoWinnipegStartDate.toLocalDate().plusDays(1)));
```
djcooke: This should be considered a failure. We don't want the date to change.

MISO only saves a date (no time), so the notification server/run scanner is only responsible for finding the date (no time). Currently, the `IlluminaTransformer` parses the startDate from the run directory. It will always assume midnight, so we'll end up with a run named '170607...' having a startDate of 170606 whether the run started at 00:00 or 23:59.

To test whether this is going to work in MISO requires converting to a Run, where the startDate will be a java.util.Date, and ensuring that it is the same date as specified by whatever class is used in the NotificationDto.

PR #892, opened 2017-06-02
Allow adjusting run dates if admin or new run

apmasell:
JIRA: GLT-1793

### DIFF
--- a/miso-web/src/main/webapp/pages/editRun.jsp
+++ b/miso-web/src/main/webapp/pages/editRun.jsp
@@ -175,8 +175,8 @@
     <td>Run Path:*</td>
     <td>
       <c:choose>
-        <c:when test="${not empty run.filePath}">${run.filePath}</c:when>
-        <c:otherwise><form:input path="filePath"/></c:otherwise>
+        <c:when test="${empty run.filePath or fn:contains(SPRING_SECURITY_CONTEXT.authentication.principal.authorities,'ROLE_ADMIN')}"><form:input path="filePath"/></c:when>
+        <c:otherwise>${run.filePath}</c:otherwise>
       </c:choose>
     </td>
   </tr>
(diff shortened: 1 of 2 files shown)


miso-web/src/main/webapp/pages/editRun.jsp
```diff
@@ -256,9 +256,22 @@
...
+              </script>
+            </c:when>
+            <c:otherwise>
+              <td id="startDate">
+                <fmt:formatDate pattern="dd/MM/yyyy" value="${run.startDate}"/>
+              </td>
+            </c:otherwise>
+          </c:choose>
+          <c:choose>
+          <c:when test="${(run.health.isDone() and empty run.completionDate) or run.id == 0 or fn:contains(SPRING_SECURITY_CONTEXT.authentication.principal.authorities,'ROLE_ADMIN')}">
```
callunity: I don't know that completion date should be editable (even by admins) before the run is actually complete. What do you think?

djcooke: True... Also as is, if you wanted to mark something completed and add the completionDate at the same time (as would be normal), you would have to
1. change to Complete
2. save
3. set completionDate
4. save again

callunity: Hmm, when I deployed to test, the completion date entry box was visible all the time so I don't know that there would have been a requirement to save twice. I think that before the big recent schema changes, there was an event handler to make the completion date box visible when the status is set to Failed or Completed. IMO, that would be the clearest way to implement it.

apmasell: So `isDone && (empty || admin)`? That would still require two saves.

djcooke: Make input if admin and use JS to control whether it's editable?

apmasell: Turns out that logic was there all along...just broken.

callunity:
Could you please make `startDate` a required field (in the `validateRun` function) if that input is present?

apmasell:
Done. And I check completion date too.

apmasell:
Fixed.

apmasell:
Yeah. I'm not a fan, but It Was Like That When I Found It™.

PR #893, opened 2017-06-02
Fix manual container creation & update service record wording

callunity:


miso-web/src/main/webapp/scripts/sequencer_partition_container_ajax.js
```diff
@@ -282,7 +282,7 @@ Container.ui = {
       {
         'updateElement': 'containerdiv',
         'doOnSuccess': function(json) {
-            jQuery('#containerdiv').find('input[type="radio"]').first().click();
+          jQuery('#containerdiv').find('input.selectMe').click();
```
apmasell: You should be able to put `checked="checked"` on the radio button you want and not bother with this jQuery.

callunity: The partitions are fetched by a separate AJAX call which is triggered by the click, and the string concatenation and session attribute setting in the ContainerControllerHelperService is complex enough that refactoring it would not be straightforward (for me). I'd prefer to fix the behaviour now and refactor the functions if/when we work out the ControllerHelperServices.

apmasell: Sure. Right...this is a mess I had forgotten about.

callunity:
@djcooke I agree that this would be nicer. I've made a ticket for it: https://jira.oicr.on.ca/browse/GLT-1802

PR #894, opened 2017-06-02
populate RunService and SampleService

callunity:


sqlstore/src/main/resources/db/migration_afterMigrate/05_standard.sql
```diff
@@ -886,12 +909,12 @@ DELIMITER ;
 
 DROP VIEW IF EXISTS CompletedPartitions;
 CREATE OR REPLACE VIEW RunPartitionsByHealth AS
-  SELECT pool_poolId AS poolId, sequencingParameters_parametersId as parametersId, COUNT(*) AS num_partitions, health AS health, MAX((SELECT MAX(changeTime) FROM RunChangeLog WHERE RunChangeLog.runId = Run.runId)) as lastUpdated
```
apmasell: This reverts my fix for the order completions.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/RunService.java
```diff
@@ -1,8 +1,87 @@
...
 
+  Collection<Run> list() throws IOException;
+
+  Collection<Run> listWithLimit(long limit) throws IOException;
+
+  int count() throws IOException;
+
+  Map<String, Integer> getRunColumnSizes() throws IOException;
+
+  Collection<Run> listBySearch(String query) throws IOException;
```
apmasell: Do we need these (and the sample versions)? Can we use the pagination interface to do that? We might need sequencer and container, but I think the rest are covered.

sqlstore/src/main/resources/db/migration_afterMigrate/05_standard.sql
callunity: Weird. I'll fix that.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/RunService.java
callunity: I'll make some updates to that in another PR.

PR #895, opened 2017-06-05
GLT-1785: fixed saving of runs during migration

djcooke:


djcooke:
`CascadeType.ALL` is causing different issues (though not with migration). Closing for now

PR #896, opened 2017-06-05
Refactor Run

callunity:
- moves the `pairedEnd` field out of `Run` and into `IlluminaRun`, `LS454Run`, `SolidRun`
- removes `creationDate` and `wellName` from `PacBioRun`

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Run.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Run.java
@@ -55,9 +55,6 @@ import javax.persistence.Temporal;
 import javax.persistence.TemporalType;
 import javax.persistence.Transient;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import com.eaglegenomics.simlims.core.Group;
 import com.eaglegenomics.simlims.core.Note;
 import com.eaglegenomics.simlims.core.SecurityProfile;
@@ -93,8 +90,6 @@ public class Run
     implements SecurableByProfile, Comparable<Run>, Reportable, Watchable, Deletable, Nameable, Alertable, ChangeLoggable, Aliasable,
     Serializable {
   private static final long serialVersionUID = 1L;
-
-  private static final Logger log = LoggerFactory.getLogger(Run.class);
 
   /** Field PREFIX */
   public static final String PREFIX = "RUN";
@@ -553,7 +548,7 @@ public class Run
   }
 
   public void setPairedEnd(boolean pairedEnd) {
-    // do nothing;
+    throw new UnsupportedOperationException("Cannot set paired end on runs from this platform.");
   }
 
 }
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Run.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Run.java
@@ -548,4 +548,12 @@ public class Run
     return changeLogs;
   }
 
+  public Boolean getPairedEnd() {
+    return null;
+  }
+
+  public void setPairedEnd(boolean pairedEnd) {
+    // do nothing;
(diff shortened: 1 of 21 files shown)


callunity:
The PacBio dashboard URL uses the startDate, not the creationDate.

That would be much simpler than all the casting, wouldn't it. I'll do that.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Run.java
```diff
@@ -559,4 +548,12 @@ public boolean userCanWrite(User user) {
     return changeLogs;
   }
 
+  public Boolean getPairedEnd() {
+    return null;
+  }
+
+  public void setPairedEnd(boolean pairedEnd) {
+    // do nothing;
```
djcooke: Getter is fine for convenience, but either remove the setter or throw `UnsupportedOperationException` instead of ignoring misuse that can cause confusion

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/SolidRun.java
```diff
@@ -0,0 +1,42 @@
...
+import com.eaglegenomics.simlims.core.User;
+
+import uk.ac.bbsrc.tgac.miso.core.data.IlluminaRun;
+import uk.ac.bbsrc.tgac.miso.core.data.Run;
+
+@Entity
+@Table(name = "RunSolid")
+public class SolidRun extends Run {
+  private static final long serialVersionUID = 1L;
+  private static final Logger log = LoggerFactory.getLogger(IlluminaRun.class);
```
djcooke: Wrong class, but also not used. :fire:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
```diff
@@ -791,16 +791,20 @@ public long saveRun(Run run) throws IOException {
...
           PacBioRun runPacBio = (PacBioRun) run;
-          managedPacBio.setCreationDate(runPacBio.getCreationDate());
           managedPacBio.setMovieDuration(runPacBio.getMovieDuration());
-          managedPacBio.setWellName(runPacBio.getWellName());
         } else if (managed instanceof LS454Run) {
           LS454Run managedLS454 = (LS454Run) managed;
           LS454Run runLS454 = (LS454Run) run;
           managedLS454.setCycles(runLS454.getCycles());
+          managedLS454.setPairedEnd(runLS454.getPairedEnd());
+        } else if (managed instanceof SolidRun) {
```
djcooke: Use the `LimsUtils` method you added (could fix those above as well)

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -380,25 +381,28 @@ private void applyChanges(Run target, Run source) throws IOException {
     if (isIlluminaRun(target)) applyIlluminaChanges((IlluminaRun) target, (IlluminaRun) source);
     if (isPacBioRun(target)) applyPacBioChanges((PacBioRun) target, (PacBioRun) source);
     if (isLS454Run(target)) applyLS454Changes((LS454Run) target, (LS454Run) source);
+    if (isSolidRun(target)) applySolidChanges((SolidRun) target, (SolidRun) source);
```
djcooke: This and above can be 'else ifs'

Remove `target.setPairedEnd` above since it should only be set on specific subclasses

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Run.java
callunity: Fixed this and comments below.

PR #900, opened 2017-06-06
GLT-1803: fix properties used in header.jsp

djcooke:
Bug URL was provided by ApplicationContextProvider, which still worked on my machine, but apparently stopped working on staging and prod.

Instance name was in the ROOT.xml instead of properties file.

Both are now set in miso.properties and injected via CommonModelAttributeProvider into models returned from all Controllers.

### DIFF
--- a/miso-web/src/main/webapp/header.jsp
+++ b/miso-web/src/main/webapp/header.jsp
@@ -104,7 +104,7 @@
       <a href='<c:url value='/'/>'>
         <img src="<c:url value='/styles/images/miso_bowl1_logo-tm.png'/>" alt="MISO Logo" name="logo" border="0" id="misologo"/>
       </a>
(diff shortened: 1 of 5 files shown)


miso-web/src/main/webapp/header.jsp
```diff
@@ -104,7 +104,7 @@
       <a href='<c:url value='/'/>'>
         <img src="<c:url value='/styles/images/miso_bowl1_logo-tm.png'/>" alt="MISO Logo" name="logo" border="0" id="misologo"/>
       </a>
-       <span id="instanceName">${initParam['miso.name']}</span>
+       <span id="instanceName">${misoInstanceName}</span>
```
callunity: If it's null will it just display nothing?

djcooke: Correct. It will be blank, and will not show "null"

PR #901, opened 2017-06-06
GLT-1785: fixed saving of runs during migration

djcooke:


migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -630,7 +630,7 @@ public void saveRuns(final Collection<Run> runs) throws IOException {
       }
       for (SequencerPartitionContainer container : run.getSequencerPartitionContainers()) {
         container.setLastModifier(migrationUser);
-        container.setId(serviceManager.getSequencerPartitionContainerDao().save(container));
+        container.setId(serviceManager.getRequestManager().saveSequencerPartitionContainer(container));
```
callunity: :crying_cat_face:

PR #905, opened 2017-06-08
Base of a run scanner

apmasell:
JIRA: GLT-1793


runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/Main.java
```diff
@@ -0,0 +1,71 @@
...
+
+    PlatformType pt = PlatformType.valueOf(platformName);
+    String name = System.getProperty("name", "default");
+    for (RunProcessor rp : RunProcessor.INSTANCES) {
+      if (rp.getPlatformType() == pt && rp.getName().equals(name)) {
+        List<NotificationDto> results = new ArrayList<>();
+        boolean success = true;
+        for (String path : args) {
+          File directory = new File(path);
+          if (!directory.isDirectory() || !directory.canExecute() || !directory.canExecute()) {
```
callunity: duplication of `!directory.canExecute()`?

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
```diff
@@ -0,0 +1,71 @@
...
+import io.prometheus.client.Histogram;
+import uk.ac.bbsrc.tgac.miso.core.data.type.PlatformType;
+import uk.ac.bbsrc.tgac.miso.core.util.LimsUtils;
+import uk.ac.bbsrc.tgac.miso.dto.NotificationDto;
+
+public abstract class RunProcessor {
+  private static final Counter errors = Counter.build().name("miso_runscanner_errors").help("The number of bad directories encountered.")
+      .labelNames("platform").register();
+
+  public static final Iterable<RunProcessor> INSTANCES = Arrays.asList();
```
djcooke: How do you intend to add RunProcessors to this?

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/Scheduler.java
```diff
@@ -0,0 +1,222 @@
...
+        for (RunProcessor processor : RunProcessor.INSTANCES) {
+          if (processor.getName().equals(source.getName()) && processor.getPlatformType() == source.getPlatformType()) {
+            destination.setProcessor(processor);
+            break;
+          }
+        }
+        return destination;
+      }).collect(Collectors.toConcurrentMap(Configuration::getPath, Function.identity()));
+    } catch (IOException e) {
+      log.error("Confguration is bad.", e);
```
djcooke: Bad configuration should probably cause the servlet to fail startup

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/Scheduler.java
```diff
@@ -0,0 +1,222 @@
...
+              if (result == null) {
+                failed.add(directory);
+              } else {
+                finishedWork.put(directory, result);
+              }
+              processing.remove(directory);
+            });
+          });
+        } catch (Exception e) {
+          log.error("Error scanning directory.", e);
```
djcooke: Add to failed set?

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
```diff
@@ -0,0 +1,71 @@
...
+      .labelNames("platform", "instrument").register();
+
+  private final String name;
+
+  private final PlatformType platformType;
+
+  public RunProcessor(PlatformType platformType, String name) {
+    super();
+    this.platformType = platformType;
+    this.name = name;
```
djcooke: Is the name so that you can have multiple Processors per platform? Like one for Illumina HiSeq and one for Illumina NextSeq?

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/Main.java
apmasell: Sorry. Should be `canRead`.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
apmasell: We've going to have to add them as we build them. I'll add a TODO. I looked briefly at Spring annotation magic, but I'm not sure it's worth the effort.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
apmasell: It's probably going to be GenomeAnalyzer II versus every other instrument in the case of Illumina. For the PacBio, it's probably going to reflect the version of the webservice rather than the version of the instrument itself.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  public final Supplier<NotificationDto> apply(File runDirectory, TimeZone tz) {
```
apmasell: Because of the threading model. The main thread is going to run through the directories and queue the suppliers for later execution by the thread pool.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
```diff
@@ -0,0 +1,71 @@
...
+      long startTime = System.nanoTime();
+      String instrumentName = "unknown";
+      NotificationDto dto = null;
+      try {
+        dto = process(runDirectory, tz);
+        if (!LimsUtils.isStringBlankOrNull(dto.getSequencerName())) {
+          instrumentName = dto.getSequencerName();
+        }
+      } catch (Exception e) {
+        log.error("Failed to process run: " + runDirectory.getPath(), e);
```
apmasell: The consumer is the thread pool, so it doesn't really care. I'm happy to push this logic into the scheduler, so it seemed desirable to remove as much code as possible from that logic.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
```diff
@@ -0,0 +1,71 @@
...
+  public String getName() {
+    return name;
+  }
+
+  public PlatformType getPlatformType() {
+    return platformType;
+
+  }
+
+  protected abstract NotificationDto process(File runDirectory, TimeZone tz) throws Exception;
```
apmasell: If it's going to be rummaging around in XML files, those can produce a bunch of errors. Ultimately, whatever error the processor has, it can't deal with. It makes sense to just log any errors, skip the run, and move on.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/Scheduler.java
apmasell: I've changed this logic. I think the servlet should keep running even if the configuration is bad. It should report bad configuration and scan for a changed configuration. If I correct the configuration on disk, having the servlet fail on start up only adds steps for me having to fix it.

Also, consider the configuration where the configuration lives in git. The servlet just picks up a new configuration as it gets pulled from git. If junk is checked into git, the servlet either continues to use the old config, or does nothing until a new one comes a long. It can report the failure to the monitoring system.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/Scheduler.java
apmasell: This is at the instruments level, not the run level. If an instrument directory is bad, we should ignore and continue and let monitoring pick it up. We'll need to have an alert that if `attempted > accepted`, some instrument is dead.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/Pair.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  @Override
+  public V getValue() {
+    return value;
+  }
+
+  @Override
+  public V setValue(V arg0) {
+    throw new IllegalStateException("Pairs are immutable.");
```
djcooke:
```
Pair<File, Run> pair = new Pair(file, run);
run.setName("mutated name"); // pair has been mutated
```
I'm not saying it matters. Just that the message is wrong. You'd have to use defensive copying or something, on unknown types, for Pair to actually be immutable

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
djcooke: How about a list field in the Scheduler that's populated in the application context xml? It means the single-directory scanner would also need to be aware of all the processors, but that's cleaner than a mutable public static field in a superclass keeping track of subclass instances

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
djcooke: This does nothing except return the Supplier though. If you had a `get` method on the RunProcessor instead, you could just call that from the other thread. I guess it's a way of storing the parameters?

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
djcooke: I think of it this way:
* the processor's job is to scan the run folder and report what it finds. This could be a proper run, or an error
* the scheduler's job is to manage those results

A clear separation of responsibilities would mean that the processor shouldn't decide what gets done with an exception that it can't properly handle itself

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/Scheduler.java
djcooke: :+1: to hot-swappable config.

I think maybe I'm missing something though. It seems like a config file will only define one directory. Do we have to run a separate instance of the server for each instrument, or just per platform?

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
apmasell: Right. It's closing the parameters in the lambda. Think of it as a future: https://en.wikipedia.org/wiki/Futures_and_promises

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
apmasell: Okay. I've changed it so the `RunProcessor` just has the abstract `process` method and that logic has moved to the `Scheduler`.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/Main.java
```diff
@@ -0,0 +1,70 @@
...
+          File directory = new File(path);
+          if (!directory.isDirectory() || !directory.canExecute() || !directory.canRead()) {
+            System.err.println("Target is not a useable directory: " + path);
+            success = false;
+            continue;
+          }
+          try {
+            results.add(rp.process(directory, tz));
+          } catch (Exception e) {
+            System.err.println("Cannot process directory: " + path);
```
djcooke: If you were running this to troubleshoot a run scan, the details of the exception would probably be the most useful thing to see

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
djcooke: The point of a future is to return from an already-called function before the value is resolved. This is the opposite. You're delaying the function call instead of the result. It works though

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
```diff
@@ -0,0 +1,36 @@
...
+  public String getName() {
+    return name;
+  }
+
+  public PlatformType getPlatformType() {
+    return platformType;
+
+  }
+
+  public abstract NotificationDto process(File runDirectory, TimeZone tz) throws Exception;
```
djcooke: Make it throw `IOException` for now and add others later if they make sense? The caller can still lazily catch `Exception` if desired

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/Scheduler.java
```diff
@@ -148,11 +166,12 @@ public boolean isScanningNow() {
...
 
-  @Value("${runscanner.configFile}")
-  public void setConfigurationFile(String filename) {
+  private void readConfiguration() {
     ObjectMapper mapper = new ObjectMapper();
+    configurationLastRead = Instant.now();
+    configurationTimestamp.set(configurationLastRead.getEpochSecond());
     try {
-      roots = Arrays.stream(mapper.readValue(new File(filename), SuppliedDirectoryConfig[].class)).map(source -> {
+      roots = Arrays.stream(mapper.readValue(configurationFile, SuppliedDirectoryConfig[].class)).map(source -> {
```
djcooke: Please include an example config file

apmasell:
Fixed.

PR #906, opened 2017-06-09
Refactor BulkTableBackend classes for different use cases

callunity:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -788,76 +819,38 @@ protected LibraryDto packageDtoFromParent(Sample sample) {
...
-    protected Class<? extends LibraryDto> getDtoClass() {
-      return isDetailedSampleEnabled() ? DetailedLibraryDto.class : LibraryDto.class;
-    }
-
-    @Override
     protected void writeConfiguration(ObjectMapper mapper, ObjectNode config) {
-      config.put("showDescription", showDescription);
-      config.put("showVolume", showVolume);
-      config.put("showLibraryAlias", showLibraryAlias);
+      writeLibraryConfiguration(mapper, config);
```
djcooke: Why another method?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/util/BulkCreateTableBackend.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+
+import org.springframework.ui.ModelMap;
+import org.springframework.web.servlet.ModelAndView;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * Create a Handsontable for propagating or editing a particular entity type
```
djcooke: There are 3 classes with the same description saying "propagating or editing," but one has "Propagate" in its name, and one has "Edit"

callunity: Bad copypasta. I'll fix that.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
callunity: Because it applies to both BulkPropagateTableBackend and BulkEditTableBackend.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -733,11 +734,41 @@ public ModelAndView setupForm(@PathVariable Long libraryId, @PathVariable Long s
...
+    protected Iterable<Library> load(List<Long> ids) throws IOException {
+      List<Library> results = libraryService.listByIdList(ids);
+      SampleClass sampleClass = null;
+      for (Library library : results) {
+        if (!isDetailedSampleEnabled()) {
+          // Do nothing about sample classes.
+        } else if (sampleClass == null) {
+          sampleClass = ((DetailedSample) library.getSample()).getSampleClass();
+        } else if (((DetailedSample) library.getSample()).getSampleClass().getId() != sampleClass.getId()) {
+          throw new IOException("Can only update libraries when samples all have the same class.");
```
callunity: DNA and RNA have different LibraryDesigns available, and adding a LibraryDesign that's incompatible with the sample aliquot's class throws an error on save. I don't think we've had any complaints from our users about this, have we?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
djcooke: Oh, sorry. Missed a curly brace and thought `writeLibraryConfiguration` was also inside of `libraryBulkPropagateBackend`. :+1:

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
djcooke: Not that I'm aware of. It could cause frustration since the sample class isn't visible on the List Libraries page, and the SampleClass isn't what actually matters. Deal with it later if it becomes a problem I guess. :+1:

PR #909, opened 2017-06-13
Glt 1723: Generate Library Alias

djcooke:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditLibraryController.java
```diff
@@ -750,16 +754,6 @@ protected LibraryDto packageDtoFromParent(Sample sample) {
...
-      if (namingScheme.hasLibraryAliasGenerator()) {
-        try {
-        Library tempLibrary = new LibraryImpl();
-        tempLibrary.setSample(sample);
-        final String libraryAlias = namingScheme.generateLibraryAlias(tempLibrary);
-        dto.setAlias(libraryAlias);
-        } catch (MisoNamingException e) {
-          log.error("Failed to generate library name", e);
-        }
-      }
```
callunity: EI wants to see the library alias already pre-generated when they start bulk creating new libraries. Did this functionality get moved somewhere else?

djcooke: @frankie13 had set it up this way because that's how it was done before, but said it wasn't a requirement. Generating after save has the advantage of preventing duplicates if multiple people are creating libraries at the same time. When we have one to many propagation, generating before save gets even more problematic, and wouldn't have worked as it was.

djcooke: I do need to update the DefaultLibraryAliasGenerator for the change though, so will do that

miso-web/src/main/webapp/scripts/hot.js
```diff
@@ -459,6 +458,16 @@ var HotUtils = {
     });
   },
   
+  toFlatObj : function(item, columns) {
```
apmasell: Can you make this function locally-scoped? If it's not safe for external entities to call it, it shouldn't be visible.

djcooke:
Addressed all comments and rebased to fix merge conflict

callunity:
On Create Single Library page (single sample page > Add Library), it says "Alias: (blank to auto-generate)". The "(blank to auto-generate)" should be removed if `nonStandardAlias` is set on the parent sample, because of the logic set [here](https://github.com/TGAC/miso-lims/pull/909/files#diff-fb5d988190af198ad06a06e95179ef97R63).

djcooke:
@callunity: Fixed now. Thanks.

It may actually be possible to generate a sample/library alias even when the parent's alias is non-standard, and this is allowed in bulk mode. On the single pages, there's no good way of handling the error when it fails though.

PR #911, opened 2017-06-13
GLT-1517 Dates are no longer Strings.

tonydebat:


miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -1589,4 +1593,40 @@ public static StainDto asDto(Stain from) {
...
+  private static Run setIlluminaRunValues(IlluminaNotificationDto from, IlluminaRun to) {
+    to.setPairedEnd(from.isPairedEndRun());
+    return to;
+  }
+
+  private static Run setCommonRunValues(NotificationDto from, Run to) {
+    to.setName(from.getRunName());
+    to.setFilePath(from.getSequencerFolderPath().toString());
+    to.setHealth(from.getHealthType());
+    to.setStartDate(Date.from(from.getStartDate().atStartOfDay(ZoneId.systemDefault()).toInstant()));
```
djcooke: Would be good to add as a LimsUtils method if mixing date types is going to become more common

PR #917, opened 2017-06-14
Create a stored procedure for adding identities

apmasell:


sqlstore/src/main/resources/db/migration_beforeMigrate/05_standard.sql
```diff
@@ -419,6 +419,32 @@ CREATE PROCEDURE addTissueType(
...
 END//
 
+--  Use inconjunction with the spreadsheet formula: =CONCATENATE("CALL addIdentity('",A2,"', '",B2,"', '",C2,"', ", IF(ISBLANK(D2), "NULL", CONCATENATE("'", D2, "'")), ", ", IF(ISBLANK(E2), "NULL", CONCATENATE("DATE '", E2, "'")), ", '", F2, "', '", G2, "');")
+DROP PROCEDURE IF EXISTS addIdentity//
+CREATE PROCEDURE addIdentity(
+  iProject varchar(255),
+  iExternalName varchar(255),
+  iAlias varchar(255),
+  iDescription varchar(255),
+  iDate date,
```
callunity: We don't actually receive donors, so iDate is not necessary.

sqlstore/src/main/resources/db/migration_beforeMigrate/05_standard.sql
```diff
@@ -419,6 +419,32 @@ CREATE PROCEDURE addTissueType(
...
+  iScientificName varchar(255),
+  iSampleType varchar(255)
+) BEGIN
+  DECLARE newSampleId bigint(20);
+  
+  IF NOT EXISTS (SELECT 1 FROM Identity WHERE externalName = iExternalName) THEN
```
callunity: Needs to be project-constrained as well. There's no reason we can't have the same external name in two projects. 
This also needs to confirm that there is no pre-existing identity with the custom alias.

sqlstore/src/main/resources/db/migration_beforeMigrate/05_standard.sql
apmasell: It's in the spread sheet, oddly.

sqlstore/src/main/resources/db/migration_beforeMigrate/05_standard.sql
callunity: We should probably also have an "else" case that throws an error to indicate that an identity with that external name already exists in the project. Right now it returns `Query OK, 0 rows affected`, which does not clearly indicate that no record was created, or why.

sqlstore/src/main/resources/db/migration_beforeMigrate/05_standard.sql
```diff
@@ -419,6 +419,35 @@ CREATE PROCEDURE addTissueType(
   END IF;
 END//
 
+--  Use inconjunction with the spreadsheet formula: =CONCATENATE("CALL addIdentity('",A2,"', '",B2,"', '",C2,"', ", IF(ISBLANK(D2), "NULL", CONCATENATE("'", D2, "'")), ", ", IF(ISBLANK(E2), "NULL", CONCATENATE("DATE '", E2, "'")), ", '", F2, "', '", G2, "');")
```
callunity: `in conjunction`
:trophy: for the formula!

sqlstore/src/main/resources/db/migration_beforeMigrate/05_standard.sql
apmasell: Both fixed.

sqlstore/src/main/resources/db/migration_beforeMigrate/05_standard.sql
```diff
@@ -430,21 +430,25 @@ CREATE PROCEDURE addIdentity(
   iSampleType varchar(255)
 ) BEGIN
   DECLARE newSampleId bigint(20);
+  DECLARE errorMessage varchar(300);
   
-  IF NOT EXISTS (SELECT 1 FROM Identity WHERE externalName = iExternalName) THEN
-    INSERT INTO Sample(name, alias, description, sampleType, project_projectId, scientificName, lastModifier, securityProfile_profileId) VALUES 
-        ('TEMP_IDENT', iAlias, iDescription, iSampleType, (SELECT projectId FROM Project WHERE shortName = iProject), iScientificName, getAdminUserId(), (SELECT securityProfile_profileId FROM Project WHERE shortName = iProject));
-    SET newSampleId = LAST_INSERT_ID();
+  IF EXISTS (SELECT 1 FROM Identity JOIN Sample USING (sampleId) JOIN Project ON Project.projectId = Sample.project_projectId WHERE externalName = iExternalName AND Project.alias = iAlias) THEN
```
callunity: `and Sample.alias = iAlias` not Project.alias

callunity: This checks the case where an identical identity has already been entered, but doesn't fix the sub-cases where a) the lab tells us to enter `DYS_8001` on two separate occasions, or b) the lab tells us to enter an identity with external name `EXT1` for both `DYS_8001` and `DYS_5001`.

apmasell: Fixed. I thought aliases were unique.

sqlstore/src/main/resources/db/migration_beforeMigrate/05_standard.sql
```diff
@@ -419,6 +419,38 @@ CREATE PROCEDURE addTissueType(
...
+  iExternalName varchar(255),
+  iAlias varchar(255),
+  iDescription varchar(255),
+  iScientificName varchar(255),
+  iSampleType varchar(255)
+) BEGIN
+  DECLARE newSampleId bigint(20);
+  DECLARE errorMessage varchar(300);
+  
+  IF EXISTS (SELECT 1 FROM Identity JOIN Sample USING (sampleId) JOIN Project ON Project.projectId = Sample.project_projectId WHERE externalName = iExternalName AND Project.alias = iProject OR Sample.alias = iAlias) THEN
```
callunity: `Project.alias` or `Project.shortName`? You use `Project.shortName` below, and we have 15 projects where alias != shortName. 

Might be worth renaming `iProject` to `iProject[Alias|ShortName]` to make that clearer for when we come back to this function in several weeks...

apmasell: I think the the lab only uses short name, so I'll go with that.

PR #919, opened 2017-06-15
Create list pages with tabs

apmasell:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/util/TabbedListItemsPage.java
```diff
@@ -0,0 +1,103 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import uk.ac.bbsrc.tgac.miso.core.data.Project;
+import uk.ac.bbsrc.tgac.miso.core.data.type.PlatformType;
+import uk.ac.bbsrc.tgac.miso.core.manager.RequestManager;
+
+public class TabbedListItemsPage {
+  private static final Logger log = LoggerFactory.getLogger(TabbedListItemsPage.class);
+
+  public static TabbedListItemsPage createForPlatformType(String targetType, String property, RequestManager requestManager)
```
callunity: If we can get [this reviewed](https://doodle.com/poll/w3kxspsefv9k743z), you'll have a PlatformService and won't need the RequestManager. Or the PlatformService could even be extracted out of that PR.

miso-web/src/main/webapp/pages/adminsub.jsp
```diff
@@ -30,11 +30,6 @@
 				Export</a></li>
 	</ul>
 
-	<h2>Sequencing</h2>
-	<ul class="bullets">
-		<li><a href="<c:url value="/miso/pools/ready"/>">Ready to Run</a></li>
```
callunity: I think the lab will still want quick access to which pools are in line to be sequenced next, but that might be better done by replacing "Ready to Run" with "Outstanding Orders". I've added an agenda item to discuss this.

apmasell: I agree. I thought about this for a while when I first did it, but it came down to ready-to-run being kind of useless as it is. I'll recondier.

apmasell:
Runs are in fact, already stabbed and waiting for this.

apmasell:
Fixed.

PR #923, opened 2017-06-20
Create runs list using new utilities

apmasell:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/RunRestController.java
```diff
@@ -135,19 +136,39 @@ public void setSecurityManager(SecurityManager securityManager) {
...
   }
 
+  @RequestMapping(value = "/dt/platform/{platform}", method = RequestMethod.GET, produces = "application/json")
+  @ResponseBody
+  public DataTablesResponseDto<RunDto> dataTableByPlatform(@PathVariable("platform") String platform, HttpServletRequest request,
+      HttpServletResponse response, UriComponentsBuilder uriBuilder)
+      throws IOException {
+    PlatformType platformType = PlatformType.valueOf(platform);
+    if (platformType == null) {
+      throw new RestException("Invalid platform type.");
```
djcooke: This will default to returning HTTP 500 (Internal Server Error). Include `Status.BAD_REQUEST` as second constructor parameter to return HTTP 400 instead

apmasell: Fixed. I'll do this on the other PRs too.

miso-web/src/main/webapp/scripts/list_run.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+  },
+  createBulkActions : function(config, projectId) {
+    return [];
+  },
+  createStaticActions : function(config, projectId) {
+    return (!projectId && config.platformType) ? [ {
+      name : "Add",
+      handler : function() {
+        window.location = '/miso/run/new/' + config.platformType;
```
callunity: Could you update the button text to include the platformType? It's not immediately obvious that the add button changes depending on the tab.

PR #924, opened 2017-06-20
Create containers list using new utilities

apmasell:


miso-web/src/main/webapp/scripts/list_container.js
```diff
@@ -0,0 +1,80 @@
...
+            return (data ? "<a href=\"/miso/run/" + full.lastRunId + "\">" + data + "</a>" : "");
+          },
+          "bSortable": false,
+          "include" : true,
+          "iSortPriority" : 0
+        },
+        {
+          "sTitle": "Last Sequencer Used",
+          "include" : true,
+          "mData": "id",
```
callunity: This should display the sequencer's name. The old code did this using a `fnRowCallback`

apmasell: Fixed.

PR #925, opened 2017-06-20
Create kits list with new utilities

apmasell:


spring/src/main/java/uk/ac/bbsrc/tgac/miso/spring/ajax/ExperimentControllerHelperService.java
```diff
@@ -247,42 +247,6 @@ public JSONObject addLibraryKit(HttpSession session, JSONObject json) {
     return addKit(session, json, KitType.LIBRARY);
   }
 
-  // empcr
-  public JSONObject getEmPcrKitDescriptors(HttpSession session, JSONObject json) {
```
callunity: :fire:

miso-web/src/main/webapp/scripts/list_kits.js
```diff
@@ -0,0 +1,84 @@
...
+    return "/miso/rest/kitdescriptor/dt" + (config.kitType
+        ? "/type/" + config.kitType : "");
+  },
+  createBulkActions : function(config, projectId) {
+    
+    return [];
+  },
+  createStaticActions : function(config, projectId) {
+    return [ {
+      "name" : "Add",
```
callunity: Similarly, could this button text be updated with the type of kit?

apmasell: It actually brings you to the same kit either way. You have to pick the kit type on the target page. :-(

callunity: Oh ok. All the more reason to be explicit with the text on the Runs one then, since that does bring you to different pages.

PR #927, opened 2017-06-20
Convert order completions to new list utilities

apmasell:


miso-web/src/main/webapp/scripts/list_completions.js
```diff
@@ -0,0 +1,139 @@
...
+          "iSortPriority" : 0,
+          "include" : true
+        
+        }, {
+          "sTitle" : "Completed",
+          "mData" : "completed",
+          "bSortable" : false,
+          "iSortPriority" : 0,
+          "include" : true,
+          "visibilityFilter" : nonZero
```
callunity: :100: to `nonZero`

miso-web/src/main/webapp/scripts/list_completions.js
```diff
@@ -0,0 +1,139 @@
...
+        }, {
+          "sTitle" : "Remaining",
+          "mData" : "remaining",
+          "iSortPriority" : 0,
+          "include" : true,
+          "visibilityFilter" : nonZero
+        }, {
+          "sTitle" : "Last Updated",
+          "mData" : "lastUpdated",
```
callunity: This is pretty verbose (displays down to the millisecond). Would you mind adjusting this in the DTO?

apmasell: It's just date now. Interestingly, this is the date time format we have everywhere (that's a zone not millis), I'm just not sure it ever gets displayed..

callunity: Thanks for making the change!

PR #928, opened 2017-06-21
Remove alerts and reportables

apmasell:


callunity:
Failing tests due to bad migration.

apmasell:
Fixed.

miso-web/src/main/webapp/WEB-INF/event-config.xml
```diff
@@ -42,140 +42,4 @@
          <value>uk\.ac\.bbsrc\.tgac\.miso\.persistence\.impl\.Hibernate.*</value>
```
djcooke: Should move the `prometheusInterceptor` bean definition to `miso-config.xml` and delete this file (remove import from `applicationContext.xml` too

PR #930, opened 2017-06-22
Refactor sample Handsontable

callunity:
Note: the actions from the Samples page haven't yet been changed, because they will be refactored by @apmasell . To test this, use variations on the following addresses:

- create new: `your.local.miso:8080/miso/sample/bulk/new?quantity={quantity}&sampleClassId={receivableSampleClassId}`
- propagate new: `your.local.miso:8080/sample/bulk/propagate?parentIds={comma,sep,parent,ids}&sampleClassId={validSampleClassId}`
- edit remains the same


miso-web/src/main/webapp/scripts/hot_sample.js
```diff
@@ -0,0 +1,485 @@
...
+          header: 'Project',
+          data: 'projectAlias',
+          type: 'dropdown',
+          source: (function () {
+              var comparator = Constants.isDetailedSample ? 'shortName' : 'id';
+              var label = Constants.isDetailedSample ? 'shortName' : 'name';
+              var projectLabels = (config.projects ? config.projects.sort(Utils.array.standardSort(comparator))
+              .map(function(item) {
+                return item[label];
+              }) : []); // use empty array if projects are not provided
```
apmasell: If the projects failed to load, this is a problem. Maybe `requestConfiguration` can check and show a dialog if none are provided and skip creating the table.

miso-web/src/main/webapp/scripts/hot_sample.js
```diff
@@ -0,0 +1,485 @@
...
+          depends: 'externalName',
+          update: function (sam, flat, value, setReadOnly, setOptions, setData) {
+            if (!Utils.validation.isEmpty(flat.externalName)) {
+              getIdentities(HotUtils.counter);
+            }
+
+            function getIdentities(requestCounter) {
+              jQuery.ajax({
+                url: "/miso/rest/tree/identities",
+                data: "{\"identitiesSearches\":" + JSON.stringify(flat.externalName)
```
apmasell: Just do `JSON.stringify({identitySearches: flat.externalName, requestCounter: requestCounter})`.

miso-web/src/main/webapp/scripts/hot_sample.js
```diff
@@ -0,0 +1,485 @@
...
+            'discards', HotUtils.validator.permitEmpty),
+        HotUtils.makeColumnForInt('Thickness', (show['Tissue Processing'] && config.targetSampleClass.alias == 'Slide'),
+            'thickness', HotUtils.validator.permitEmpty),
+        {
+          header: 'Stain',
+          data: 'stainName',
+          type: 'dropdown',
+          trimDropdown: false,
+          source: Constants.stains.sort(function (a, b) {
+            return (a.category || '') > (b.category || '') ? 1 : ((b.category || '') > (a.category || '') ? -1 : 0);
```
apmasell: This can be `(a.category||'').localeCompare(b.category||'')`.

miso-web/src/main/webapp/scripts/hot.js
```diff
@@ -468,6 +517,33 @@ var HotUtils = {
...
+    if (!alreadyShown.length) {
+      // add this response to the server errors to display
+      HotUtils.serverErrors.push({ rows: 1, message: responseText.detail, status: serverStatus });
+    }
+    
+    if (HotUtils.serverErrors.length) {
+      var serverErrorMessages = document.getElementById('serverErrors');
+      
+      serverErrorMessages.innerHTML = '<ul>' + HotUtils.serverErrors.map(function(error) {
+        return '<li>' + error.rows + ' row' + (error.rows == 1 ? '' : 's') + ': ' + error.status + '. ' + error.message + '</li>';
```
callunity: This syntax would result in a message like:
`Bad Request. Must give search terms to look up identities on 2 rows`
`Internal Server Error. NullPointerException on 2 rows`
That doesn't quite parse correctly to me because it's unclear that the "on 2 rows" is independent of the error message.
Perhaps something like the following?
`2 Bad Request errors. Must give search terms to look up identities`
`2 Internal Server Error errors. NullPointerException`

miso-web/src/main/webapp/scripts/hot_sample.js
callunity: The check for `config.projects` exists because there's no need to load all the projects during propagate so it should only do the sort if `config.projects` exists. We could indeed put check up in `requestConfiguration` to confirm whether `config.projects && config.create && !config.propagate` and error/dialog if not. Would you like me to do that in this PR or in another?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/util/BulkCreateTableBackend.java
```diff
@@ -18,15 +18,20 @@
...
-  public BulkCreateTableBackend(String targetType, Class<? extends Dto> dtoClass, String name, Dto dto) {
+  public BulkCreateTableBackend(String targetType, Class<? extends Dto> dtoClass, String name, Dto dto, Integer quantity) {
     super(targetType, dtoClass);
     this.name = name;
     this.dto = dto;
+    this.quantity = quantity;
   }
 
   public final ModelAndView create(ModelMap model) throws IOException {
-    List<Dto> dtos = Collections.singletonList(dto);
```
callunity: Very handy! I've made use of that.

miso-web/src/main/webapp/scripts/hot.js
apmasell: Sure. That's easier to understand.

miso-web/src/main/webapp/scripts/hot_sample.js
apmasell: Yes. I think that check should go in since it's not throwing on the backend.

miso-web/src/main/webapp/scripts/hot_sample.js
```diff
@@ -0,0 +1,490 @@
...
+              return statusToInt(a.status) - statusToInt(b.status);
+            }).map(function (s) { return s.description; });
+            statuses.unshift('Not Ready'); // can't chain this because unshift returns the length of the array
+            return statuses;
+          })(),
+          validator: HotUtils.validator.requiredText,
+          unpack: function (sam, flat, setCellMeta) {
+            flat.detailedQcStatusDescription = Utils.array.maybeGetProperty(
+                Utils.array.findFirstOrNull(
+                    Utils.array.idPredicate(sam.detailedQcStatusId), Constants.detailedQcStatuses),
```
djcooke: QC Status should default to "Not Ready" if null. Currently if you bulk edit Samples with null DetailedQcStatus (Not Ready), you'll have to re-select "Not Ready" before re-saving

callunity: This came from the lab -- Danielle preferred that it show up blank so she can double-click to fill down, rather than have to change the top one and drag down to the bottom of the table fill in the others. (Handsontable double-click-to-fill only fills in empty cells)

djcooke: That's fine on create, but on update too? That means you have to choose from the dropdown every time you bulk edit samples that were set to 'Not Ready.' That's confusing if you were only intending to update other fields

miso-web/src/main/webapp/scripts/hot.js
```diff
@@ -561,30 +570,41 @@ var HotUtils = {
...
+    baseobj.validator = HotUtils.validator.requiredAutocomplete, // if it's not
+                                                                  // a required
+                                                                  // field,
+                                                                  // '(None)'
+                                                                  // will be
+                                                                  // present in
+                                                                  // the list
+                                                                  // and
+                                                                  // selected by
+                                                                  // default
```
djcooke: I'm guessing this commit was the result of running a formatter?

callunity: :frowning_face:

PR #931, opened 2017-06-22
Exclude special characters during bulk library alias validation

callunity:


callunity:
Yes, I'd prefer them to be over-cautious than to push the envelope of what counts as "special". Though ideally, the naming guidelines would prevent most special characters...

PR #933, opened 2017-06-22
Glt 1845 Sample Performance Improvements

djcooke:
Does not completely solve the slowness, but it is an improvement. Boxes are still causing the most Sample-related performance issues, and that will be improved in a later PR.

Will also make separate PR for removing other XDerivedInfo entities after this one is approved


callunity:
Upon further investigation, the out-of-order sample IDs only occurred for samples created at the exact same millisecond, so it makes sense that they would come out in a random order.

PR #940, opened 2017-06-27
GLT-1845: fixed creationDate/creationTime fields

djcooke:
Fixed field names and un-removed Library and Pool creationDate

* creationDate = user-specified
* creationTime = generated save time


migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/DefaultMigrationTarget.java
```diff
@@ -608,7 +609,10 @@ private void resolvePoolables(Pool pool) throws IOException {
   }
 
   private void setPoolModifiedDetails(Pool pool) throws IOException {
-    if (pool.getId() == PoolImpl.UNSAVED_ID) pool.setCreationTime(timeStamp);
+    if (pool.getId() == PoolImpl.UNSAVED_ID) {
+      pool.setCreationTime(timeStamp);
+      pool.setCreationDate(timeStamp);
```
callunity: creationDate should be user-specified

PR #942, opened 2017-06-27
New infrastructure for MISO to access runscanners

apmasell:
JIRA: MISO-1758


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -506,4 +527,160 @@ public void setRunDao(RunStore runDao) {
...
   }
 
+  @Override
+  public boolean processNotification(NotificationDto dto) throws IOException, MisoNamingException {
+    final Date now = new Date();
+    User user = securityManager.getUserByLoginName("notification");
+    SecurityContextHolder.getContext().setAuthentication(new SuperuserAuthentication(user));
+    final Run run;
+
+    Run runFromDb = runDao.getByAlias(dto.getRunName());
```
callunity: I'm concerned about the ambiguity of the DTO field being called `runName` but using it to represent the run's `alias`. Is it possible to disambiguate this, maybe by renaming the DTO field to `runAlias`?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -506,4 +527,160 @@ public void setRunDao(RunStore runDao) {
...
+    }
+    run.setSequencerReference(sequencer);
+
+    final Collection<SequencerPartitionContainer> containers = containerService.listByBarcode(dto.getContainerId());
+    switch (containers.size()) {
+    case 0:
+      SequencerPartitionContainer newContainer = new SequencerPartitionContainerImpl(user);
+      newContainer.setPlatform(sequencer.getPlatform());
+      newContainer.setCreator(user);
+      newContainer.setIdentificationBarcode(dto.getContainerId());
```
callunity: Similarly, I think there's ambiguity with `containerId` -- I would have expected it would be a Long, and wouldn't have expected it to be the serial number...

miso-web/src/main/resources/internal/miso.properties
```diff
@@ -59,6 +59,9 @@ miso.notification.interop.enabled:false
 miso.notification.server.host:your.notification.server
 miso.notification.server.port:8787
 
+# A white-space delimited list of run scanner instances to poll
+miso.runscanner.urls:http://localhost:8080/runscanner
```
callunity: Should this be added to the `miso.properties` in `resources/external` as well? If so, please make some note in the PR (title or commit message) that this is a configuration change so it can be easily found in the release notes.

apmasell: Done. Worth pointing out that 1) it causes non-critical failure 2) runscanner fails to deploy beside a MISO instance properly due to log4j weirdness, 3) the VisionMate is set to localhost.

callunity: Is it worth noting 1) and 2) somewhere less ephemeral?

apmasell: Well, I'd like to fix 2. Run scanner is going to need a bunch of docs, but I want to wait until it's deployed and working and we've tinkered with it.

runscanner/src/main/webapp/pages/status.jsp
```diff
@@ -43,14 +43,15 @@
...
+        <tr><td>Time Since Last Scan</td><td>${timeSinceLastScan}</td></tr>
         <tr><td>Processed Runs</td><td>${finished}</td></tr>
         <tr><td>Waiting Runs</td><td>${scheduled}</td></tr>
       </table>
 
       <h1>Processors</h1>
       <table>
         <tr><th>Name</th><th>Platform</th></tr>
-        <c:forEach items="${processors}" var="processor"><td>${processor.name}</td><td>${processor.platformType}</td></c:forEach>
+        <c:forEach items="${processors}" var="processor"><tr><td>${processor.name}</td><td>${processor.platformType}</td></td></c:forEach>
```
djcooke: Last `td` closing tag should be `tr`

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -506,4 +527,160 @@ public void setRunDao(RunStore runDao) {
...
+      if (possibleParameters.size() == 1) {
+        isMutated |= run.getSequencingParameters() == null || possibleParameters.get(0).getId() != run.getSequencingParameters().getId();
+        run.setSequencingParameters(possibleParameters.get(0));
+      }
+    }
+
+    if (isNew) {
+      create(run);
+    } else if (isMutated) {
+      update(run);
```
djcooke: Call `save(Run)` to skip DB fetches and other unnecessary things that happen in `update(Run)`. The call to `setChangeDetails` should be moved from `create` and `update` methods to the start of `save` to support this

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ChangeLoggable.java
```diff
@@ -10,8 +11,15 @@
...
 public interface ChangeLoggable {
 
+  public static final Pattern COMMA = Pattern.compile(",");
+
   public Collection<ChangeLog> getChangeLog();
 
   public ChangeLog createChangeLog(String summary, String columnsChanged, User user);
 
+  public default boolean didSomeoneElseChangeColumn(String referenceColumnName, User nonHumanUser) {
+    return getChangeLog().stream().filter(cl -> cl.getUser().getUserId() != nonHumanUser.getUserId())
```
djcooke: This means the lazy-loaded changelogs will be initialized for every notification processed. Might not matter, but since this is only useful for runs and only for two fields, would it be better to just include a boolean field for each to mark not-overrideable-by-runscanner?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/RunScannerClient.java
```diff
@@ -0,0 +1,113 @@
...
+            ProgressiveResponseDto.class);
+        entry.getValue().update(response);
+        results.addAll(response.getUpdates());
+      } catch (Exception e) {
+        log.error("Failed to get runs from " + entry.getKey(), e);
+        serverFailures.labels(entry.getKey()).inc();
+      }
+    }
+    // Perform this in a transaction
+    client.process(results);
```
djcooke: Why control the transaction here? RunService will already create a separate transaction for each run save. If one run fails to save, I don't think it should affect other runs?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -506,4 +527,160 @@ public void setRunDao(RunStore runDao) {
     this.runDao = runDao;
   }
 
+  @Override
+  public boolean processNotification(NotificationDto dto) throws IOException, MisoNamingException {
```
apmasell: We discussed this before and this is the design you requested.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/RunScannerClient.java
```diff
@@ -0,0 +1,113 @@
...
+import uk.ac.bbsrc.tgac.miso.dto.NotificationDto;
+import uk.ac.bbsrc.tgac.miso.dto.ProgressiveRequestDto;
+import uk.ac.bbsrc.tgac.miso.dto.ProgressiveResponseDto;
+
+import io.prometheus.client.Counter;
+import io.prometheus.client.Gauge;
+
+@Service
+@Transactional(rollbackFor = Exception.class)
+public class RunScannerClient {
```
apmasell: I can move this service, but since `LatencyHistogram` is also used by the run scanner itself, it can't be moved.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ChangeLoggable.java
apmasell: I say we wait and see how it performs.

apmasell:
Fixed.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -528,148 +520,167 @@ public void setRunDao(RunStore runDao) {
...
-      filterParameters = p -> true;
       break;
     default:
       throw new NotImplementedException();
     }
 
+    isMutated |= updateSequencingParameters(target, user, filterParameters, sequencer);
+
+    if (isNew) {
+      setChangeDetails(target);
```
djcooke: `setChangeDetails` could be called at the start of the `save` method, in which case all other calls to `setChangeDetails` could be removed from this class

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
djcooke: Sorry I was unclear. Thanks for adjusting

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/service/RunScannerClient.java
```diff
@@ -57,34 +65,14 @@
...
-      } catch (Exception e) {
-        log.error("Failed to save run: " + dto.getRunAlias(), e);
-        saveFailures.inc();
-      }
-    }
-    lock.release();
-  }
-
-  @Scheduled(fixedDelay = 5 * 60 * 1000)
+  @Scheduled(fixedDelayString = "${miso.runscanner.interval}")
```
djcooke: Use `"${miso.runscanner.interval:300000}"` to make the property optional (startup won't fail if the property is missing)

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/service/RunScannerClient.java
```diff
@@ -101,7 +89,32 @@ public void scheduler() {
       }
     }
     // Perform this in a transaction
```
djcooke: comment should be removed

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/Scheduler.java
```diff
@@ -39,8 +40,13 @@
 @Service
 public class Scheduler {
   private static class FinishedWork {
+    Instant created = Instant.now();
     NotificationDto dto;
     int epoch;
+
+    public boolean shouldRerun() {
+      return !dto.getHealthType().isDone() && Duration.between(created, Instant.now()).toMinutes() < 10;
```
djcooke: Is this necessary with the scheduler hard-coded to scan every 15 minutes? If so, should this be greater than instead of less than?

apmasell: Yes. This is necessary. When a directory is scanned, it's unclear how long it will be in the queue for. You can imagine a scenario like this:

t = 0: find a whole lot of directories (including X, which is running) and queue them
t = 14min: process directory X and write a finished work entry
t = 15min: scan directories again. Requeue X
t = 15 min: backlog clears, start reprocessing X

apmasell:
Fixed.

PR #946, opened 2017-06-28
Improvement for lists

apmasell:


miso-web/src/main/webapp/scripts/list.js
```diff
@@ -23,81 +23,99 @@
...
+                .getElementById(elementId + "_toggle" + item.id);
+            if (element) {
+              element.checked = true;
+            }
+          });
+        }
+      });
+      staticActions.push({
+        "name" : "☐",
+        "title" : "Clear selected items",
```
callunity: What do you think of `Unselect all items` or `Unselect all`?

apmasell: How about “Select none”? I'm looking at GMail...

callunity: I have a mild preference for `Deselect all` because `Select`/`Deselect` is more prominent than `all`/`none`, but it's alt text and I'd really be happy with either `Deselect all` or `Select none`.

apmasell: Okay. I went with deselect because I found this war entertaining: https://english.stackexchange.com/questions/18465/unselect-or-deselect

callunity: :heart_eyes_cat:

miso-web/src/main/webapp/scripts/list.js
```diff
@@ -97,6 +97,7 @@ ListUtils = (function() {
           'bAutoWidth' : false,
           'iDisplayLength' : 25,
           'iDisplayStart' : 0,
+          'sScrollX' : '100%',
```
djcooke: Is this supposed to add a horizontal scrollbar instead of overflowing list contents off the right side of the screen? It doesn't seem to be working if so

PR #948, opened 2017-06-28
GLT-1853 Avoid redrawing order table when adjusting visibility of columns

tonydebat:
GLT-1853 Avoid redrawing order table when dynamically adjusting the visibility
of the columns. This reduces the number REST calls made and improves the display
time of the table.
Made retrieval of OrderCompletion_Items EAGER. This reduces the total number of
SQL queries hibernate issues when displaying the order page.


tonydebat:
Removed duplicates.

PR #950, opened 2017-06-29
Convert indicies to common list infrastructure

apmasell:


miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -1609,21 +1609,32 @@ public static LibraryStrategyTypeDto asDto(LibraryStrategyType from) {
...
+      dto.setFamily(asDto(from.getFamily(), false));
+    }
     return dto;
   }
 
   public static IndexFamilyDto asDto(IndexFamily from) {
+    return asDto(from, true);
+  }
+
+  private static IndexFamilyDto asDto(IndexFamily from, boolean includeChildren) {
```
callunity: is `includeChildren` ever used?

apmasell: Yes. In the `asDto(IndexFamily)` it's true, while in the `asDto(Index)`, it's false.

callunity: Right, so should the `asDto` method include a check for `includeChildren` before setting the indices?

apmasell: Sorry, I was thinking about the callers and you meant in the method. :-/ Fixed.

callunity: I can see how you would, my initial comment turns out to have been ambiguous!

miso-web/src/main/webapp/scripts/list.js
```diff
@@ -258,6 +258,9 @@ ListUtils = (function() {
       }
     },
     render : {
+      archived : function(data, type, full) {
+        return data ? "🗄" : "";
```
djcooke: Not sure about this character. I thought it wasn't showing up right until I pasted it into http://www.babelstone.co.uk/Unicode/whatisit.html. In Firefox it does look like it's supposed to - it's just too tiny to tell what it is. On Chrome on my Macbook, it shows up as an empty square. Checkmark might be clearer

Ideally, archived indices just wouldn't be included in the list with current indices

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateIndexDao.java
```diff
@@ -65,24 +76,51 @@ public IndexFamily getIndexFamilyByName(String name) {
...
 
   @Override
-  public List<Index> listAllIndices(PlatformType platformType) {
-    Query query = currentSession().createQuery("from Index where family.platformType = :platform");
-    query.setParameter("platform", platformType);
-    @SuppressWarnings("unchecked")
-    List<Index> list = query.list();
-    return list;
+  public String getProjectColumn() {
+    return null;
```
djcooke: `HibernatePaginatedDataSource#restrictPaginationByProjectId` will try adding a restriction to the criteria based on this. Should that method check for null and report to errorHandler instead?

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateIndexDao.java
```diff
@@ -65,24 +76,51 @@ public IndexFamily getIndexFamilyByName(String name) {
...
+  }
+
+  @Override
+  public Iterable<String> listAliases() {
+    return STANDARD_ALIASES;
+  }
+
+  @Override
+  public String propertyForDate(Criteria criteria, DateType type) {
+    return null;
```
djcooke: Nulls on this aren't reported to the errorHandler either

miso-web/src/main/webapp/scripts/list.js
callunity: Maybe tabbify it into archived and active?

apmasell: Since we have it tabbified by platform, I don't think more tabs is the answer. I'll limit these to just active ones.

apmasell:
Fixed.

miso-web/src/main/webapp/scripts/list.js
djcooke: If the archived ones are hidden completely, could that be noted on the page somewhere?

PR #952, opened 2017-06-29
Box Save Performance Improvement

djcooke:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultBoxService.java
```diff
@@ -175,6 +176,15 @@ public long save(Box box) throws IOException {
...
       StringBuilder message = new StringBuilder();
 
+      // get persisted version of new box contents before change
+      List<BoxableId> ids = box.getBoxables().values()
+          .stream()
+          .map(b -> b.getId())
+          .collect(Collectors.toList());
+      Map<BoxableId, BoxableView> oldOccupants = boxStore.getBoxableViewsByIdList(ids)
+          .stream()
+          .collect(Collectors.toMap(BoxableView::getId, b -> b));
```
apmasell: There's a `Function.identity()` if you want to do that instead of `b -> b`.

PR #954, opened 2017-06-29
Add pool description to orders page

apmasell:
JIRA: GLT-1676


callunity:
It's a little strange to have it present in the Order Completion page on the pool (do we want to rename that to "Order Status"?) but not a huge issue.

PR #957, opened 2017-06-30
Show longest index on completions page

apmasell:
JIRA: GLT-1872


miso-web/src/main/webapp/scripts/list_completions.js
```diff
@@ -67,6 +67,12 @@ ListTarget.completion = {
           "iSortPriority" : 0,
           "include" : true
         }, {
+          "sTitle" : "Index Length",
```
callunity: Perhaps `Longest Index` instead of `Index Length`?

PR #958, opened 2017-06-30
GLT-1884: removed miso-service dependency on miso-dto

djcooke:


djcooke:
Rebased to fix build in CI. No changes to the code in PR

PR #960, opened 2017-07-04
Convert samples to new list infrastructure and support replicates dur…

apmasell:
…ing propagation

JIRA: GLT-1541


miso-web/src/main/webapp/scripts/hot_sample.js
```diff
@@ -586,41 +611,102 @@ HotTarget.sample = (function() {
...
+            if (targets.length == 0) {
+              alert("No propagation is possible from the samples.");
+              return;
+            }
+            
+            Utils.showDialog(targets.length > 1 ? 'Propagate Samples'
+                : ('Propagate to ' + targets[0].name), 'Propagate', [ {
+              property : 'replicates',
+              type : 'int',
+              label : 'Replicates',
```
apmasell: This is the terminology the lab asked for.

callunity: "Replicates" is a term that would be familiar to most lab users.

miso-web/src/main/webapp/scripts/list_samples.js
```diff
@@ -0,0 +1,214 @@
...
+            }
+          }, ]);
+    }
+    return actions;
+    
+  },
+  createStaticActions : function(config, projectId) {
+    return [
+        {
+          name : "Create",
```
callunity: The "Lookup Identities" button is indeed gone. I'll make the correction to make the external name look required; the Sex column indicates it's required in #964 but I can change that to defaulting to Unknown. (It gets cranky that "" aka empty string isn't in the Sex dropdown list, since it's represented by Unknown.)

apmasell: I fixed the create-from-project error. The external name and sex column is out of scope for this PR.

miso-web/src/main/webapp/scripts/hot_sample.js
djcooke: :+1:

PR #961, opened 2017-07-04
Validation caching

apmasell:


djcooke:
:100: Makes a big difference when saving lots of samples/libraries

PR #962, opened 2017-07-04
Add custom-named identities

callunity:
And several small bug fixes and improvements


callunity:
@apmasell made the point that it might make sense to expand the regex to validate 6 digits, instead of allowing the lab to indicate "Non-standard alias" on the identities that they are creating.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/validation/OicrLibraryAliasValidator.java
```diff
@@ -3,7 +3,7 @@
 public class OicrLibraryAliasValidator extends RegexValidator {
 
   private static final String commonPrefix = "([A-Z\\d]{3,5})_";
-  private static final String illuminaRegex = "(\\d{3,4}|\\d[CR]\\d{1,2})_(nn|[A-Z][a-z])_([A-Zn])_(SE|PE|MP|TR|\\?\\?)_(nn|\\d{2,6}|\\dK)_(TS|EX|CH|BS|WG|TR|WT|SM|MR|AS|\\?\\?)";
+  private static final String illuminaRegex = "(\\d{3,}|\\d[CR]\\d{1,2})_(nn|[A-Z][a-z])_([A-Zn])_(SE|PE|MP|TR|\\?\\?)_(nn|\\d{2,6}|\\dK)_(TS|EX|CH|BS|WG|TR|WT|SM|MR|AS|\\?\\?)";
```
djcooke: Would rather have an upper limit on the number of digits

callunity: Shall we cap it at 6, since that is what the lab is now requesting?

djcooke: I'd say yes - keep it as constrained as possible

PR #964, opened 2017-07-05
Fix receive stock

callunity:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
```diff
@@ -168,6 +170,23 @@ public SampleDto createSample(@RequestBody SampleDto sampleDto, UriComponentsBui
           SampleClass stockClass = sampleClassService.inferStockFromAliquot(sampleClass);
           dto.setStockClassId(stockClass.getId());
         }
+      } else if (sampleDto instanceof SampleStockDto) {
```
apmasell: Should this be `else`? If you are receiving an aliquot, don't you have to infer both stock and tissue classes?

callunity: Hmm good point. It should be an else, but inferring the tissue class should also happen within the aliquot block.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleClassService.java
```diff
@@ -113,4 +114,25 @@ public SampleClass inferStockFromAliquot(SampleClass sampleClass) {
     return stockClass;
   }
 
+  @Override
+  public SampleClass inferTissueFromStock(SampleClass sampleClass) {
```
djcooke: With current OICR SampleClasses, there's only one Tissue class. With our previous Tissue classes, this would always fail because multiple tissue classes were valid for every Stock class. I'm not sure it's worth the DB lookups to pretend that you could just toss any classes into the DB and expect it to work. We could just treat it like Identity instead (expect there to only be one class in the category)?

apmasell: That might be true now, but I don't know that it will be true in general. Given there are separate, unambiguous tissues, I don't see the harm in this.

PR #968, opened 2017-07-06
Fix stain pack

callunity:


miso-web/src/main/webapp/scripts/hot_sample.js
```diff
@@ -440,6 +440,7 @@ HotTarget.sample = (function() {
             },
             pack : function(sam, flat, errorHandler) {
               if (!Utils.validation.isEmpty(flat.stainName)) {
+                sam.stain = {};
```
apmasell: Since the user can switch from stained to unstained, I think this should be:
```
sam.stain = Utils.array.findFirstOrNull(Utils.array.namePredicate(flat.stainName), Constants.stains);
```

callunity: Hmm I'd forgotten that they can switch from stained to unstained. This will become more than one line.

PR #975, opened 2017-07-06
GLT-1618 generate aliases for tissues propagated from tissues

callunity:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/generation/OicrSampleAliasGenerator.java
```diff
@@ -43,7 +43,12 @@ public String generate(Sample sample) throws MisoNamingException, IOException {
         return addSiblingTag(parent.getAlias(), detailed);
       }
       if (isTissueSample(parent)) {
-        return addSiblingTag(parent.getAlias(), detailed);
+        if (isTissueSample(detailed) && isIdentitySample(parent.getParent())) {
```
apmasell: The grand parent might not be a tissue if multiple tissue-to-tissue layers become possible. Stick this in `LimsUtils`:
```
public static <T extends DetailedSample> getParent(Class<T> clazz, DetailedSample start) {
for (DetailedSample current = start.getParent(); current != null; current = deproxify(current.getParent())) {
if (clazz.isInstance(current)) {
return (T) current;
}
}
return null;
}
```
and then do:
```
 if (isTissueSample(detailed) && isIdentitySample(parent.getParent())) {
// tissues parented to tissues
return generateTissueAlias((SampleTissue) detailed, LimsUtils.getParent(SampleIdentity.class, parent));
```

callunity: Done, and also removed the check for if the identity is the immediate grandparent that was immediately above the `// tissues parented to tissues` line

PR #977, opened 2017-07-06
fix identity lookup for samples from Edit Project page

callunity:
GLT-1898

### DIFF
--- a/miso-web/src/main/webapp/scripts/hot_sample.js
+++ b/miso-web/src/main/webapp/scripts/hot_sample.js
@@ -302,20 +302,18 @@ HotTarget.sample = (function() {
                           if (data.requestCounter == requestCounter) {
                             var label = Constants.isDetailedSample
                                 ? 'shortName' : 'name';
-                            var selectedProjectId = Utils.array
-                                .maybeGetProperty(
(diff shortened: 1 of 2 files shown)


miso-web/src/main/webapp/scripts/hot_sample.js
```diff
@@ -324,10 +322,10 @@ HotTarget.sample = (function() {
                                     return sam.alias + " -- " + sam.externalName;
                                   });
                             }
-                            var hasIdentityInProject = (identitiesSources.length > 0 && data.matchingIdentities[0].projectId == selectedProjectId);
+                            var hasIdentityInProject = (identitiesSources.length > 0 && data.matchingIdentities[0].projectId == selectedProject.id);
                             if (!hasIdentityInProject) {
                               identitiesSources
-                                  .unshift("First Receipt (" + flat.projectAlias + ")");
+                                  .unshift("First Receipt (" + (flat.projectAlias || selectedProject[label]) + ")");
```
apmasell: Might as well just be `selectedProject[label]`.

callunity: That's true.

PR #988, opened 2017-07-07
Fix save new runs via notification server bug

callunity:
GLT-1913


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/manager/MisoRequestManager.java
```diff
@@ -554,11 +554,19 @@ public SequencerPartitionContainer saveSequencerPartitionContainer(SequencerPart
         container.setSecurityProfile(securityProfileStore.get(securityProfileStore.save(container.getSecurityProfile())));
         container.setPlatform(platformStore.get(container.getPlatform().getId()));
         container.setLastModifier(securityStore.getUserById(getCurrentUser().getUserId()));
```
djcooke: Should set as creator as well

PR #971, opened 2017-07-06
Convert users and groups to new list utilities

apmasell:

PR #990, opened 2017-07-10
Collapse common static lists in JSPs

apmasell:
- Introduce a tag for creating a change log
- Introduce a tag for creating a static list
- Remove unnecessary CSS "list" classes


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSequencerPartitionContainerController.java
```diff
@@ -126,7 +128,7 @@ public ModelAndView setupForm(@PathVariable Long containerId, ModelMap model) th
 
       model.put("formObj", container);
       model.put("container", container);
-      model.put("containerRuns", runService.listByContainerId(container.getId()));
+      model.put("containerRuns", runService.listByContainerId(container.getId()).stream().map(Dtos::asDto).collect(Collectors.toList()));
```
callunity: Tangent: I was testing this and it looked like there were no runs available ("No matching records found") but it turns out it was just the search bar being populated and I had forgotten/hadn't noticed because I was not in the context of the List Runs page. It's out of scope for this PR, but I'd like to get feedback from the lab about whether they experience these "forgotten filters" as well, and if we should do anything about it. (Unless @djcooke 's suggestion for storing the state of the DataTables will fix this?)
Also a bug with this: deleting the search term removes the filter, but then the search term reappears in the filter bar (but doesn't do the server lookup). 

[JIRA ticket](https://jira.oicr.on.ca/browse/GLT-1937)

miso-web/src/main/webapp/pages/editExperiment.jsp
```diff
@@ -359,30 +359,7 @@
...
-            <td>${change.user.fullName} (${change.user.loginName})</td>
-            <td><b>${change.summary}</b></td>
-            <td>${change.time}</td>
-          </tr>
-        </c:forEach>
-        </tbody>
-      </table>
-    </div>
-  </c:if>
+    <miso:changelog item="${experiment}"/>
```
callunity: The Datatables source JS needs to be included on the page.

miso-web/src/main/webapp/pages/editLibrary.jsp
```diff
@@ -33,23 +33,10 @@
 <script src="<c:url value='/scripts/datatables_utils.js'/>" type="text/javascript"></script>
 <script src="<c:url value='/scripts/jquery/datatables/js/jquery.dataTables.min.js'/>" type="text/javascript"></script>
```
callunity: :tada: :smile: this whole file

miso-web/src/main/webapp/scripts/list_changelog.js
```diff
@@ -0,0 +1,55 @@
...
+        {
+          "sTitle" : "Summary",
+          "mData" : "summary",
+          "include" : true,
+          "iSortPriority" : 0
+        }, {
+          "sTitle" : "Time",
+          "mData" : "time",
+          "include" : true,
+          "iSortPriority" : 1
```
callunity: This filters by time descending but historically, these have always filtered by time ascending. Is this a deliberate change?

apmasell: No, it's an implicit change by the new list infrastructure. It's being changed in the #971 PR. I can do the same here.

miso-web/src/main/webapp/pages/editExperiment.jsp
apmasell: Fixed.

PR #993, opened 2017-07-10
Correctly filter by external name and project

callunity:
GLT-1915

### DIFF
--- a/miso-web/src/main/webapp/scripts/hot_sample.js
+++ b/miso-web/src/main/webapp/scripts/hot_sample.js
@@ -322,15 +322,24 @@ HotTarget.sample = (function() {
                                     return sam.alias + " -- " + sam.externalName;
                                   });
                             }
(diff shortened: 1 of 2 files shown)


miso-web/src/main/webapp/scripts/hot_sample.js
```diff
@@ -322,15 +322,24 @@ HotTarget.sample = (function() {
                                     return sam.alias + " -- " + sam.externalName;
                                   });
                             }
-                            var hasIdentityInProject = (identitiesSources.length > 0 
-                                && data.matchingIdentities[0].projectId == selectedProject.id
-                                && data.matchingIdentities[0].externalName == flat.externalName);
-                            if (!hasIdentityInProject) {
+                            var indexOfMatchingIdentityInProject = function () {
```
apmasell: You can use `findIndex` https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

PR #997, opened 2017-07-11
Use Illumina Interop library to scan data for Run Scanner

apmasell:
JIRA: GLT-1759


runscanner-illumina/main.cpp
```diff
@@ -0,0 +1,86 @@
...
+    std::stringstream date_buffer;
+    date_buffer << std::put_time(std::localtime(&t), "%Y-%m-%d");
+    result["completionDate"] = date_buffer.str();
+    has_extraction = true;
+  }
+  is_complete &= has_extraction;
+
+  /* We can't tell the difference between the stopped or running states, so we
+   * just assume running if it isn't finished.*/
+  result["healthType"] = is_complete ? "Completed" : "Running";
```
djcooke: What about if it has failed?

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
```diff
@@ -34,18 +27,18 @@ public RunProcessor(PlatformType platformType, String name) {
...
+  public final PlatformType getPlatformType() {
     return platformType;
   }
 
   public Stream<File> getRunsFromRoot(File root) {
     return Arrays.stream(root.listFiles(File::isDirectory));
   }
 
-  public abstract NotificationDto process(File runDirectory, TimeZone tz) throws IOException;
+  public abstract NotificationDto process(File runDirectory, TimeZone tz) throws Exception;
```
djcooke: Please declare only specific checked exceptions. If there should be expected exceptions that don't make sense to rethrow as IOExceptions, but should be grouped together, add a "ProcessingException" class or something

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/processor/StandardIllumina.java
```diff
@@ -0,0 +1,37 @@
...
+  public StandardIllumina() {
+    super(PlatformType.ILLUMINA, "default");
+  }
+
+  @Override
+  public NotificationDto process(File runDirectory, TimeZone tz) throws Exception {
+    Path runPath = runDirectory.toPath().toAbsolutePath();
+    ObjectMapper mapper = new ObjectMapper();
+    mapper.registerModule(new JavaTimeModule())
+        .setDateFormat(new ISO8601DateFormat());
```
djcooke: Mapper with options should probably be standardized between all RunProcessors

runscanner-illumina/README.md
```diff
@@ -0,0 +1,17 @@
+# Illumina Support for Run Scanner
+To use the Illumina support in Run Scanner, this additional package must be built.
+
+You will need a copy of [jsoncpp](https://github.com/open-source-parsers/jsoncpp) and an AutoTools-capable build environment. On Debian/Ubuntu:
```
djcooke: Did you look into Maven plugins to automate this? There's Native Maven Plugin that might do the trick: http://www.mojohaus.org/maven-native/native-maven-plugin/

runscanner-illumina/main.cpp
apmasell: There's also no way to detect that.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
apmasell: I do not understand this rationale, but fine. The basic policy we have built is convert all checked exceptions to runtime exceptions rather than just allow checked exceptions to propagate.

runscanner-illumina/README.md
apmasell: Yes. I spent several days fighting with that plugin, but it's really not capable of handling the Illumina library as it's built.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/processor/StandardIllumina.java
apmasell: Fixed.

runscanner-illumina/main.cpp
djcooke: The current `IlluminaTransformer` can detect at least some errors to report failed status, and classifies some other states as 'unknown.' I get that the InterOp library can't do it, but we can. Does it not make sense to use the same old checks, at least for sure failures?

apmasell: Those are the set of checks that keep setting runs to Unknown status. There's basically two that exist in the current code:

- look for "Application exited before completion" in the RTA logs
- look for `runParameters.xml*` as that might indicate failure

The first one has some merit. I think the second one doesn't.

runscanner/src/main/java/uk/ac/bbsrc/tgac/miso/runscanner/RunProcessor.java
djcooke: I don't know where this basic policy came from? It seems contrary to all best practices regarding exceptions in Java

runscanner-illumina/main.cpp
djcooke: Yeah, I don't remember ever seeing the `runParameters.xml*` file. May have been an EI-specific thing

PR #998, opened 2017-07-11
GLT-1920: fix creating samples with existing identities

djcooke:


miso-web/src/main/webapp/scripts/hot_sample.js
```diff
@@ -263,7 +263,13 @@ HotTarget.sample = (function() {
...
             },
             pack : function(sam, flat, errorHandler) {
-              sam.externalName = flat.externalName;
+              if (flat.potentialIdentities) {
+                var selectedIdentity = Utils.array.findFirstOrNull(function(item) {
+                  return item.label == flat.identityAlias}, flat.potentialIdentities);
+                if (!selectedIdentity) {
+                  sam.externalName = flat.externalName;
+                }
+              }
```
callunity: This needs to handle the `else` case during new identity creation -- `flat.potentialIdentities` is false at that point and `sam.externalName` is not getting set.

apmasell: It should probably just trigger the error handler and give up. There's a `;` missing after `identityAlias` on 268. This function and the other pack function are almost the same and set duplicate fields in the sample. I think it would be best to collapse them in a function like:
```
var getIdentityProperty = function(flat, property) {
return Utils.array.maybeGetProperty(Utils.array.findFirstOrNull(function(item) {
            return item.label == flat.identityAlias;
}, flat.potentialIdentities || []), property);
}
```

And then consume it as:
`sam.parentAlias = getIdentityProperty(flat, 'parentAlias');` and `sam.externalName = getIdentityProperty(flat, 'externalName');`

callunity: Will that allow new identities to be created again? They're erroring due to `sam.externalName` not getting set.

apmasell: Hmm...I'm thinking the two cases need to be made into separate columns...

PR #999, opened 2017-07-11
Disable filter box when search is in progress

apmasell:
And increase the search time delay

JIRA: GC-1166


apmasell:
Sure. There are other ways around the problem, potentially. Currently, I replace the known data with the new data, but it could append. I'm also not thrilled about there being multiple in-flight requests. Maybe we need an enter-to-search rather than a timeout.

callunity:
I'm a fan of enter-to-search.

apmasell:
Enter-to-search it is. Also using the `bStateSave`.

PR #1001, opened 2017-07-12
Refactor and fix problems with inferring tissue for tissue processing

apmasell:
This fixes a bug where tissue class was not inferred for tissue processing
samples created directly. It also shares that logic for all sample class
inference. Additionally, it limits the list of available option so that new
samples cannot be created from a class with no inference path to an identity.

JIRA: GLT-1922

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Sample.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Sample.java
@@ -62,7 +62,6 @@ public interface Sample
   public static final Long UNSAVED_ID = 0L;
   /** Field PREFIX */
   public static final String PREFIX = "SAM";
-
   public void setId(long id);
 
   /**
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
@@ -12,11 +12,11 @@
  *
  * MISO is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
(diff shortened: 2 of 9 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Sample.java
```diff
@@ -62,6 +64,8 @@
   public static final Long UNSAVED_ID = 0L;
   /** Field PREFIX */
   public static final String PREFIX = "SAM";
+  List<String> CATEGORIES = Arrays.asList(SampleIdentity.CATEGORY_NAME, SampleTissue.CATEGORY_NAME,
```
djcooke: Should be immutable (`Arrays.asList` returns a mutable `List`). SampleCategory should probably become an enum at this point. Should be on `SampleClass`, or at least `DetailedSample` rather than plain `Sample` if it stays within another class

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleController.java
```diff
@@ -218,6 +199,20 @@ public SampleDto createSample(@RequestBody SampleDto sampleDto, UriComponentsBui
     return created;
   }
 
+  private <T extends DetailedSampleDto> Long inferIntermediateSampleClassId(T dto, Function<? super T, Long> getId, String currentCategory,
```
djcooke: Please rename parameters for clarity or add Javadoc-style comment to explain
* getId = getChildClassId (or just accept childClassId instead of Function)
* currentCategory = childClassCategory

apmasell:
Fixed.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/Sample.java
```diff
@@ -62,6 +65,8 @@
   public static final Long UNSAVED_ID = 0L;
   /** Field PREFIX */
   public static final String PREFIX = "SAM";
+  List<String> CATEGORIES = Collections.unmodifiableList(Arrays.asList(SampleIdentity.CATEGORY_NAME, SampleTissue.CATEGORY_NAME,
```
djcooke: Is there a reason to have this in plain `Sample` instead of `SampleClass` or `DetailedSample`?

apmasell: It was in `EditSampleController`, so anywhere seemed better than there.

PR #1006, opened 2017-07-17
Convert project to new list utilities

apmasell:


miso-web/src/main/webapp/scripts/list_project.js
```diff
@@ -0,0 +1,61 @@
...
+        window.location.href = '/miso/project/new';
+      }
+    } ];
+  },
+  createColumns : function(config, projectId) {
+    return [
+        ListUtils.idHyperlinkColumn("Name", "project", "id",
+            Utils.array.getName, 0),
+        ListUtils.labelHyperlinkColumn("Alias", "project", Utils.array.getId,
+            "alias", 1),
```
callunity: What do you think about sorting by alias ascending? (I don't find alias descending to be intuitive, but I could be unique in that.)

apmasell: Fixed.

PR #1011, opened 2017-07-19
Use strings in front end for double fields

callunity:
This will cause the numbers to be saved with user-defined precision, instead of Handsontable-implemented precision.


miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -405,7 +407,7 @@ private static DetailedSample toDetailedSample(DetailedSampleDto from) {
       to.setSynthetic(from.getSynthetic());
     }
     if (from.getConcentration() != null) {
-      to.setConcentration(from.getConcentration());
+      to.setConcentration(Double.valueOf(from.getConcentration()));
```
djcooke: Do note that precision can still be lost due to conversion to double here

callunity: Guess I should go full-on BigDecimal, eh?

PR #1019, opened 2017-07-26
Add graphs and summary information for Illumina runs

apmasell:
JIRA: GLT-1336

### DIFF
--- a/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
+++ b/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
@@ -10,6 +10,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.function.Consumer;
 import java.util.function.IntFunction;
@@ -29,6 +30,10 @@ import org.springframework.transaction.annotation.Transactional;
 import com.eaglegenomics.simlims.core.Note;
 import com.eaglegenomics.simlims.core.User;
 import com.eaglegenomics.simlims.core.manager.SecurityManager;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.google.common.collect.Lists;
 
 import uk.ac.bbsrc.tgac.miso.core.data.AbstractQC;
@@ -69,7 +74,6 @@ import uk.ac.bbsrc.tgac.miso.service.security.AuthorizedPaginatedDataSource;
 @Transactional(rollbackFor = Exception.class)
 @Service
 public class DefaultRunService implements RunService, AuthorizedPaginatedDataSource<Run> {
-
   private static final Logger log = LoggerFactory.getLogger(DefaultRunService.class);
 
   @Autowired
@@ -554,7 +558,7 @@ public class DefaultRunService implements RunService, AuthorizedPaginatedDataSou
     target.setLastModifier(user);
     boolean isMutated = false;
     isMutated |= updateField(source.getCompletionDate(), target.getCompletionDate(), target::setCompletionDate);
-    isMutated |= updateField(source.getMetrics().toString(), target.getMetrics(), target::setMetrics);
+    isMutated |= updateMetricsFromNotification(source, target);
     isMutated |= updateField(source.getFilePath(), target.getFilePath(), target::setFilePath);
     isMutated |= updateField(source.getStartDate(), target.getCompletionDate(), target::setCompletionDate);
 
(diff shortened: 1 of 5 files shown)


apmasell:
For testing purposes, build `runscanner-illumina` per the instructions and then do:
```
mkdir ~/mnt
sshfs -o idmap=user -o BatchMode=no hn.hpc.oicr.on.ca:/.mounts ~/mnt
 ./runscanner-illumina ~/mnt/labs/prod/archive/h1179/160406_SN7001179_0025_AC8CVBACXX/ | jq -r '("UPDATE Run SET metrics = '\''\(.metrics)'\'' WHERE runId = 1;")' | mysql
```
Add whatever parameters you need to `mysql`. You can then view run 1 and the graphs will be provided.

callunity:
Following the build instructions causes this error on my Ubuntu machine: `Unable to locate package build-essentials`

apmasell:
Updated.

callunity:
In the first graph, what is meant by `Cycles` ? In the graphs coming from the Notification Server, it was a representation of `completedCycles / totalCycles` but now it's only a single number. Is there another way of seeing how many cycles have completed and how many remain?

It seems like there's a lot of white space to the right of the graphs, which is making them look scrunched. Is there a way to reduce this?

apmasell:
I think the cycles can be changed. There's a cycle state object that has extracted, called, Q-scored. That number is Q-scored. There's also a total cycles. Should I display all 4? I have no idea which one was used before.

I don't see the whitespace problem, but the layout of the graphs becomes fixed because the highcharts don't seem to be able to resize dynamically. I'm not sure what to do about this.

callunity:
I was playing around with the CSS and wasn't able to come up with anything good for resizing the Highcharts either.

In staging, the run you chose displays `Cycles: 209/209`. Are there any numbers in the cycle state object that matches up with that? Does total cycles match up with that?

apmasell:
They're all the same when the run is complete.

apmasell:
I've added all the cycle states but they seem to be zero. I don't know if the cycle states information is missing from our runs. :-/

apmasell:
The cycle stats are defined as a range, but it seems that start == end, so I'm just using end.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -605,6 +609,57 @@ public boolean processNotification(Run source, int laneCount, String containerSe
...
+      target.setMetrics(source.getMetrics());
+      return true;
+    }
+
+    ObjectMapper mapper = new ObjectMapper();
+    ArrayNode sourceMetrics;
+    try {
+      sourceMetrics = mapper.readValue(source.getMetrics(), ArrayNode.class);
+    } catch (IOException e) {
+      log.debug("Impossible junk metrics were passed in for run " + target.getId(), e);
```
djcooke: Should probably be concerning. IOException may be thrown for other reasons too. Log at error level

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -605,6 +609,57 @@ public boolean processNotification(Run source, int laneCount, String containerSe
...
+      return false;
+    }
+    ArrayNode targetMetrics;
+    try {
+      targetMetrics = mapper.readValue(target.getMetrics(), ArrayNode.class);
+    } catch (IOException e) {
+      log.debug("The database is full of garbage metrics for run " + target.getId(), e);
```
djcooke: More concerning

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -605,6 +609,57 @@ public boolean processNotification(Run source, int laneCount, String containerSe
     return isNew;
   }
 
+  private boolean updateMetricsFromNotification(Run source, Run target) {
+    if (source.getMetrics() != null && source.getMetrics().equals(target.getMetrics())) return false;
+    if (source.getMetrics() == null && target.getMetrics() != null) {
+      return false;
```
djcooke: Can probably return false if source metrics are null, regardless of target metrics?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -605,6 +609,57 @@ public boolean processNotification(Run source, int laneCount, String containerSe
...
+    targetMetricsMap.putAll(sourceMetricsMap);
+    ArrayNode combinedMetrics = mapper.createArrayNode();
+    combinedMetrics.addAll(targetMetricsMap.values());
+    try {
+      target.setMetrics(mapper.writeValueAsString(combinedMetrics));
+    } catch (JsonProcessingException e) {
+      log.error("Failed to save data just unserialised.", e);
+      return false;
+    }
+    return false;
```
djcooke: Seems like true should be returned at this point

apmasell:
Fixed.

PR #1024, opened 2017-07-27
Delete notification server

apmasell:


callunity:
Please also delete NS properties from both `miso.properties` files

apmasell:
Fixed.

PR #1028, opened 2017-07-31
fix IT HOT text entry and dropdown scanning

djcooke:
Fix for error where some characters are missing after attempt to enter text into a Handsontable cell. Can't reproduce error outside of Travis, so don't bother reviewing until Travis is happy.

Contains a fix for scanning dropdown options as well

### DIFF
--- a/miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/element/HandsOnTable.java
+++ b/miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/element/HandsOnTable.java
@@ -1,6 +1,6 @@
 package uk.ac.bbsrc.tgac.miso.webapp.integrationtest.page.element;
 
-import static org.openqa.selenium.support.ui.ExpectedConditions.invisibilityOf;
+import static org.openqa.selenium.support.ui.ExpectedConditions.*;
 
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/element/HandsOnTable.java
```diff
@@ -56,16 +58,32 @@ public int getRowCount() {
...
+    Action action = new Actions(getDriver())
         .click(cell)
+        .sendKeys(Keys.DELETE)
         .sendKeys(text)
         .sendKeys(Keys.ENTER)
-        .build().perform();
+        .build();
+    // Odd behaviour of Chrome webdriver on Travis CI causes the first attempt to fail sometimes, so try up to 3 times
+    for (int retries = 0; retries < 3 && !text.equals(getText(columnHeading, rowNum)); retries++) {
+      action.perform();
```
apmasell: This should emit a useful error when it exceeded the retry limit.

djcooke: I did that at first, but then realized the method is also used in ways where it is expected to fail such as ensuring that fields cannot be written to. I think it's ideal to use the same method in testing both cases rather than have separate method that works differently when a failure is expected. That way, it's up to the test to assert the outcome as it should

djcooke: I'll add a comment to explain this though

djcooke:
Still fails sometimes

PR #1032, opened 2017-08-02
Crawl for errors

callunity:
GLT-1211


djcooke:
Test data SQL needs fixed up I think

PR #1035, opened 2017-08-03
Show empty cells on Containers page when no last run present

callunity:
Instead of "RUNnull" and "null"

GLT-1904

### DIFF
--- a/miso-web/src/main/webapp/scripts/list_container.js
+++ b/miso-web/src/main/webapp/scripts/list_container.js
@@ -50,7 +50,7 @@ ListTarget.container = {
         },
         ListUtils.idHyperlinkColumn("Last Run Name", "run", "lastRunId",
             function(container) {
-              return (container.lastRunId ? "RUN" + container.lastRunId : null);
+              return "RUN" + container.lastRunId;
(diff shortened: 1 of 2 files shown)


miso-web/src/main/webapp/scripts/list_container.js
```diff
@@ -50,7 +50,7 @@ ListTarget.container = {
         },
         ListUtils.idHyperlinkColumn("Last Run Name", "run", "lastRunId",
             function(container) {
-              return "RUN" + container.lastRunId;
+              return (container.lastRunId ? "RUN" + container.lastRunId : null);
```
apmasell: Can you modify the `idHyperlinkColumn` function to not display the link at all when the fetched ID is missing?

callunity: That's what my change in `lims.js#289` is supposed to do, and is showing as no text on my machine. Are you seeing text when you deploy it?

apmasell: Sorry, I am. I guess, is it necessary to do this as well?

PR #1037, opened 2017-08-03
Test creating & editing projects

callunity:
GLT-1976

### DIFF
--- a/miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/ProjectPageIT.java
+++ b/miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/ProjectPageIT.java
@@ -30,7 +30,7 @@ public class ProjectPageIT extends AbstractIT {
 
   @Test
   public void testSaveEditNewProject() {
-    // Goal: save and edit one project. (Fill in all fields, then change all editable fields.)
+    // Goal: save one project.
     ProjectPage page = getProjectPage(null);
     assertNotNull(page);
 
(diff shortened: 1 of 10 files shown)


miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/ProjectPage.java
```diff
@@ -2,36 +2,118 @@
...
+    return getSelectedDropdownText(referenceGenomeLabel);
+  }
+
+  public void setReferenceGenome(String referenceGenome) {
+    setDropdown(referenceGenome, referenceGenomeLabel);
+  }
+
+  public void clickSave() {
+    saveButton.click();
+    waitWithTimeout().until(titleContains("Project "));
```
djcooke: This won't actually do anything since the title still matches before the page is reloaded. It seems that since the page can take a couple seconds to actually start reloading, the current `waitForPageRefresh` method is similarly doing nothing though - `document.readyState` is `complete` already, and we don't know that it's ever been in any other state.

Should probably at least change this to `waitForPageRefresh()` though. Then it can be fixed in one place at least.

This method (and any that cause page reload) should return a new Page object since the page may have changed significantly or failed to load, and there may be stale elements, new elements, etc.

miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/ProjectPageIT.java
```diff
@@ -0,0 +1,105 @@
...
+  public void setup() {
+    loginAdmin();
+  }
+
+  private ProjectPage getProjectPage(Long projectId) {
+    return ProjectPage.get(getDriver(), getBaseUrl(), projectId);
+  }
+
+  @Test
+  public void testSaveEditNewProject() {
```
djcooke: Saving and editing should be two separate tests

miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/AbstractPage.java
```diff
@@ -31,7 +32,13 @@ protected static void waitExplicitly(long millis) {
...
   }
 
-  protected void waitForPageRefresh() {
+  /**
+   * Assesses page refresh by checking the staleness state of the HTML tag of the current page.
+   * 
+   * @param html the HTML element of the current page
+   */
+  protected void waitForPageRefresh(WebElement html) {
+    waitUntil(ExpectedConditions.stalenessOf(html));
```
djcooke: Nice

PR #1038, opened 2017-08-04
fix selecting locked HOT cells

djcooke:
Also works around Chromedriver bug entering digits by avoiding character '2' (this and other characters may still cause issue if not avoided)


callunity:
Should we be concerned about the failed test? `IllegalStateException: Cell still not selected after clicking`. I'm assuming not, given that the other Travis passed, but this exception is new to me.

PR #1046, opened 2017-08-09
Add yield bar graph

apmasell:
JIRA: GLT-1754


apmasell:
I've capped all the tooltip decimals to 2 places.

PR #1051, opened 2017-08-11
Force metrics table to span two columns

apmasell:


miso-web/src/main/webapp/scripts/run_graph.js
```diff
@@ -33,6 +33,7 @@ var RunGraph = (function() {
       var node = document.createElement('P');
       return {
         dom : node,
+        span : false, // Span the whole row?
```
callunity: The comment is ambiguous; perhaps something like `Indicator of whether the graph spans the entire width of the row.`

PR #1063, opened 2017-08-16
GLT-1964: Bulk pool integration tests

djcooke:


miso-web/src/main/webapp/scripts/hot_pool.js
```diff
@@ -9,28 +9,41 @@ HotTarget.pool = {
...
+          header : 'Pool Name',
+          data : 'name',
+          readOnly : true,
+          include : true,
+          unpackAfterSave : true,
+          unpack : function(pool, flat, setCellMeta) {
+            flat.name = pool.name;
+          },
+          pack : function(pool, flat, errorHandler) {
+            pool.name = flat.name;
```
apmasell: In the other types, `pack` does nothing. This should be the same.

djcooke: I noticed that, but is there a reason behind it? I assumed it was that it wouldn't have any actual effect. The service layer is responsible for that decision though.

We often have bugs due to field values getting lost somewhere in DB -> Hibernate -> DTO -> HOT -> DTO -> Hibernate -> DB transformations. Without good reason to exclude it, including the field everywhere seems preferable to me.

I can still remove it if you'd prefer. Just curious as to why?

apmasell: If the name is going to not be written back to the DB, then it should be dropped at every step. If it's copied, then it mysteriously vanishes at the next step, that's confusing. If it's dropped at every transfer point, then it's clear it wasn't meant to get saved.

djcooke: The confusion of fields mysteriously vanishing at a particular step is exactly my point. This is the only point at which the name disappears for some entities. It doesn't have any actual effect though, so as your preference I will match the existing pattern

PR #1066, opened 2017-08-17
GLT-2002: Library page integration tests

djcooke:


miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/FormPage.java
```diff
@@ -0,0 +1,28 @@
...
+ * 
+ * @param <T> field enum
+ */
+public abstract class FormPage<T> extends HeaderFooterPage {
+
+  public FormPage(WebDriver driver) {
+    super(driver);
+  }
+
+  public abstract boolean isEditable(T field);
```
apmasell: You can make this simpler. Erase the generic, then add this interface:
```
public interface FieldElement {
public By getSelect();
public FieldType getType();
}
```
And make the enums extends it.
`public void setFields(Map<T, String> fields)`  can be changed to `public void setFields(Map<? extends FieldValue, String> fields)`

Then you could push all the logic you have in the derived classes and enums into this base class.

miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/util/FormPageTestUtils.java
```diff
@@ -0,0 +1,36 @@
...
+
+import uk.ac.bbsrc.tgac.miso.webapp.integrationtest.page.FormPage;
+
+public class FormPageTestUtils {
+
+  private FormPageTestUtils() {
+    throw new IllegalStateException("Util class not intended for instantiation");
+  }
+
+  public static <T> void assertFieldValues(String hintMessage, Map<T, String> fields, FormPage<T> page) {
```
apmasell: Same here. If you make the above changes, you can erase the generic argument.

miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/FormPage.java
djcooke: I hadn't done this initially because it forces the enum methods to be public and then they're accessible to the test classes which should not directly access WebElements. It is only test code though, and this change does simplify writing tests, which should probably be the priority, so :+1:

miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/util/FormPageTestUtils.java
djcooke: Would require removing the type parameter from `FormPage` as well, and that would mean you could use the wrong FieldElements on a page. Does not simplify test-writing enough to justify this (only saves you from entering a type parameter to extending `FormPage`)

PR #1078, opened 2017-08-24
GLT-2002: integration tests for notes on Library page

djcooke:


callunity:
Only one of the Travises failed, and I can't replicate it locally -- is there a race condition somewhere?

PR #1085, opened 2017-08-25
Avoid using sampleoptions for institute defaults

apmasell:


miso-web/src/main/webapp/scripts/instituteDefaults_ajax.js
```diff
@@ -58,8 +49,7 @@ var Tissue = Tissue || {
     var id, alias, endpoint, temp, rowId;
     
     // if data is coming in from AJAX request, store it in Defaults.all
-    if (xhr) Defaults.all.tissueMaterialsDtos = JSON.parse(xhr.responseText);
-    Tissue.createTable(Defaults.all.tissueMaterialsDtos, 'TM', 'allMaterials', 'Material', TMtable);
+    Tissue.createTable(JSON.parse(xhr.responseText), 'TM', 'allMaterials', 'Material', TMtable);
```
callunity: I think you need to add some more `createTable` calls here, as the only table on the page is for Tissue Material.

apmasell: I see tissue material, sample purpose, subprojects, institutes, and labs...

callunity: I'm not seeing institutes, subprojects, or labs, all failing with the same error:
```
Uncaught TypeError: Cannot set property 'subprojectsDtos' of undefined
    at createSubprojectsTable (header_script.js?version=0.2.86-SNAPSHOT:13169)
    at XMLHttpRequest.xhr.onreadystatechange (header_script.js?version=0.2.86-SNAPSHOT:13509)
```
And it's failing because `Defaults.all` is undefined.

callunity: (In case it's relevant: I'm deploying this PR as-is; it's not rebased on the QC refactoring branch.)

PR #1123, opened 2017-09-13
Show low quality libraries in pool picker

apmasell:


miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/PoolDto.java
```diff
@@ -208,4 +209,12 @@ public int getLongestIndex() {
   public void setLongestIndex(int longestIndex) {
     this.longestIndex = longestIndex;
   }
+
+  public boolean isHasLowQualityLibraries() {
```
djcooke: isHas?

apmasell: It's what Eclipse generated. :-( This bean naming convention makes me sad.

PR #1129, opened 2017-09-15
GLT-2020: revised OICR PacBio library naming scheme

djcooke:


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateSiblingNumberGenerator.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @Override
+  public <T extends Aliasable> int getNextSiblingNumber(Class<T> clazz, String partialAlias) throws IOException {
+    // Find highest existing siblingNumber matching this partialAlias
+    Criteria criteria = currentSession().createCriteria(clazz);
+    criteria.add(Restrictions.like("alias", partialAlias, MatchMode.START));
+    @SuppressWarnings("unchecked")
+    List<Aliasable> aliasables = criteria.list();
+    String regex = "^.{" + partialAlias.length() + "}(\\d*)$";
```
apmasell: You don't have to change this, but I think it's would be clearer as a stream:
```
criteria.list().stream().map(Aliasable::getAlias).map(pattern::matcher).filter(Matcher::matches).mapToInt(m -> Integer.parseInt(m.group(1))).max() + 1
```

PR #1134, opened 2017-09-15
Don't allow changing the platform type of a library through bulk edit after creation

apmasell:
This matches the single edit page


apmasell:
The title is wrong. I'll fix the commit as well. It's modifiable on the edit single page if an admin.

PR #1144, opened 2017-09-20
Add dilution counts to pools

apmasell:
JIRA: GLT-2043

### DIFF
--- a/miso-web/src/main/webapp/scripts/list_partition.js
+++ b/miso-web/src/main/webapp/scripts/list_partition.js
@@ -279,6 +279,23 @@ ListTarget.partition = {
 
           }
         }, {
+          "sTitle": "Dilutions",
+          "mData": "pool",
+          "include": true,
+          "iSortPriority": 0,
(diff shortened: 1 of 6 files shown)


callunity:
Would you consider dropping the "Created Date" column since we have a "Last Modified" column?

miso-web/src/main/webapp/scripts/list_partition.js
```diff
@@ -279,6 +279,22 @@ ListTarget.partition = {
...
+            if (data) {
+              return data.dilutionCount;
+            } else {
+              if (type === 'display') {
+                return "(None)";
+              } else {
+                return "";
+              }
+            }
+          }
```
callunity: `bSortable: false` for both tables

apmasell: Fixed.

apmasell:
I have no strong feelings on the creation date. It's there for hysterical raisins.

PR #1153, opened 2017-09-26
Add Pinery queries as stored procedures

callunity:


miso-web/src/it/resources/db/scripts/schema_translator.groovy
```diff
@@ -18,6 +18,7 @@ println('Translating schema files from ' + productionSchemaDir.getAbsolutePath()
 
 final String productionScriptPattern = '^(V\\d{4}_.*|afterMigrate|beforeMigrate)\\.sql$'
 final String testSchemaDir = basedir + '/target/test-classes/db/migration/'
+final String testStoredProceduresFile = basedir + '/src/it/resources/db/migration/afterAfterMigrate.sql'
```
djcooke: Can't you just put this file in `migration_afterMigrate` and name it `99_test_procedures.sql` or something similar to have it run after all the other afterMigrate scripts, which are all numbered `05`?

sqlstore/src/main/resources/db/migration_afterMigrate/10_pinery_queryInstruments.sql
```diff
@@ -0,0 +1,15 @@
+--StartNoTest
+DELIMITER //
+
+DROP PROCEDURE IF EXISTS queryAllInstruments//
+CREATE PROCEDURE queryAllInstruments() BEGIN
+  PREPARE stmt FROM 'SELECT sr.referenceId
```
djcooke: Prepared statements work to prevent SQL injection, but is there any purpose to using them within a stored procedure for queries that have no parameters and are only executed once (per procedure call)?

callunity: It's intended as a developer-friendliness thing -- using prepared statements for both the parameterized as well as paramater-less queries means there are fewer lines in a diff of the two files.

miso-web/src/it/resources/db/scripts/schema_translator.groovy
callunity: Unfortunately, we can't put it into `migration_afterMigrate` until we convert the unit tests to MySQL. Straight up moving it there causes the tests to fail.

djcooke: `-- StartNoTest`?

callunity: :bulb: ...it works!

PR #1159, opened 2017-09-29
Visually indicate multiple options are available in HoT

callunity:
GLT-2059


apmasell:
Is there matching CSS?

callunity:
...there is now.

callunity:
The feedback I got from the lab is that the colour is sufficiently noticeable; we can certainly adjust it if they find that differs once it's in production. I've added a [comment](https://jira.oicr.on.ca/browse/GLT-2057) on the ticket about adding a Quick Help section that the quick help for Identity Alias should note that coloured cells indicate that more than one possible identity are available in the dropdown menu.

PR #1161, opened 2017-09-29
Integration tests for run page

callunity:
GLT-1969

### DIFF
--- a/miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/RunPage.java
+++ b/miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/RunPage.java
@@ -105,6 +105,10 @@ public class RunPage extends FormPage<RunPage.Field> {
     return new RunPage(getDriver());
   }
 
+  public DataTable getTable(String tableId) {
+    return new DataTable(getDriver().findElement(By.id(tableId)));
+  }
+
   public RunPage addContainer(String serialNumber, String platformType, boolean isRunFull) {
     WebElement html = getHtmlElement();
     containersSection.findElement(By.linkText("Add " + PlatformType.get(platformType).getContainerName())).click();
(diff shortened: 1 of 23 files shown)


miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/RunPage.java
```diff
@@ -0,0 +1,227 @@
...
+    partitions.forEach(partition -> partitionsTable.checkBoxForRow(partition));
+    partitionsSection.findElement(By.linkText(titleText)).click();
+    waitUntil(visibilityOf(dialog));
+    waitUntil(textToBe(By.className("ui-dialog-title"), titleText));
+    dialog.findElement(By.linkText(option)).click();
+  }
+
+  private void clickFirstPoolTile(boolean extraWait, String searchText) {
+    if (extraWait) {
+      waitUntil(textToBe(By.className("ui-dialog-title"), "Select Pool"), 60);
```
djcooke: Wow, searching by "ready to run" is pretty slow for some reason...

miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/element/DataTable.java
```diff
@@ -77,6 +79,14 @@ public String getTextAtCell(String columnHeading, int rowNum) {
     return cell.getText();
   }
 
+  public void checkBoxForRow(int rowNum) {
+    List<WebElement> checkbox = getCell(Columns.SORT, rowNum).findElements(By.tagName("input"));
+    if (checkbox.isEmpty()) {
+      throw new IllegalArgumentException("Row " + rowNum + " does not have a checkbox to click.");
+    }
+    checkbox.get(0).click();
```
djcooke: Use `findElementIfExists` and null check instead. Includes check for unexpected multiple results

callunity: In this case, the "find" needs to be only on the results of `getCell`. `findElementIfExists` will look over an entire page, and tables with more than one row will cause it to throw an error.

miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/page/RunPage.java
callunity: Yeah, I'm not sure why that is.

PR #1163, opened 2017-10-03
QC migration patches

froggleston:
Fix a few QC issues when migrating.

### DIFF
--- a/sqlstore/src/main/resources/db/migration/V0320__auto_main.sql
+++ b/sqlstore/src/main/resources/db/migration/V0320__auto_main.sql
@@ -80,7 +80,7 @@ CREATE TABLE PlatformSizes (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 INSERT INTO PlatformSizes(platform_platformId, partitionSize)
-  SELECT DISTINCT platform, COUNT(*) AS c FROM SequencerPartitionContainer JOIN SequencerPartitionContainer_Partition ON containerId = container_containerId GROUP BY containerId
+  SELECT platform, c FROM (SELECT DISTINCT platform, COUNT(*) AS c FROM SequencerPartitionContainer JOIN SequencerPartitionContainer_Partition ON containerId = container_containerId GROUP BY containerId) as p where platform > 0
   UNION SELECT platformId, 1 FROM Platform WHERE name = 'ILLUMINA' AND instrumentModel LIKE '%MiSeq%'
   UNION SELECT platformId, 2 FROM Platform WHERE name = 'ILLUMINA' AND instrumentModel LIKE '%HiSeq%'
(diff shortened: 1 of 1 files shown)


sqlstore/src/main/resources/db/migration/V0320__auto_main.sql
```diff
@@ -45,11 +44,11 @@ CREATE TABLE `PoolQC2` (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 INSERT INTO LibraryQC2(`qcId`, `library_libraryId`, `creator`, `date`, `type`, `results`)
-	SELECT qcId, library_libraryId, (SELECT userId FROM `User` WHERE fullName = qcCreator), qcDate, qcMethod, results FROM LibraryQC;
+	SELECT qcId, library_libraryId, IFNULL((SELECT userId FROM `User` WHERE fullName = qcCreator OR loginName = qcCreator),1), qcDate, qcMethod, results FROM LibraryQC;
```
apmasell: 1. I don't think a substitute user is appropriate. If your database has bad data, it should fail to migrate.
2. Why user 1? It's not even guaranteed to be admin. I believe it isn't admin in your database.

sqlstore/src/main/resources/db/migration/V0320__auto_main.sql
```diff
@@ -81,7 +80,7 @@ CREATE TABLE PlatformSizes (
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 INSERT INTO PlatformSizes(platform_platformId, partitionSize)
-  SELECT DISTINCT platform, COUNT(*) AS c FROM SequencerPartitionContainer JOIN SequencerPartitionContainer_Partition ON containerId = container_containerId GROUP BY containerId
+  SELECT platform, c FROM (SELECT DISTINCT platform, COUNT(*) AS c FROM SequencerPartitionContainer JOIN SequencerPartitionContainer_Partition ON containerId = container_containerId GROUP BY containerId) as p where platform > 0
```
apmasell: If the platform is 0, the database is wrong and should be fixed. How is this possible? You have containers with no associated platform? Are they attached to a sequencer?

sqlstore/src/main/resources/db/migration/V0320__auto_main.sql
froggleston: I'm unsure how to represent this in this instance. We can't delete the QCs. We could map them to an existing user in our DB, but then we'll lose the real data. The only solution we have if we can't map it to an existing user is to create a new user for each external QC creator. I'll speak to the team here.

sqlstore/src/main/resources/db/migration/V0320__auto_main.sql
froggleston: I've checked this and these containers can be deleted as they're duplicates.

froggleston:
This can be closed now. I've worked through the local DB issues so that the develop migrations now work properly.

PR #1176, opened 2017-10-17
GLT-2070: updated Pinery queries

djcooke:
to match Pinery update: https://github.com/oicr-gsi/pinery/pull/65


djcooke:
Will be deleting the stored procedures in another PR soon, so not merging this

PR #1177, opened 2017-10-19
Add default BoxUse and BoxSize

callunity:
And call Pinery stored procedures only during tests, not during regular migration.

### DIFF
--- a/sqlstore/src/main/resources/db/migration_afterMigrate/05_trigger_sampleqc.sql
+++ b/sqlstore/src/main/resources/db/migration_afterMigrate/05_trigger_sampleqc.sql
@@ -4,11 +4,12 @@ DELIMITER //
 DROP TRIGGER IF EXISTS SampleQCInsert//
 CREATE TRIGGER SampleQCInsert AFTER INSERT ON SampleQC
 FOR EACH ROW
-  INSERT INTO SampleChangeLog(sampleId, columnsChanged, userId, message) VALUES (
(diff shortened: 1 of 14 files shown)


sqlstore/src/main/resources/db/migration_afterMigrate/05_trigger_sampleqc.sql
```diff
@@ -20,11 +21,12 @@ FOR EACH ROW
...
         THEN CONCAT('Updated ', (SELECT name FROM QCType WHERE qcTypeId = NEW.type), ' QC: ', OLD.results, ' → ', NEW.results, (SELECT units FROM QCType WHERE qcTypeId = NEW.type)) END);
       IF log_message IS NOT NULL AND log_message <> '' THEN
-        INSERT INTO SampleChangeLog(sampleId, columnsChanged, userId, message) VALUES (
+        INSERT INTO SampleChangeLog(sampleId, columnsChanged, userId, message, changeTime) VALUES (
           NEW.sample_sampleId,
           'QC',
           (SELECT lastModifier FROM Sample WHERE sampleId = NEW.sample_sampleId),
-          log_message);
+          log_message,
+          (SELECT lastModified FROM Sample WHERE sampleId = NEW.sample_sampleId));
```
djcooke: Could use INSERT INTO -> SELECT syntax to reduce subqueries on all these similar statements instead of finding `lastModifier` and `lastModified` separately

PR #1186, opened 2017-10-25
UI tests for bulk propagating samples

callunity:
GLT-1956


callunity:
I'm leaving this open until we figure out how to run the UI tests in parallel Travis jobs; addition additional jobs causes Travis to time out about half the time.

PR #1200, opened 2017-11-13
Fix container barcode validation

apmasell:


apmasell:
Agreed, but it seems that the save actually succeeded in some cases, so it's not purely a test failure.

PR #1208, opened 2017-11-21
Automatically choose Add Container options when there's no choice

apmasell:


callunity:
You'll need to update the failing test; it's still expecting to have to select the "1 Lane" option.

PR #1267, opened 2018-01-04
GLT-2123: require unique pool alias

djcooke:


callunity:
Do pools already have an "alias validation" step before save? If not, could one please be added to give users nicer errors?

PR #1273, opened 2018-01-05
GLT-2144: fixed missing sample relatives on Edit Sample page

djcooke:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
```diff
@@ -650,20 +656,8 @@ public ModelAndView setupForm(@PathVariable Long sampleId, @PathVariable Long pr
...
-          Project project = projectService.getProjectById(projectId);
-          if (project == null) throw new SecurityException("No such project.");
-          model.addAttribute("project", project);
-          sample.setProject(project);
-          sample.inheritPermissions(project);
-
-          Map<String, Sample> adjacentSamples = getAdjacentSamplesInProject(sample, sample.getProject().getProjectId());
-          if (!adjacentSamples.isEmpty()) {
-            model.put("previousSample", adjacentSamples.get("previousSample"));
-            model.put("nextSample", adjacentSamples.get("nextSample"));
```
callunity: Were "previousSample" and "nextSample" intended to be removed here? After we removed them previously, Chris re-added them in #550 at the lab's request so I'm hesitant to remove them here without more information.

djcooke: These were only getting added for the url `/sample/{sampleId}/project/{projectId}`, which was not linked anywhere, and also doesn't make sense since you could request a sample that belongs in PRO1, but with the ID for PRO2, it would distort the view so that the sample appeared to be in the wrong project.

I'll fix to use the sample's actual project and re-include them instead, which will fix #1275

PR #1288, opened 2018-01-16
GLT-1408: Add CUD logo to login page

ndrein-oicr:
Looks like this: 
image


miso-web/src/main/webapp/login.jsp
```diff
@@ -51,6 +51,7 @@
         </tr>
         <tr>
           <td align="left">
+            <img src="<c:url value='/styles/images/blue-cud-button.png'/>" alt="CUD logo"/>
             <!--<small><a href="registerUser">New user</a></small>-->
```
djcooke: Could remove this commented-out line while you're at it since it will never make sense to allow a user to register like this

ndrein-oicr:
CUD logo is only relevant to OICR (nice catch Dillan).  Moving to OICR fork.

PR #1302, opened 2018-01-30
Update documentation with create user syntax for newer MySQL

callunity:


docs/_posts/2016-01-11-admin-manual.md
```diff
@@ -75,7 +75,13 @@ Once installed, start the MySQL console and create the database:
...
     USE lims;
 
-Then add a user that has all grant access on the 'lims' db:
+Then add a user that has all grant access on the 'lims' db. For MySQL 5.7.7 or
+higher:
+
+    CREATE USER 'tgaclims'@'localhost' IDENTIFIED BY 'tgaclims';
+    GRANT ALL ON `lims`.* TO 'tgaclims'@'localhost';
+
+For earlier versions of MySQL:
```
djcooke: Can probably omit the old syntax as the new syntax should be fine in earlier versions of mysql too

callunity: Would you be able to test that out for me and confirm it?

djcooke: New syntax looks fine according to MySQL 5.6 manual:
https://dev.mysql.com/doc/refman/5.6/en/create-user.html
https://dev.mysql.com/doc/refman/5.6/en/grant.html

And we list MySQL 5.7 as a requirement anyway

djcooke: Also fairly certain I've never used the "old syntax"

callunity: Thanks. I had definitely used the old syntax before.

PR #1304, opened 2018-01-31
GLT-1825: Add ID column to list sequencing containers page

ndrein-oicr:
Also updated integration tests to look for this new column.

### DIFF
--- a/miso-web/src/main/webapp/scripts/list_container.js
+++ b/miso-web/src/main/webapp/scripts/list_container.js
@@ -111,23 +111,30 @@ ListTarget.container = {
     }, ];
   },
   createColumns: function(config, projectId) {
-    return [ListUtils.labelHyperlinkColumn("Serial Number", "container", Utils.array.getId, "identificationBarcode", 1, true), {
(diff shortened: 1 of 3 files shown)


miso-web/src/main/webapp/scripts/list_container.js
```diff
@@ -111,23 +111,30 @@ ListTarget.container = {
...
+        return container.lastRunId;
+      }, "lastRunAlias", -1, true),
+      ListUtils.labelHyperlinkColumn("Last Sequencer Used", "sequencer", function(container) {
+        return container.lastSequencerId;
+      }, "lastSequencerName", -1, true),
+      {
+        "sTitle": "Last Modified",
+        "mData": "lastModified",
+        "include": true,
+        "iSortPriority": 0
```
callunity: It's not in the spec, but could you please also adjust the `iSortPriority` of the Last Modified item to be `2`? This will make the table sort by last modified date.

ndrein-oicr: :+1:

PR #1319, opened 2018-02-07
GLT-2108: Add Lane QC changes to RunChangeLog

ndrein-oicr:
When the Run_Partition_QC table is changed, compose some fields and store them in temporary user variables.  Store the changed records in both `SequencerPartitionContainerChangeLog` and `RunChangeLog`.

I auto-formatted the file for readability, apologies in advance.


djcooke:
Can further reduce the number of queries by getting `containerId`, `userId`, and `partitionNumber` from a single query

PR #1320, opened 2018-02-07
GLT-2173: added tool to check index edit distance

djcooke:
This is for checking indices before library prep. Pool-level warnings will be added in GLT-2101


miso-web/src/main/webapp/scripts/indexDistance_ajax.js
```diff
@@ -0,0 +1,70 @@
+(function(IndexDistance, $, undefined) { // NOSONAR (paranoid assurance that undefined is undefined)
```
apmasell: LOL

PR #1326, opened 2018-02-08
GLT-2101: Display warnings for near-duplicate indices in a pool

djcooke:


djcooke:
@callunity: Yes. Duplicate index warnings were already shown on the run page too, which I hadn't noticed. Fixed now

PR #1429, opened 2018-05-02
Add total cluster count to summary table

apmasell:
JIRA: GLT-2298


apmasell:
No. It has to be put into production in Run Scanner and let it update everything to be seen. :-1:

PR #1445, opened 2018-05-14
Change reads/cluster terminology in Illumina summary

apmasell:
The changes read to clusters, remove this previous ill-conceived cluster number
and adds a reads/cluster PF number.


apmasell:
For release notes:

* Changed Illumina summary. Removed “cluster” entry that was not calculated correctly. Renamed “total reads” to “clusters”, which is the sum cluster count across all lanes. Added “clusters PF” for sum clusters passing filter across all lanes.

PR #1534, opened 2018-07-11
Make sample group ID more visible on library (DetailedSample)

callunity:
GLT-2042


callunity:
Closing for further work as per a conversation with @djcooke

PR #1543, opened 2018-07-16
Changelog the detailedLibrary fields

callunity:
GLT-2399


sqlstore/src/main/resources/db/migration_afterMigrate/05_trigger_library.sql
```diff
@@ -37,10 +39,11 @@ FOR EACH ROW
         CASE WHEN (NEW.libraryStrategyType IS NULL) <> (OLD.libraryStrategyType IS NULL) OR NEW.libraryStrategyType <> OLD.libraryStrategyType THEN 'libraryStrategyType' END,
         CASE WHEN (NEW.libraryType IS NULL) <> (OLD.libraryType IS NULL) OR NEW.libraryType <> OLD.libraryType THEN 'libraryType' END,
         CASE WHEN (NEW.locationBarcode IS NULL) <> (OLD.locationBarcode IS NULL) OR NEW.locationBarcode <> OLD.locationBarcode THEN 'locationBarcode' END,
+        CASE WHEN NEW.lowQuality <> OLD.lowQuality THEN 'lowQuality' END,
         CASE WHEN NEW.paired <> OLD.paired THEN 'paired' END,
         CASE WHEN (NEW.platformType IS NULL) <> (OLD.platformType IS NULL) OR NEW.platformType <> OLD.platformType THEN 'platformType' END,
         CASE WHEN (NEW.qcPassed IS NULL) <> (OLD.qcPassed IS NULL) OR NEW.qcPassed <> OLD.qcPassed THEN 'qcPassed' END,
-        CASE WHEN NEW.sample_sampleId <> OLD.sample_sampleId THEN 'sample_sampleId' END,
```
Kashif-S: Could you explain why this is removed?

callunity: I had meant to also remove the corresponding sample_sampleId message because there's no way to reparent libraries in the UI and we don't do that for the lab either as part of our workflows. On further reflection I suppose other MISO admins may choose to reparent libraries and rely on changelogs to indicate that (in the same way as we sometimes reassign samples to different projects) so I'll re-add this.

PR #1565, opened 2018-07-27
Add Description To Bulk Pools Table

Kashif-S:
Jira: GLT-2417


callunity:
You'll want to update the `BulkPoolIT` commonColumns with the description column

PR #1584, opened 2018-08-07
Refactor Search ITs into UTs

Kashif-S:


djcooke:
Bad rebase?

Kashif-S:
Fixed now.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateLibraryDilutionDao.java
```diff
@@ -175,7 +175,9 @@ public String propertyForDate(Criteria item, DateType type) {
 
   @Override
   public String propertyForUserName(Criteria item, boolean creator) {
-    return creator ? "dilutionUserName" : "derivedInfo.lastModifier.loginName";
+    item.createAlias("creator", "creator");
+    item.createAlias("lastModifier", "lastModifier");
```
Kashif-S: I'm a little unsure if I put these `createAlias` calls in the right spot. The tests produce errors if this isn't here, but none of the other Daos, which use the same return statement as this function, use `createAlias` here. If there's a better spot to put these `createAlias` calls, please let me know.

djcooke: Add "creator" and "lastModifier" to the `STANDARD_ALIASES` list at the top of this class. These get added to all the criteria used in PaginatedDataSource

PR #1607, opened 2018-08-15
Fixed create container QC error

djcooke:


djcooke:
Will run the modified script in production after approved

PR #1610, opened 2018-08-15
Fix Travis ITs

djcooke:


djcooke:
Closing for now because the fix doesn't work... Tests just randomly passed the first couple times I tried

PR #1614, opened 2018-08-20
Start Workflows from Search Widget

Kashif-S:
Jira: GLT-2267

### DIFF
--- a/miso-web/src/main/webapp/scripts/pane.js
+++ b/miso-web/src/main/webapp/scripts/pane.js
@@ -173,6 +173,7 @@ var Pane = (function() {
         content: contentDiv
       };
     },
+    
     createPane: function(paneId, title) {
       var titleDiv = createTitleDiv(title);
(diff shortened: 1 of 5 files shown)


miso-web/src/main/webapp/scripts/pane.js
```diff
@@ -186,18 +187,29 @@ var Pane = (function() {
...
-        showError(xhr, textStatus, errorThrown, div);
-      }, url, query);
+    updateTiles: function(div, transform, url, query, add, checkWorkflow) {
+      var workflow = null;
+      if(checkWorkflow){
+        workflow = Constants.workflows.find(function(wkflow){
+          return wkflow.barcode == query;
+        });
+        if(workflow){
+          window.location = window.location.origin + '/miso/workflow/new/' + workflow.workflowName;
```
apmasell: Put in a `return;` after this line and then remove the `if(!workflow)` below.

PR #1627, opened 2018-08-23
Fix Library Templates List Page

Kashif-S:
Jira: GLT-2308


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/ListLibraryTemplatesController.java
```diff
@@ -0,0 +1,54 @@
...
+
+import uk.ac.bbsrc.tgac.miso.webapp.util.ListItemsPage;
+
+/**
+ * uk.ac.bbsrc.tgac.miso.webapp.controller
+ * <p/>
+ * Info
+ * 
+ * @author Rob Davey
+ * @since 0.0.2
```
djcooke: Pretty sure these aren't true! Remove Javadoc

PR #1646, opened 2018-08-30
GLT-2460: specify Flyway version and describe DB URL

djcooke:


docs/_posts/2016-01-11-installation-guide.md
```diff
@@ -92,7 +92,8 @@ need to add a grant privilege to the MISO database from your remote machine:
 
 # Setting Up the Application Server
 
-Download the [Flyway command line tool](https://flywaydb.org/documentation/commandline/) and install it.
+Download the [Flyway command line tool](https://flywaydb.org/download/community) version 3.2.1 and install it.
```
avarsava: Later in the article we require Maven, and here we use the Flyway plugin for Maven, should we advise using that in the installation guide as well?

docs/_posts/2016-01-11-installation-guide.md
```diff
@@ -398,6 +399,12 @@ The same path should be used for `MISO_FILES_DIR` as is set for `miso.fileStorag
     unzip -xjo $CATALINA_HOME/webapps/ROOT.war 'WEB-INF/lib/sqlstore-*.jar' -d lib
     ./flyway -user=$MISO_DB_USER -password=$MISO_DB_PASS -url=$MISO_DB_URL -outOfOrder=true -locations=classpath:db/migration,classpath:uk.ac.bbsrc.tgac.miso.db.migration migrate -placeholders.filesDir=${MISO_FILES_DIR}
 
+`$DB_URL` should be in the same format as in the `ROOT.xml`, except replacing `&amp;` with just `&`:
+
+```
+jdbc:mysql://localhost:3306/lims?autoReconnect=true&zeroDateTimeBehavior=convertToNull&useUnicode=true&characterEncoding=UTF-8
```
avarsava: Should we advise to strip the options off the URL? Mine ends at 'lims' and works just fine

apmasell: We've had all kinds of weird Unicode problems without the encoding.

docs/_posts/2016-01-11-installation-guide.md
djcooke: We use the command-line (v3.2.1) tool when releasing to our staging and production servers to avoid needing the source code on them. I think recommending the same here is ideal in case users want to build and deploy on different machines

docs/_posts/2016-01-11-installation-guide.md
djcooke: We've been using these options in production and testing for a long time. They may or may not be necessary, but we know they're not harmful, so I'd rather keep them for now.

djcooke:
@avarsava: Fixed `runscanner-*.war` path. `runscanner/target/` is the normal location

PR #1654, opened 2018-09-07
GLT-2481: removed timestamp 'ON UPDATE' values

djcooke:


djcooke:
Also dropped `Status` table which still exists for some reason

PR #1664, opened 2018-09-14
[GLT-2414] Add line graph with custom x axis floor

avarsava:
Add lineGraph_CustomX() which draws a line graph with a custom starting
point ("floor") on x axis.
Convert lineGraph() to calling lineGraph_CustomX() with '0' as floor.
Changes 'Lane' graph to drawing lineGraph_CustomX with floor of '1', as
per ticket.

New function allows changing floor of any other line graphs in the future.


djcooke:
Please make sure ticket number is included in the commit message

PR #1674, opened 2018-09-26
GLT-2515: added Site Config section of new user manual

djcooke:


docs/user_manual/site_configuration.md
```diff
@@ -0,0 +1,191 @@
...
+
+
+
+{% assign sub = sub | plus: 1 %}
+{% include userman-heading.md section=page.section sub=sub title="Report Links" %}
+
+{% assign figure = figure | plus: 1 %}
+If you are generating reports for projects or runs outside of MISO, it may be possible to automatically link to them
+from the MISO Project and Run pages. For projects, it must be possible to generate the report's URL given the project's
+ID, name, and short name. For runs, it must be possible to generate the report's URL given the run's ID, name, and
```
callunity: Project: `ID, name, or short name`
Run: `ID, name, or alias`

docs/user_manual/site_configuration.md
```diff
@@ -0,0 +1,191 @@
...
+{% assign sub = sub | plus: 1 %}
+{% include userman-heading.md section=page.section sub=sub title="Naming Schemes" %}
+
+Some institutes will have a specific and uniform way of naming things. MISO allows you to customize the validation of
+some fields in order to enforce these naming schemes:
+
+* All item names
+* Sample alias
+* Library alias
+* Project ShortName
```
callunity: `Project short name` ? (At least, that's the casing we usually use)

docs/user_manual/site_configuration.md
```diff
@@ -0,0 +1,191 @@
...
+
+{% assign sub = sub | plus: 1 %}
+{% include userman-heading.md section=page.section sub=sub title="Default Values" %}
+
+It is possible to set a default value for some fields to save users from repetitive entry. If a default value is
+configured, it will be filled in automatically when creating new items. The user is still free to change the value if
+they would like.
+
+The only field that can currently be configured with a default value is sample scientific name. This setting only
+affects the bulk entry form.
```
callunity: It also affects the single entry form ( `/miso/sample/new/157` or any valid project number).

docs/user_manual/site_configuration.md
callunity: Then maybe `and/or`? `and` alone made it sound to my ears like you might need to have all of them in the report's URL. But maybe that's just me?

docs/user_manual/site_configuration.md
callunity: Hmm, that is indeed very specific...

docs/user_manual/site_configuration.md
apmasell: and/or

djcooke: Changed the wording to "using x, y, and/or z" and added examples

PR #1687, opened 2018-10-03
GLT-2518: added type data section of user manual

djcooke:


docs/user_manual/type_data.md
```diff
@@ -0,0 +1,514 @@
...
+[Detailed Sample Mode](site_configuration.html#detailed_sample_mode) for more information about the sample categories
+and ghost samples.
+
+{% assign figure = 1 %}
+Sample classes, and the relationships between them, can be customized to suit your requirements. Figure {{figure}}
+shows the sample hierarchy used at OICR.
+
+{% include userman-figure.md num=figure cap="OICR sample hierarchy" img="type-data-sample-hierarchy.png" %}
+
+Sample classes and valid relationships can only be modified via direct access to the MISO database, and so must be done
```
callunity: Can you explain "valid relationships" a bit further here? I'd particularly want to see a bit about how they're used to define a custom hierarchy, and a definition of "parenting" as it's used a lot in the bullet points below.

docs/user_manual/type_data.md
```diff
@@ -0,0 +1,514 @@
...
+{% include userman-toplink.md %}
+
+
+
+{% assign sub = sub | plus: 1 %}
+{% include userman-heading.md section=page.section sub=sub title="Tissue Types" %}
+
+Note: This item only applies if your site uses [detailed sample mode](site_configuration.html#detailed_sample_mode).
+
+Tissue Type is a classifier for tissue samples. It may be used to identify the use of the sample. For example, tissue
```
callunity: `the use of the sample`: `use` feels like not quite the right term to me. Maybe `It may be used to categorize or describe the sample` ? (I'm not fully sold on either of these, and this is more a point of discussion than a request for change -- if it makes sense to everyone else then keep it!)

docs/user_manual/type_data.md
```diff
@@ -0,0 +1,514 @@
...
+institute. Any user may create new labs and institutes in MISO. Administrators may also edit and delete existing labs
+and institutes.
+
+Available institutes can be viewed on the Institutes list page. To get there, click "Institutes" in the Institute
+Defaults list near the bottom of the menu on the left side of the screen. All operations affecting institutes will be
+started from this page.
+
+Available labs can be viewed on the Institutes list page. To get there, click "Labs" in the Institute Defaults list
+near the bottom of the menu on the left side of the screen. All operations affecting institutes will be started from
+this page.
```
callunity: `Available labs can be viewed on the Labs list page`, and `All operations affecting labs will be started from this page.`

docs/user_manual/type_data.md
```diff
@@ -0,0 +1,514 @@
...
+"Save" button at the top right to confirm.
+
+To edit an existing lab or institute, go to the appropriate list page, check the checkboxes next to the items you wish
+to modify, and clock the "Edit" button in the toolbar at the top of the table. This will take you to the corresponding
+bulk Edit page, which is identical to the bulk Create page. Make the changes you would like, then click the "Save"
+button at the top right to confirm. Keep in mind that any changes will affect all samples that are already using the
+affected lab or institute.
+
+To delete a lab or institute, go to the appropriate list page, check the checkboxes next to the items you wish
+to modify, and clock the "Delete" button in the toolbar at the top of the table. You will only be allowed to delete a
```
callunity: `clock` -> `click`

docs/user_manual/type_data.md
djcooke: "Categorize" works, though it's just as vague as "type." It could be said that every sample field "describes" the sample, so that's getting even more vague. A lot of these terms overlap and it's difficult to describe specifically what each is meant to represent. I don't think "the use of the sample" is a great description either. I was hoping the examples were helpful enough. Open to other suggestions

djcooke: Changing to "may be used to categorize the sample" for now

docs/user_manual/type_data.md
```diff
@@ -0,0 +1,514 @@
...
+MISO are:
+
+* Library kit - used for libraries
+* Sequencing kit - not currently used in MISO
+* Clustering kit - used for sequencing containers
+* Multiplexing kit - used for sequencing containers
+* Extraction kit - not currently used in MISO
+
+Modifications to kit descriptors can only be made via direct access to the MISO database. As such, they must be
+performed by a MISO adminstrator.
```
callunity: Users can add and edit kit descriptors (called "Kits" in the List page, and "Kit Descriptor" in the Edit page...) from the List Kits page.

docs/user_manual/type_data.md
```diff
@@ -0,0 +1,514 @@
...
+
+{% include userman-toplink.md %}
+
+
+
+{% assign sub = sub | plus: 1 %}
+{% include userman-heading.md section=page.section sub=sub title="Indices" %}
+
+Indices, also known as barcodes or primers, are sequences that are added to libraries in order to identify which
+library each individual DNA fragment came from in a multiplexed pool. An index family is a collection of these indices
```
callunity: Single-plexed pools can also have indices (and MISO warns when they don't).

docs/user_manual/type_data.md
```diff
@@ -0,0 +1,514 @@
...
+that are intended to be used together. Index sequences should all be unique within an index family.
+
+Some index families are dual-indexed. This means that two indices will be added to each library. This increases the
```
callunity: `This means that one or two indices can be added to each library`

docs/user_manual/type_data.md
```diff
@@ -0,0 +1,514 @@
...
+{% include userman-toplink.md %}
+
+
+
+{% assign sub = sub | plus: 1 %}
+{% include userman-heading.md section=page.section sub=sub title="Targeted Sequencing" %}
+
+Targeted sequencing allows specific sections of the genome to be sequenced resulting in more depth and lower cost. The
+targeted sequencing value specified on a library dilution should be associated with a .bed file that will be used for
+analysis and QC.
```
callunity: Maybe an indication that the .bed file isn't tracked in MISO?

docs/user_manual/type_data.md
djcooke: True... But do they really have a purpose in single-plexed pools? Is there reason to change the wording here?

docs/user_manual/type_data.md
callunity: It's true, we would! but the functionality exists in the GUI. I'm ok with leaving it like this though if the document's primary goal is to support our very-real users, with a secondary goal of supporting hypothetical external institute users.

djcooke: The goal is to document everything that MISO does and everything that can be done in MISO, so I'm updating to give details of how to add kit descriptors in MISO

docs/user_manual/type_data.md
apmasell: I think it's also worth explaining class vs category in detail here.

docs/user_manual/type_data.md
apmasell: Yeah, that's our convention. They could do it if they wanted, but they can't add targeted sequencing via the UI.

docs/user_manual/type_data.md
callunity: They can also be used to confirm no sample swaps occurred (confirming that the library which was sequenced has the same index as the library in the single-plex pool in MISO). I don't feel strongly about changing the wording.

docs/user_manual/type_data.md
djcooke: A previous paragraph contains "Sample classes identify where a sample fits into the hierarchy. Sample categories are used to group sample classes into levels," and the diagram helps to illustrate this. I added more description of sample classes and relationships. Are there any other details regarding categories that would be useful?

callunity: Could you explain what something being  "parented to a class" means?

djcooke: Do you mean beyond what I added already?

> Sample relationships have a parent and child class. This means that if you have a sample of the parent class, you can propagate a sample of the child class. In figure {{figure}}, the relationships are illustrated using arrows that point from the child class to the parent class.

docs/user_manual/type_data.md
```diff
@@ -47,8 +47,11 @@ Sample categories are used to group sample classes into levels. Sample categorie
 and ghost samples.
 
 {% assign figure = 1 %}
-Sample classes, and the relationships between them, can be customized to suit your requirements. Figure {{figure}}
-shows the sample hierarchy used at OICR.
+Sample classes, and the relationships between them define the sample hierarchy. Both the classes and relationships can
+be customized to suit your requirements. Figure {{figure}} shows the sample hierarchy used at OICR. Sample
+relationships have a parent and child class. This means that if you have a sample of the parent class, you can
+propagate a sample of the child class. In figure {{figure}}, the relationships are illustrated using arrows that point
+from the child class to the parent class.
```
callunity: `from the child class to the parent class` ? It looks to me like they point from the parent class to the child class...

In the diagram:
* `Single Cell DNA (sotck)` -> `(stock)`
* The arrows from Tissue -> LCM Tube and Slide -> LCM Tube are joined, which could be interpreted as "one tissue plus one slide = one LCM tube"...

docs/user_manual/type_data.md
callunity: Ugh sorry, I hadn't seen that. That looks good!

docs/user_manual/type_data.md
djcooke: Fixed. Also, I had removed whole RNA (aliquot) instead of the three subclasses. Fixed that too.

PR #1726, opened 2018-10-25
GLT-2601: fixed box scanning

djcooke:


integration-tools/src/main/java/uk/ac/bbsrc/tgac/miso/integration/visionmate/VisionMateScanner.java
```diff
@@ -26,7 +27,7 @@
   private static final int defaultCommunicationTimeout = 10000; // 10 sec
   private static final int maxCommunicationTimeout = 60000; // 1 min
   
-  private static final int defaultScanTimeout = 15000; // 15 sec
+  private static final int defaultScanTimeout = 30000; // 30 sec
```
apmasell: Please make this `30_000` for readability.

PR #1730, opened 2018-10-30
GLT-2485: show external names on Edit Sample and Library pages

djcooke:


callunity:
Approved once LibraryPageIT tests are fixed

PR #1733, opened 2018-10-31
Sort sequencing parameters

callunity:
GLT-2606 & GLT-1470

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
@@ -1,8 +1,10 @@
 package uk.ac.bbsrc.tgac.miso.core.data;
 
 import java.io.Serializable;
-import java.util.Comparator;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
+import java.util.List;
 
 import javax.persistence.Column;
 import javax.persistence.Entity;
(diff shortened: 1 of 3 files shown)


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditRunController.java
```diff
@@ -234,7 +234,9 @@ public ModelAndView setupForm(Run run, ModelMap model) throws IOException {
       }
 
       model.put("sequencingParameters",
-          sequencingParametersService.getForPlatform((long) run.getSequencer().getPlatform().getId()));
+          sequencingParametersService.getForPlatform((long) run.getSequencer().getPlatform().getId()).stream()
+              .sorted((sp1, sp2) -> sp1.getName().compareTo(sp2.getName()))
```
apmasell: There's `.sorted(Comparator.comparing(SequencingParameters::getName))` which I think reads better.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
```diff
@@ -143,4 +144,32 @@ public void setUpdatedBy(User updatedBy) {
     this.updatedBy = updatedBy;
   }
 
+  public static Comparator<SequencingParameters> nameComparator = (sp1, sp2) -> sp1.compareTo(sp2);
+
+  @Override
+  public int compareTo(SequencingParameters other) {
+    // v4 2x126, v4, Rapid Run, 10X, Custom, v3, other
+    if (!(other instanceof SequencingParameters)) return -1;
```
djcooke: Is this meant to be a null check? If non-null, `other` has to be an instance of `SequencingParameters`

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
```diff
@@ -143,4 +144,32 @@ public void setUpdatedBy(User updatedBy) {
...
+    } else if (name.startsWith("Rapid Run")) {
+      if (other.name.startsWith("v4")) return 1;
+      if (other.name.startsWith("Rapid Run")) return other.name.compareTo(name);
+      return -1; // Rapid Run > everything else
+    } else if (name.startsWith("10X")) {
+      if (other.name.startsWith("v4") || other.name.startsWith("Rapid Run")) return 1;
+      if (other.name.startsWith("10X")) return other.name.compareTo(name);
+      return -1; // 10X > everything else
+    } else if (name.startsWith("v3")) {
+      if (other.name.startsWith("v4") || other.name.startsWith("Rapid Run") || other.name.startsWith("10X")) return 1;
```
djcooke: A more manageable way to write this would be to assign an int to each pattern (1 for `v4 2x126`, 2 for `v4*`, etc.), compare the ints first, and then compare names if the ints are equal. Would result in much less conditional logic, especially if/when new patterns are added

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditRunController.java
```diff
@@ -234,7 +236,9 @@ public ModelAndView setupForm(Run run, ModelMap model) throws IOException {
       }
 
       model.put("sequencingParameters",
-          sequencingParametersService.getForPlatform((long) run.getSequencer().getPlatform().getId()));
+          sequencingParametersService.getForPlatform((long) run.getSequencer().getPlatform().getId()).stream()
+              .sorted(SequencingParameters.nameComparator)
```
djcooke: Can get rid of `nameComparator` and just use `.sorted()` here

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
callunity: Right sorry, too much copypasta. I'll remove this line.

Issue #1736, opened 2018-11-02
Possible StartNoTest syntax error


microbioticajon:
Hi Guys,

I "think" I might have found a subtle bug in the migration scripts

**miso version:** v0.2.148
**mysql version:** mysql  Ver 14.14 Distrib 5.7.24, for Linux (x86_64) using  EditLine wrapper

When initialising a new lims database by running the flyway migration, the migration fails at: sqlstore/src/main/resources/db/migration/V0004__boxes.sql

reporting a syntax error at line 7 "--StartNoTest"

According to:

https://dev.mysql.com/doc/refman/5.7/en/comments.html

comment lines require a white space character immediately after the "--". I found a number of migration scripts that did not have this space. I modified these files by hand then recompiled and that seems to have done the trick.

`find sqlstore/src -name '*.sql' |xargs perl -pi -e 's/--StartNoTest/-- StartNoTest/g'`
`find sqlstore/src -name '*.sql' |xargs perl -pi -e 's/--EndNoTest/-- EndNoTest/g'`

Im surprised that I would be the first person reporting this which suggests that I might be doing something wrong in the first place....

Cheers
Jon

djcooke:
Hello,

What version of Flyway are you using? You should not have this issue if using Flyway version 3.2.1 as recommended in the [Install Guide](https://github.com/TGAC/miso-lims/blob/master/docs/_posts/2016-01-11-installation-guide.md). We are aware of this issue when using newer Flyway versions.

We've chosen to leave the existing migrations as they are for now, as changing them would affect the checksums, invalidating existing users' databases. We may decide in the future to correct these if we can provide a script to automate the schema version (checksum) corrections; however, this may be difficult as some versions of Flyway generate different checksums than others.

Note that if you choose to use a newer version of Flyway, you will have to ensure that you alter the migrations in the exact same way every time before migrating to newer versions. We also cannot guarantee that it will be possible at all to apply future migrations, as we test solely against Flyway 3.2.1.

microbioticajon:
That indeed seems to be the issue:

flyway-5.2.1

I obviously picked the wrong version. Is there any chance you could mention the specific behaviour in the installation guide? It might help muppets like me debug the issue a little quicker.

Many thanks for your response.

PR #1739, opened 2018-11-02
#1736: added note about errors with different Flyway versions

djcooke:


docs/_posts/2016-01-11-installation-guide.md
```diff
@@ -405,6 +405,9 @@ The same path should be used for `MISO_FILES_DIR` as is set for `miso.fileStorag
 jdbc:mysql://localhost:3306/lims?autoReconnect=true&zeroDateTimeBehavior=convertToNull&useUnicode=true&characterEncoding=UTF-8
 ```
 
+If you encounter errors migrating the database, make sure that you are using the recommended version of Flyway (see above).
+There are known issues with using newer Flyway versions with the MISO migrations.
```
callunity: Might be worth saying something like "If migration fails on V000#, you may be using a newer version of Flyway which MISO does not support. As Flyway has changed their checksum calculations in newer versions and we don't want to break completed migrations for existing users, there are no plans to support newer versions of Flyway." (or a similar explainer of why we won't be supporting a newer version of Flyway)

docs/_posts/2016-01-11-installation-guide.md
```diff
@@ -405,6 +405,9 @@ The same path should be used for `MISO_FILES_DIR` as is set for `miso.fileStorag
 jdbc:mysql://localhost:3306/lims?autoReconnect=true&zeroDateTimeBehavior=convertToNull&useUnicode=true&characterEncoding=UTF-8
 ```
 
+If you encounter errors migrating the database, make sure that you are using the recommended version of Flyway (see above).
```
callunity: Could you please link to the line above?

docs/_posts/2016-01-11-installation-guide.md
djcooke: I don't want to mention specific errors or locations of errors since they may vary depending on the exact Flyway version. If _anything_ fails, step 1 should be to make sure you're using the correct version.

I'll add the note about why we don't plan on supporting newer versions though.

PR #1760, opened 2018-11-23
GLT-2322: added partition loading concentration

djcooke:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/spreadsheet/PartitionSpreadsheets.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import uk.ac.bbsrc.tgac.miso.core.data.Partition;
+
+public enum PartitionSpreadsheets implements Spreadsheet<Partition> {
+
+  LOADING_CONCENTRATIONS("Loading Concentrations", //
+      Column.forString("Container", p -> p.getSequencerPartitionContainer().getAlias()), //
+      Column.forInteger("Number", Partition::getPartitionNumber), //
```
callunity: The integer is being converted to a decimal in the spreadsheet.

Because of this I found "Number" to be confusing... "Partition" or "{platform-specific term}" would also make it clearer.

djcooke: Fixed, though it's now written as a string. Changed column name to "Partition" as using the platform-specific terms would have required more refactoring of the spreadsheet utilities

callunity: Sounds good, on both counts

PR #1761, opened 2018-11-23
GP-1840: include lane QC data in Pinery

djcooke:
Will not merge until Pinery is changed to a release version


pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/MisoClient.java
```diff
@@ -644,6 +644,8 @@ public MisoRunPosition mapRow(ResultSet rs, int rowNum) throws SQLException {
       p.setPoolBarcode(rs.getString("pool_barcode"));
       p.setPoolCreatedById(rs.getInt("pool_createdById"));
       p.setPoolCreated(rs.getDate("pool_created"));
+      p.setAnalysisSkipped(rs.getBoolean("analysis_skipped"));
```
apmasell: Doing [getBoolean](https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#getBoolean(java.lang.String)) drops the null values and converts them to false. I think that's okay, unless we've got some reason to be concerned about that.

PR #1766, opened 2018-11-27
Sample & library distribution

callunity:
GLT-2034


miso-web/src/main/webapp/scripts/hot_sample.js
```diff
@@ -814,13 +816,101 @@ HotTarget.sample = (function() {
...
-              (show['Aliquot'] && config.targetSampleClass.alias.indexOf('Single Cell') != -1), 'inputIntoLibrary', 14, 10, false, false)];
+              (show['Aliquot'] && config.targetSampleClass.alias.indexOf('Single Cell') != -1), 'inputIntoLibrary', 14, 10, false, false),
+
+          // Distribution to collaborator or outside destination
+          {
+            header: 'Distributed',
+            data: 'distributed',
+            type: 'dropdown',
+            trimDropdown: false,
+            source: ['True', 'False'],
```
apmasell: Are these the best terms for these? Maybe _Sent_ and _On-site_. I dunno if Illinca has preferences.

PR #1771, opened 2018-11-29
Add distribution info to Pinery

callunity:
GLT-2034
Also fixes a bug with saving distribution info

### DIFF
--- a/miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/BulkSampleEditIT.java
+++ b/miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/BulkSampleEditIT.java
@@ -286,6 +286,90 @@ public class BulkSampleEditIT extends AbstractBulkSampleIT {
   }
 
   @Test
+  public void testDistributeTissue() throws Exception {
+    // Goal: ensure a tissue can be put in a box and then distributed and saved which should remove it from the box
+    BulkSamplePage page = getEditPage(Arrays.asList(getSampleId("Tissue")));
+    HandsOnTable table = page.getTable();
+
+    Map<String, String> editable = new HashMap<>();
+    editable.put(SamColumns.ALIAS, "TEST_0002_Pa_M_13_2-2");
+    editable.put(SamColumns.DESCRIPTION, "changed description");
+    editable.put(SamColumns.RECEIVE_DATE, "2016-07-17");
+    editable.put(SamColumns.ID_BARCODE, "3001"); // increment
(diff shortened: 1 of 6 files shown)


miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/BulkSampleEditIT.java
```diff
@@ -285,6 +285,90 @@ public void testEditSaveEmptySaveTissue() throws Exception {
     assertAllForTissue(empty, getIdForRow(newTable, 0), false);
   }
 
+  @Test
+  public void testDistributeTissue() throws Exception {
```
djcooke: This is really here to test the Service and DAO rather than the UI. Would be simpler and faster to use unit tests instead - I think a test for `DefaultSampleService#applyChanges` and another for `HibernateSampleDao#save` could do the job

pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/MisoClient.java
```diff
@@ -832,6 +832,15 @@ public String extractStringValueFrom(ResultSet rs) throws SQLException {
       STAIN("stain", "Stain"),
       SLIDES("slides", "Slides"),
       DISCARDS("discards", "Discards"),
+      DISTRIBUTED("distributed", "Distributed") {
+        @Override
+        public String extractStringValueFrom(ResultSet rs) throws SQLException {
+          boolean distributed = rs.getBoolean("distributed");
```
djcooke: Should use `getSqlKey()` instead of repeating String `"distributed"` that gets set in the constructor

miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/BulkSampleEditIT.java
callunity: I was actually trying to test the UI effects (that the box field would be cleared and that the Distribution Date and Distribution Recipient would only be editable after Distributed is set to Sent Out) but I didn't make that clear in my test! Will update the test to confirm it does what I want it to, and will add unit tests as well.

djcooke: Those UI effects could be tested more simply by loading the edit page for a sample already in a box and changing the distribution properties. No saves would be necessary. There's `HandsOnTable#isWritable` to check if fields are writable or read-only too

PR #1781, opened 2018-12-07
Lengthen project short names

apmasell:
JIRA: GLT-2647

### DIFF
--- a/sqlstore/src/main/resources/db/migration/V8080__shortname.sql
+++ b/sqlstore/src/main/resources/db/migration/V8080__shortname.sql
@@ -1,0 +1,1 @@
+ALTER TABLE Project CHANGE COLUMN shortName shortName VARCHAR(255);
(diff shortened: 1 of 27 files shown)


sqlstore/src/main/resources/db/migration/V8080__shortname.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE Project CHANGE COLUMN shortName shortName VARCHAR(255);
```
callunity: The edit project page uses the length of this column to dictate how many characters are allowed in the input field. This will fail.

apmasell: Wait, so we should change the page to get length information from the naming scheme rather than the DB (or probably some min of the two)?

apmasell:
I've linked the shortnames regex to the sample and library alias generators. Still figuring out the validation.

callunity:
Bumping the request to alter a sample receipt/library receipt (one which generates ghost samples) to confirm that the sample alias generator still works correctly with this change.

djcooke:
Alias generator can be tested with unit tests rather than UI integration test

sqlstore/src/main/resources/db/migration/V8080__shortname.sql
callunity: I'd be up for that. What do you think, @djcooke ?

djcooke: Makes sense to get the length from the validator instead

apmasell:
Okay. This has been fixed and the receipt and library alias generator are working again.

PR #1782, opened 2018-12-10
Remove project overviews pending redesign

apmasell:
JIRA: GLT-2649


sqlstore/src/main/resources/db/migration/V8899__project_overview.sql
```diff
@@ -0,0 +1,4 @@
+DROP TABLE ProjectOverview_Watcher;
+-- TODO DROP TABLE ProjectOverview;
+-- TODO DROP TABLE ProjectOverview_Sample;
+-- TODO DROP TABLE ProjectOverview_Note;
```
djcooke: remove comments

apmasell:
Fixed.

PR #1784, opened 2018-12-10
Remove watcher funcionality since it doesn't do anything

apmasell:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditProjectController.java
```diff
@@ -283,11 +283,6 @@ public ModelAndView setupForm(@PathVariable Long projectId, ModelMap model) thro
     model.put("overviews", project.getOverviews());
 
     Map<Long, String> overviewMap = new HashMap<>();
-    for (ProjectOverview po : project.getOverviews()) {
-      if (po.getWatchers().contains(user)) {
-        overviewMap.put(po.getId(), user.getLoginName());
-      }
-    }
     model.put("overviewMap", overviewMap);
```
djcooke: remove map declaration and model attribute

sqlstore/src/main/resources/db/migration/V88977_watch.sql
```diff
@@ -0,0 +1,3 @@
+DROP TABLE Pool_Watcher;
+DROP TABLE Project_Watcher;
+DROP TABLE Run_Watcher;
```
djcooke: Should also drop from the `_Group` table where `name` in (RunWatchers, ProjectWatchers, PoolWatchers)

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditProjectController.java
apmasell: This happens in #1782. I didn't want to have to edit the JSP much at this point.

PR #1786, opened 2018-12-12
GLT-2641 fix boxing bugs for discards/distributed

callunity:


miso-web/src/main/webapp/scripts/hot_boxable.js
```diff
@@ -90,6 +90,19 @@ HotTarget.boxable = (function() {
             },
             pack: function(obj, flat, errorHandler) {
               // search field only
+            },
+            depends: ['discarded', 'distributed'],
+            update: function(obj, flat, flatProperty, value, setReadOnly, setOptions, setData) {
+              var applyChanges = function(source, autoSelect) {
```
djcooke: function doesn't seem to be used anywhere, and doesn't make sense for the search column

miso-web/src/main/webapp/scripts/hot_boxable.js
```diff
@@ -90,6 +90,19 @@ HotTarget.boxable = (function() {
...
+                setOptions({
+                  source: source
+                });
+                setData(source.length > 1 && !autoSelect ? 'SELECT' : source[0]);
+              };
+              if ((flatProperty === 'discarded' && value == 'True') || (flatProperty == 'distributed' && value == 'Sent Out')) {
+                setData(null);
+                setReadOnly(true);
+              }
```
djcooke: Need `else` block to call `setReadOnly(false);`

Box Alias column below should depend on `discarded` similarly to this column

callunity: Done

miso-web/src/main/webapp/scripts/hot_boxable.js
```diff
@@ -118,58 +127,69 @@ HotTarget.boxable = (function() {
...
+                    (boxSearchCache[value.toLowerCase()][0].name.toLowerCase() == value.toLowerCase()
+                        || boxSearchCache[value.toLowerCase()][0].alias.toLowerCase() == value.toLowerCase() || (boxSearchCache[value
+                        .toLowerCase()][0].identificationBarcode && boxSearchCache[value.toLowerCase()][0].identificationBarcode
+                        .toLowerCase() == value.toLowerCase())));
                 return;
               }
 
               setData('(...searching...)');
               var deferred = jQuery.Deferred();
-              jQuery.ajax({
```
callunity: No actual code changes from here on down, Eclipse just decided to reformat things.

PR #1787, opened 2018-12-12
Make compact migrations yell loudly on fail

callunity:


compact-migrations
```diff
@@ -48,6 +48,19 @@ do
 	echo >> "${NEXT}"
 	git rm "${FILE}"
 done
+
+ls | grep "^V8.*" > dangling_migrations
+DANGLING_MIGRATION_COUNT=$(cat dangling_migrations | wc -l)
```
djcooke: Could just do

```
DANGLING_MIGRATION_COUNT=`ls | grep "^V8.*" | wc -l`
```

and then repeat the `ls | grep "^V8.*"` part below to avoid writing a file.

Or delete the file after

callunity: Updated

PR #1788, opened 2018-12-13
GLT-2652: added spreadsheet import function to bulk input tables

djcooke:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/HotRestController.java
```diff
@@ -47,4 +56,71 @@ public void setRows(List<List<String>> rows) {
...
+      this.data = data;
+    }
+
+  }
+
+  @PostMapping("/import")
+  public @ResponseBody List<ColumnDataDto> importSpreadsheet(@RequestParam("file") MultipartFile file) throws IOException {
+    String extension = file.getOriginalFilename().substring(file.getOriginalFilename().lastIndexOf(".") + 1).toLowerCase();
+    SpreadsheetWrapper sheet = null;
+    if (extension.equals("xlsx")) {
```
apmasell: If you're going to go to the trouble of extracting and extension, I think it makes more sense to use a `switch` block. That, or don't bother and just do `filename.endsWith(".xlsx")`.

djcooke:
@callunity: updated all Open Document spreadsheet exports to use the .ods extension. .odt is for Open Document text documents.

PR #1790, opened 2018-12-17
added Dockerfile for Pinery-MISO

djcooke:


pinery-miso-dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM maven:3.6.0-jdk-8 as builder
+COPY . /miso-lims
+RUN cd miso-lims && mvn clean package
```
callunity: Would be nice to have this run maven with `-q` to move the internet-downloading offscreen

djcooke: added this as well as `-B` (batch-mode)

djcooke:
Switched to Alpine base images to reduce final image size

PR #1791, opened 2018-12-17
GLT-2651: removed old form import/export functionality

djcooke:


miso-web/src/main/webapp/pages/adminsub.jsp
```diff
@@ -29,8 +29,6 @@
 	<div class="menu">
 		<a href="<c:url value="/miso/sample/receipt"/>">Receive
```
callunity: Please also remove this line, as the sampleReceipt JSP is deleted :)

djcooke: Weird - I thought I had. It's gone now

PR #1792, opened 2018-12-18
Move alias validation to REST controllers

callunity:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ContainerRestController.java
```diff
@@ -147,4 +153,53 @@ public void bulkDelete(@RequestBody(required = true) List<Long> ids) throws IOEx
     return MisoWebUtils.generateSpreadsheet(input, PartitionSpreadsheets::valueOf, request, response);
   }
 
+  @PostMapping(value = "/validate-serial-number")
+  public ResponseEntity<?> validateSerialNumber(@RequestBody String params, HttpServletRequest request, HttpServletResponse response,
```
djcooke: These `RestController` methods should all return DTOs, or annotate with `@HttpStatus(Status.NO_CONTENT)` to indicate that no data is returned on success.  `RestException` should be thrown for any errors

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ContainerRestController.java
```diff
@@ -147,4 +150,53 @@ public void bulkDelete(@RequestBody(required = true) List<Long> ids) throws IOEx
...
+      containerId = Long.valueOf(maybeContainerId);
+    }
+    try {
+      List<SequencerPartitionContainer> matchingContainers = (List<SequencerPartitionContainer>) containerService
+          .listByBarcode(serialNumber);
+      if (matchingContainers.isEmpty()
+          || (matchingContainers.size() == 1 && matchingContainers.get(0).getId() == containerId)) {
+        return;
+      } else {
+        throw new RestException("Serial number is already associated with another container", Status.PRECONDITION_FAILED);
```
djcooke: All instances of `Status.PRECONDITION_FAILED` should be changed to `Status.BAD_REQUEST`. Precondition failed status refers to conditions specified in request headers - not the user/application data

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ContainerRestController.java
```diff
@@ -147,4 +150,53 @@ public void bulkDelete(@RequestBody(required = true) List<Long> ids) throws IOEx
...
+      }
+    } catch (IOException e) {
+      String error = String.format("Error looking up containers by serial number '%s'", serialNumber);
+      log.error(error, e);
```
djcooke: Exceptions are already logged, so this log call is unnecessary. This catch block can be removed entirely though. If the method throws an `IOException`, it'll already be translated to a `RestError` with 500 status in `RestExceptionHandler`. The only difference is the error message will be less useful, but neither message will really be useful to a user

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ContainerRestController.java
```diff
@@ -147,4 +150,53 @@ public void bulkDelete(@RequestBody(required = true) List<Long> ids) throws IOEx
...
+    Long containerId = SequencerPartitionContainerImpl.UNSAVED_ID;
+    if (!maybeContainerId.contains("Unsaved")) {
+      containerId = Long.valueOf(maybeContainerId);
+    }
+    try {
+      List<SequencerPartitionContainer> matchingContainers = (List<SequencerPartitionContainer>) containerService
+          .listByBarcode(serialNumber);
+      if (matchingContainers.isEmpty()
+          || (matchingContainers.size() == 1 && matchingContainers.get(0).getId() == containerId)) {
+        return;
```
djcooke: I agree with Sonar here - could just reverse the conditions and throw exception without the need for an else block

miso-web/src/main/webapp/scripts/project_ajax.js
```diff
@@ -28,21 +28,22 @@ var projectId_sample, sampleQcTypesString, libraryQcTypesString, projectId_d3gra
...
+    }).success(function(json) {
+      deferred.resolve();
+    }).fail(function(response, textStatus, serverStatus) {
+      deferred.reject(response); // need to upgrade Parsley to get custom error messages
     });
     return deferred.promise();
   },
   messages: {
-    en: 'Short name must conform to the naming scheme.'
+    en: 'Short name must be unique and must conform to the naming scheme.'
```
djcooke: The naming scheme dictates whether duplicates are allowed, so that should not be stated here

PR #1794, opened 2018-12-18
GLT-2654: moved dashboard search functionality to REST controllers

djcooke:


miso-web/src/main/webapp/scripts/pane_project.js
```diff
@@ -0,0 +1,53 @@
...
+ * *********************************************************************
+ */
+
+PaneTarget.project = (function() {
+  var title = "Project";
+  var url = "/miso/rest/project/search";
+
+  var transform = function(project) {
+    return Tile.make([Tile.title(project.shortName || project.alias),
+        Tile.lines(["Name: " + project.name, project.shortName ? ("Alias: " + project.alias) : null].filter(function(x) {
```
callunity: Is this ternary correct? If shortName is falsey, use `null` ?

djcooke: Yes. If `shortName` is falsey, then `alias` is already used for the tile title, so doesn't need to be included in the details

PR #1799, opened 2018-12-19
Update Dockerfile

leipzig:
This fixes a mysql "Can't open and lock privilege tables: Table storage engine for 'user' doesn't have this option" error that MySQL 5.7.23 raises in the miso-lims Dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,8 @@ RUN     apt-get -y update && apt-get -y install --no-install-recommends     \
         cd /home/ && bash /root/miso-ansible/files/miso-install.sh $version   &&\
         apt-get purge --auto-remove -q -y                                   \
             unzip xmlstarlet                                                &&\
-        rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+        rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
+	chown -R mysql:mysql /var/lib/mysql /var/run/mysqld
 
 CMD	["/root/miso-ansible/misoStart.sh"]
 


callunity:
@leipzig thanks for the patch! How are you building the container? I ask because when I build from this patch using the following, Tomcat fails to start due to a failure to connect to the database:
```
docker build -t "misolims/miso-lims:leipzig-patch-1" --no-cache .
docker run -p 8092:8080 -t "misolims/miso-lims:leipzig-patch-1"
```

leipzig:
I opened up a shell and restarted it manually
```
/etc/init.d/mysql restart
```

callunity:
Can the PR be updated to make it so that no manual restart is necessary? As is, the image still ends up in a state that isn't straightforward to use.

callunity:
@leipzig are you downloading the image from Docker Hub (via `docker pull misolims/miso-lims`), or are you building it on your machine? If the latter, could you share the command you are using to build it?

leipzig:
from the hub (i don't have the WAR file built)

callunity:
@leipzig I fixed this issue in the `miso-base` where MySQL is installed, and have a branch [here](https://github.com/TGAC/miso-lims/tree/fix_docker) that will allow you to build a working copy of MISO using the following:
```
MISO_VERSION=$(xmlstarlet sel -t -v /_:project/_:version pom.xml)
sudo docker build -t "misolims/miso-lims:${MISO_VERSION}" --build-arg version="${MISO_VERSION}" --no-cache .
```
I'm not going to merge this yet because due to our release schedule, the fixed Docker version won't be available for another eight weeks or so (I don't want it to look like it's fixed when it isn't yet). You should be able to cherry-pick the one or two most recent commits on that branch onto any other branch and build an image that will run correctly the first time on your machine.

PR #1800, opened 2019-01-02
Update deleteRun procedure

callunity:
GLT-2666


djcooke:
Hmm.. The procedure for deleting a run (or anything else attachable) should now also include deleting attachments from the file system

apmasell:
I think it makes more sense to have some kind of GC tool that deletes any unreferenced files and removes and references from the DB to files that no longer exist.

callunity:
@apmasell wanna make a ticket for that?

PR #1804, opened 2019-01-07
Handle null Volume and Concentration units

avarsava:
Change LimsUtils.makeVolumeAndConcentrationLabel to accept VolumeUnit
and ConcentrationUnit types rather than String.
makeVolumeAndConcentrationLabel tests for null inside method, removing
need for testing getVolumeUnits() and getConcentrationUnits() for null
within the call to makeVolumeAndConcentrationLabel.


avarsava:
@djcooke as I understand it, the only time the units can be null is when there is no value for Volume or Concentration set. When there is no value input by the user, no unit choice is saved by MISO.

djcooke:
@avarsava Oh, I see. The problem was NPE in the calls rather than the method itself. No bug then :+1:

PR #1807, opened 2019-01-09
Fix bug in distributing from edit Sample/Lib page

callunity:
GLT-2676


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/ValidationUtils.java
```diff
@@ -77,9 +77,4 @@ public static void validateDistributionFields(boolean isDistributed, Date distri
     }
   }
 
-  public static void validateUnboxableFields(boolean discarded, boolean distributed, Box box, Collection<ValidationError> errors) {
```
djcooke: It wouldn't hurt to keep this validation in place to assert that things are working as intended

PR #1811, opened 2019-01-10
Fix up multiline label layout

apmasell:
JIRA: GLT-2678

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/printing/LabelCanvas.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/printing/LabelCanvas.java
@@ -74,7 +74,7 @@ public abstract class LabelCanvas {
   }
 
   /**
-   * Create a mulit-line text block
+   * Create a multi-line text block
    *
    * @param x the x coordinator the top-left position of the text block in mm
    * @param y the y coordinator the top-left position of the text block in mm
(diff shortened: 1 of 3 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/printing/LabelCanvas.java
```diff
@@ -90,18 +90,24 @@ public final void multilineText(double x, double y, double height, int numCharsP
...
         .filter(input -> !LimsUtils.isStringBlankOrNull(input.getValue()))//
         .reduce(0, (offset, input) -> {
+          // This block is full. This line of text gets ignored. So sad.
+          if (offset >= maxLines) {
+            return offset;
+          }
           int line;
-          for (line = offset; line < maxLines - 1 && input.getValue().length() > line * numCharsPerLine; line++) {
-            text(x + direction.xMultiplier * 1.1 * height * line, y + direction.yMultiplier * 1.1 * height * line, height, direction,
+          // Print this text over as many lines as it needs and or the block can hold
```
callunity: `and or` -> `and/or`

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/printing/LabelCanvas.java
```diff
@@ -90,18 +90,24 @@ public final void multilineText(double x, double y, double height, int numCharsP
...
+          for (line = 0; line + offset < maxLines && input.getValue().length() > line * numCharsPerLine; line++) {
+            String str;
+            // If we are out of lines and this string needs more than one, than add some dots.
```
callunity: `If we are on our last line and this string needs more than one, then add some dots at the end of this line to stand in for the additional line(s).`

PR #1817, opened 2019-01-23
User manual: link kits & targeted sequencing

callunity:
GLT-2674

### DIFF
--- a/docs/user_manual/type_data.md
+++ b/docs/user_manual/type_data.md
@@ -340,10 +340,10 @@ will take you to the Create Kit Descriptor page. Enter all of the kit details, t
 right to create the kit descriptor.
 
 Once a kit has been created, it is possible to link existing [Targeted Sequencing](#targeted_sequencing) values to the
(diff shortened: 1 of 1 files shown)


docs/user_manual/type_data.md
```diff
@@ -339,6 +339,12 @@ To add a new kit descriptor, click the "Add" button in the toolbar at the top of
 will take you to the Create Kit Descriptor page. Enter all of the kit details, then click the "Save" button at the top
 right to create the kit descriptor.
 
+Once a kit has been created, it is possible to link existing [Targeted Sequencing](#targeted_sequencing) values to the
+kit descriptor. Then, if a library is prepared using this kit descriptor, dilutions created from the library can select
+a targeted sequencing from the ones linked to the kit descriptor. This means that if a desired targeted sequencing is
```
avarsava: Does this mean MISO has logic so that when Dilutions are created, they select their own targeted sequencing?

callunity: Hmm, I'll change it so the dilution has less agency.

PR #1829, opened 2019-02-13
Check targeted sequencing when changing library code

apmasell:
JIRA: GLT-2727

### DIFF
--- a/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
+++ b/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
@@ -12,6 +12,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang.exception.ExceptionUtils;
 import org.hibernate.exception.ConstraintViolationException;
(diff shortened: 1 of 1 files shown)


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
```diff
@@ -588,6 +594,20 @@ private void validateChange(Library library, Library beforeChange) throws IOExce
...
   }
 
+  private void validateTargetedSequencing(LibraryDesignCode libraryDesignCode, Collection<LibraryDilution> libraryDilutions,
+      List<ValidationError> errors) throws IOException {
+    libraryDesignCode = libraryDesignCodeService.get(libraryDesignCode.getId());
+    if (libraryDesignCode.isTargetedSequencingRequired()) {
+      List<String> badDilutions = libraryDilutions.stream().filter(dilution -> dilution.getTargetedSequencing() == null)
+          .map(LibraryDilution::getName).collect(Collectors.toList());
+      if (!badDilutions.isEmpty()) {
+        errors.add(new ValidationError("libraryDesignCodeId",
```
djcooke: Property should be "libraryDesignCode" rather than "libraryDesignCodeId"

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
```diff
@@ -588,6 +594,20 @@ private void validateChange(Library library, Library beforeChange) throws IOExce
...
+            String.format("Library design code %s requires targeted sequencing, but the following dilutions have none: %s",
+                libraryDesignCode.getCode(), String.join(", ", badDilutions))));
```
callunity: Could the error message be changed to include the action the user needs to take to get it into a non-error state? Maybe something like "Library design code %s requires targeted sequencing. Please add a targeted sequencing to the affected dilutions (%s) then update the library design code." ?

djcooke: "Targeted sequencing must be assigned to the affected dilutions (%s) to use this library design code" for conciseness and consistent wording with other validation messages

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
apmasell: I'm unclear on these. This isn't the name of the column nor the name of the DTO field. Is this the name in the JSP?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryService.java
apmasell: Done.

PR #1830, opened 2019-02-14
GLT-2716: export NovaSeq workflow type via Pinery

djcooke:
DO NOT MERGE until Pinery version is changed to a release containing https://github.com/oicr-gsi/pinery/pull/96

### DIFF
--- a/pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/MisoClient.java
+++ b/pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/MisoClient.java
@@ -603,6 +603,13 @@ public class MisoClient implements Lims {
 
   private static class RunRowMapper implements RowMapper<Run> {
 
+    private static final Map<String, String> workflowTypeMapper = new HashMap<>();
+
+    static {
+      workflowTypeMapper.put("NOVASEQ_XP", "NovaSeqXp");
+      workflowTypeMapper.put("NOVASEQ_STANDARD", "NovaSeqStandard");
+    }
+
     @Override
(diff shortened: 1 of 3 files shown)


pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/MisoClient.java
```diff
@@ -622,6 +629,11 @@ public Run mapRow(ResultSet rs, int rowNum) throws SQLException {
       r.setRunDirectory(rs.getString("filePath"));
       r.setRunBasesMask(rs.getString("runBasesMask"));
       r.setSequencingParameters(rs.getString("sequencingParameters"));
+      String rawWorkflowType = rs.getString("workflowType");
+      if (!rs.wasNull()) {
```
apmasell: You can just us `r.setWorkflowType(rs.getString("workflowType"))`. It will propagate null correctly:
https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#getString(java.lang.String)
The `wasNull` thing is only necessary for things that return primitive types.

djcooke: The raw value from the query is the MISO enum constant name and is not what we want to use. It gets mapped on the next couple of lines and the raw value is only used in the case of an unknown/unexpected workflow type.

The use of rs.wasNull here is the same as a null check on `rawWorkflowType`

djcooke:
Moved to https://github.com/miso-lims/miso-lims/pull/1994

PR #1834, opened 2019-02-22
GLT-2539: added Home Screen Widgets section to user manual

djcooke:


docs/user_manual/home_screen_widgets.md
```diff
@@ -0,0 +1,73 @@
...
+section: 26
+---
+
+
+{% assign sub = 1 %}
+{% include userman-heading.md section=page.section sub=sub title="Search" %}
+
+The Search widget can be used to quickly find an item in MISO. This includes samples, libraries, library dilutions,
+pools, sequencing containers, and boxes. You can search for an item by its name, alias, or barcode. Scanning an
+item's barcode using a barcode scanner also works. The search returns exact matches only, so must enter the entire
```
avarsava: "so must" -> "so you must"?

docs/user_manual/home_screen_widgets.md
```diff
@@ -0,0 +1,73 @@
...
+layout: page
+title: "26. Home Screen Widgets"
+section: 26
+---
+
+
+{% assign sub = 1 %}
+{% include userman-heading.md section=page.section sub=sub title="Search" %}
+
+The Search widget can be used to quickly find an item in MISO. This includes samples, libraries, library dilutions,
```
apmasell: I think it's worth mentioning that projects and runs can't be searched from this box.

docs/user_manual/home_screen_widgets.md
```diff
@@ -0,0 +1,73 @@
...
+Sequencers that can perform multiple simultaneous runs will show the status of each position. Status can be one of the
+following.
+
+* **Idle**: The sequencer is not currently in use
+* **Available**: One or more of the sequencer's positions are either active or out of service, but one or more
+  positions are still idle
+* **Busy**: All of the sequencer's positions are active
+* **Out of Service**: All of the sequencer's positions are out of service
+
+CLicking on a tile in the Instrument Status widget will open a menu where you can choose to view the instrument, one of
```
avarsava: "CLicking" -> "Clicking"

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultBarcodableViewService.java
```diff
@@ -25,7 +25,7 @@
   BarcodableViewDao barcodableViewDao;
 
   @Resource()
-  private Map<EntityType, BarcodableService> barcodableServicesMap;
+  private Map<EntityType, BarcodableService<?>> barcodableServicesMap;
```
avarsava: Is this related to the Widget documentation?

docs/user_manual/home_screen_widgets.md
djcooke: Added note about other widgets for those

docs/user_manual/home_screen_widgets.md
djcooke: fixed

docs/user_manual/home_screen_widgets.md
djcooke: fixed

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultBarcodableViewService.java
djcooke: Just a missing type parameter I noticed while going through the code related to the documentation. As there was no type before and the type I added is a wildcard, it has no effect besides clearing the warnings.

avarsava: Should these changes be moved to another PR so that the git history for these lines are clearer?

djcooke: If it were a bug fix or behaviour change, I'd agree that it deserved a separate commit, but I'd say no in this case since there is no functional change.

PR #1839, opened 2019-02-22
Fix error when search-by-barcode returns container model

callunity:
Also adds an undeclared dependency and fixes a test


callunity:
Closing as testing still needs to be done for whether the JavaScript tile-making function tries to make a linked tile if no URL is provided. Tracked in ticket [GLT-2735](https://jira.oicr.on.ca/browse/GLT-2735); if it's still open after I get back, I'll pick it up again.

PR #1840, opened 2019-02-22
Remove SecurityProfile

apmasell:
JIRA: GLT-1559


apmasell:
There's a failing test, but it's also failing on develop.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/store/ExperimentStore.java
```diff
@@ -75,5 +76,5 @@
 
   Collection<Experiment> listByLibrary(long id) throws IOException;
 
-  Collection<Experiment> listByRun(long runId) throws IOException;
+  List<Experiment> listByRun(long runId) throws IOException;
```
avarsava: Should these changes be in a different PR?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/ProjectImpl.java
```diff
@@ -112,29 +109,15 @@
...
 
   @Column(nullable = false)
   @Temporal(TemporalType.TIMESTAMP)
   private Date lastUpdated;
 
   /**
-   * Construct a new Project with a default empty SecurityProfile
+   * Construct a new Project
    */
   public ProjectImpl() {
```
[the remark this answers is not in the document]
apmasell: They are implied, but since the have (arguably useless) JavaDoc, I didn't want to delete them.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/store/ExperimentStore.java
apmasell: So, we actually use some kind of list for all storage, but return a mix of `Collection` and `List`. This mostly didn't matter because the security filters would was `<T> List<T> filterUnreadable(Collection<T>)`. With that gone, it was a more straight forward to update everything to use just `List`.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultPoolOrderService.java
```diff
@@ -116,10 +111,4 @@ public void beforeDelete(PoolOrder object) throws IOException {
     poolService.save(pool);
   }
 
-  @Override
-  public void authorizeDeletion(PoolOrder object) throws IOException {
-    Pool pool = poolService.get(object.getPool().getId());
-    authorizationManager.throwIfNotWritable(pool);
```
djcooke: Removing this means that only admins will be able to delete orders. Please update user manual to indicate this

apmasell: Hmm. Should we replace that with a admin-or-creator?

djcooke: Probably. "Writeable" technically meant everyone before, but I don't think everyone should be able to delete orders

apmasell: Sounds good to me.

PR #1842, opened 2019-02-25
QCs Section and Finalizing New User Manual

djcooke:


docs/user_manual/qcs.md
```diff
@@ -0,0 +1,41 @@
+---
+layout: page
+title: "20. QCs"
+section: 20
+---
+
+QCs in MISO are used to record the results of quality control checks. They can be added to samples, libraries, pools,
+and sequencing containers. It is possible to add the same QC to the same item multiple times to show a change in the
```
avarsava: Unclear to me how adding the /same/ QC to an item multiple times would show a change. Knowing little about the feature, 1 QC to me seems like 1 record of a quality control check having been completed, so individual QCs should not change (that'd be rewriting history). Could the data representation here be elaborated on / is it explained elsewhere?

docs/user_manual/qcs.md
```diff
@@ -0,0 +1,41 @@
...
+results. There can be several QC types for any particular type of item, and these can be set up as described in the
+[Type Data section - QC Types](type_data.html#qc_types).
+
+{% assign sub = 1 %}
+{% include userman-heading.md section=page.section sub=sub title="Adding QCs" %}
+
+A bulk input page is used for adding QCs. YOu can get there from the list page of the item type you would like to add
```
avarsava: YOu -> You

docs/user_manual/qcs.md
djcooke: Reworded to clarify that multiple QCs of the same type can be added, rather than multiple of the same QCs

docs/user_manual/qcs.md
djcooke: I seem to do this a lot... fixed

djcooke:
@apmasell: added a note that partition QCs are different, which links to a new subsection of the Sequencer Runs section to explain partition QCs, and also added a Partition QC Types subsection to the Type Data section

PR #1855, opened 2019-03-20
GLT-2746: fixed warning messages in assign pool dialog; refactored wa…

djcooke:
…rnings js

### DIFF
--- a/miso-web/src/main/webapp/scripts/list_partition.js
+++ b/miso-web/src/main/webapp/scripts/list_partition.js
@@ -50,8 +50,6 @@ ListTarget.partition = {
       });
 
       response.items.forEach(function(item) {
-        var warnings = WarningTarget.pool.tileWarnings(item.pool);
-
         var dilutionInfo = item.pool.pooledElements.filter(function(element, index, array) {
           return array.length < maxDilutions || index < maxDilutions - 1;
         }).map(function(dilution) {
@@ -69,8 +67,8 @@ ListTarget.partition = {
                   + (order.remaining == 1 ? platformType.partitionName : platformType.pluralPartitionName) + " remaining";
             });
 
-        var tileParts = [Tile.titleAndStatus(item.pool.name + " (" + item.pool.alias + ")", warnings.length == 0 ? Tile.statusOk() : Tile
-            .statusBad())].concat(warnings);
+        var tileParts = [Tile.titleAndStatus(item.pool.name + " (" + item.pool.alias + ")", Warning.hasWarnings(WarningTarget.pool,
+            item.pool) ? Tile.statusBad() : Tile.statusOk())].concat(Warning.generateTileWarnings(WarningTarget.pool, item.pool));
         tileParts.push(Tile.lines(dilutionInfo, false));
         tileParts.push(Tile.lines(orderInfo, true));
 
@@ -277,10 +275,14 @@ ListTarget.partition = {
       "bSortDirection": true
     }, {
       "sTitle": "Pool",
-      "mData": "pool",
+      "mData": function(full) {
+        return full.pool ? full.pool.name + ' (' + full.pool.alias + ')' : '(none)';
+      },
       "include": config.showPool,
       "iSortPriority": 0,
-      "mRender": WarningTarget.partition.tableWarnings
+      "mRender": Warning.tableWarningRenderer(WarningTarget.partition, function(full) {
+        return '/miso/pool/' + full.pool.id;
+      })
     }, {
       "sTitle": "Dilutions",
       "mData": "pool",
--- a/miso-web/src/main/webapp/scripts/list_partition.js
+++ b/miso-web/src/main/webapp/scripts/list_partition.js
@@ -68,7 +68,7 @@ ListTarget.partition = {
             });
 
         var tileParts = [Tile.titleAndStatus(item.pool.name + " (" + item.pool.alias + ")", Warning.hasWarnings(WarningTarget.pool,
-            item.pool) ? Tile.statusBad() : Tile.statusOk())].concat(Warning.generateTileWarnings(WarningTarget.pool, item.pool));
+            item.pool) ? Tile.statusBad('Warning') : null)].concat(Warning.generateTileWarnings(WarningTarget.pool, item.pool));
         tileParts.push(Tile.lines(dilutionInfo, false));
         tileParts.push(Tile.lines(orderInfo, true));
 
@@ -281,7 +281,7 @@ ListTarget.partition = {
       "include": config.showPool,
       "iSortPriority": 0,
       "mRender": Warning.tableWarningRenderer(WarningTarget.partition, function(full) {
-        return '/miso/pool/' + full.pool.id;
+        return full.pool ? ('/miso/pool/' + full.pool.id) : null;
       })
     }, {
       "sTitle": "Dilutions",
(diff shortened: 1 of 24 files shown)


miso-web/src/main/webapp/scripts/list_poolelement.js
```diff
@@ -71,104 +71,107 @@ ListTarget.poolelement = {
...
+          'sTitle': 'Dilution Name',
+          'mData': 'name',
+          'include': true,
+          'iSortPriority': 1
+        },
+        {
+          "sTitle": "Warnings",
+          "mData": null,
+          "mRender": Warning.tableWarningRenderer(WarningTarget.poolelement.makeTarget(config.duplicateIndicesSequences,
+              config.nearDuplicateIndicesSequences)),
```
callunity: If the information about consent revoked is available here, could it be warned about as well?

miso-web/src/main/webapp/scripts/list_poolelement.js
```diff
@@ -71,104 +71,107 @@ ListTarget.poolelement = {
...
+          "bSortable": false
+        }, {
+          'sTitle': 'Proportion',
+          'sType': 'numeric',
+          'mData': 'proportion',
+          'include': !config.add,
+          'iSortPriority': 0
+        }, ListUtils.idHyperlinkColumn("Library Name", "library", "library.id", function(dilution) {
+          return dilution.library.name;
+        }, 0, true, "noPrint"), ListUtils.labelHyperlinkColumn("Library Alias", "library", function(dilution) {
```
callunity: The formatting here makes it harder to see at a glance what is going in, because it look like the close-braces is the end of the column - but the end of the column is the close-parentheses after "noPrint". It might be more readable if the "return....}" lines of the inner functions were indented more? Or if each ListUtils.X was on a new line?

miso-web/src/main/webapp/scripts/list_poolelement.js
```diff
@@ -71,104 +71,107 @@ ListTarget.poolelement = {
...
+            return html;
+          }
+        }, {
+          'sTitle': 'Last Modified',
+          'mData': 'lastModified',
+          'include': true,
+          'iSortPriority': 0,
+          'sClass': 'noPrint'
+        }, {
+          'sTitle': 'Low Quality',
```
callunity: What do you think about displaying this low quality flag in the Warnings column instead of its own column?

miso-web/src/main/webapp/scripts/warning_library.js
```diff
@@ -22,23 +22,21 @@
...
-    warnings = Warning.addWarnings([
-      [library.subprojectPriority, 'PRIORITY (' + library.subprojectAlias + ')', "info"],
-      [parseFloat(library.volume) < 0, '(NEGATIVE VOLUME)'],
-      [library.identityConsentLevel === 'Revoked', '(CONSENT REVOKED)']
-      ], warnings);
-    return Warning.generateTableWarnings(data, warnings);
-  },
-  
+  getWarnings: function(library) {
+    return [{
```
callunity: I had expected that this would warn about low-quality library, but it looks like that's never been there. What do you think about adding it?

miso-web/src/main/webapp/scripts/warning_pool.js
```diff
@@ -22,61 +22,52 @@
...
+          tileMessage: 'MISSING INDEX'
+        }, {
+          include: pool.hasLowQualityLibraries,
+          headerMessage: 'This pool contains at least one low quality library!',
+          tableMessage: '(LOW QUALITY LIBRARIES)',
+          tileMessage: 'LOW QUALITY LIBRARIES'
+        }, {
+          include: revokedDilutions.length > 0,
+          headerMessage: "Donor has revoked consent for " + revokedDilutions.toString(),
+          tableMessage: '(CONSENT REVOKED)',
```
callunity: For some reason, I'm not able to get this warning to show up on the list pools table...

miso-web/src/main/webapp/scripts/list_poolelement.js
djcooke: That's already done. The near/duplicate sequences are required in order to generate the WarningTarget because that information isn't available in the poolelement itself. Consent is available in the poolelement though, and the warnings are being displayed correctly as far as I can see?

miso-web/src/main/webapp/scripts/list_poolelement.js
djcooke: The JavaScript formatter in Eclipse is terrible. You are welcome to fight with it if you'd like :grin:

miso-web/src/main/webapp/scripts/list_poolelement.js
djcooke: Sure, will do

miso-web/src/main/webapp/scripts/list_partition.js
```diff
@@ -277,10 +275,14 @@ ListTarget.partition = {
...
       "sTitle": "Pool",
-      "mData": "pool",
+      "mData": function(full) {
+        return full.pool ? full.pool.name + ' (' + full.pool.alias + ')' : '(none)';
+      },
       "include": config.showPool,
       "iSortPriority": 0,
-      "mRender": WarningTarget.partition.tableWarnings
+      "mRender": Warning.tableWarningRenderer(WarningTarget.partition, function(full) {
+        return '/miso/pool/' + full.pool.id;
```
callunity: `Cannot read property 'id' of null` (when there's no pool in the lane)

miso-web/src/main/webapp/scripts/warning.js
```diff
@@ -22,38 +22,63 @@
...
+
+  return {
+    generateHeaderWarnings: function(containerId, target, item) {
+      // Note: header warnings currently ignore the warning level and show all warnings in red
+      var warnings = getWarnings(target, item);
+      if (warnings.length == 0) {
+        return;
+      }
+      var container = $('#' + containerId);
+      container.append($('<span>').css('float', 'right').append($('<img>').attr('src', '/styles/images/fail.png')));
```
callunity: These pass/fail icons are the one place where we use red & green for contrast, and they're pretty identical for users with red/green colourblindness. Could alt text be added to give additional cues?

miso-web/src/main/webapp/scripts/list_poolelement.js
callunity: Oh, I see. And yes, looks like it's displaying correctly.

miso-web/src/main/webapp/scripts/warning_pool.js
djcooke: The dilutions weren't being included in the PoolDto. This used to be a performance optimization, but the dilutions all get loaded now either way since we're checking for duplicate indices. I'll include them

miso-web/src/main/webapp/scripts/warning_library.js
djcooke: Adding

miso-web/src/main/webapp/scripts/warning.js
djcooke: The pool selection dialog now includes red icon with alt text "Warnings" for ones with warnings and no icon for ones without

miso-web/src/main/webapp/scripts/list_partition.js
djcooke: fixed

PR #1859, opened 2019-03-20
Add detailed sample docker-compose; demo data for plain sample

morgantaschuk:
Split the docker-compose file into composable sections.

* docker-compose.yml : starts up plain sample with no local storage
* docker-compose.override.yml : adds the migrations needed for detailed sample and for orders (Sequencing Parameters, Instruments, Targeted Seqeuencing) and adds volumes to persist data between startups

As per [Using multiple compose files](https://docs.docker.com/compose/extends/#different-environments), both of these are used by default when launching `docker-compose up`.

Other compose files:
* docker-compose.detailed.yml : loads the miso.properties that configures detailed sample + OICR alias autogeneration
* docker-compose.env.yml : allows you to override the environment variables for each service
* docker-compose.build.yml : builds the flyway and webapp Docker containers using the current state of the miso-lims directory

You can specify multiple compose files. For detailed samples demo mode using the local directory builds:

```
docker-compose -f docker-compose.yml -f docker-compose.override.yml -f docker-compose.build.yml -f docker-compose.detailed.yml 
```

I also updated the stored procedure documentation (some of it was out of date) and changed the flyway webapp to be able to run any flyway command passed to it. It cleans the DB on failure. Maybe a bad idea? 

Any SQL migrations that are mounted to `/flyway/migrations` in the flyway-migration docker container will be run.


.docker/run-flyway
```diff
@@ -0,0 +1,21 @@
...
+
+password=$(cat ${MISO_DB_PASS_FILE})
+
+command="flyway/flyway  -user='${MISO_DB_USER}' -password='${password}' -url='jdbc:mysql://${MISO_DB_HOST_PORT}/${MISO_DB}?autoReconnect=true&zeroDateTimeBehavior=convertToNull&useUnicode=true&characterEncoding=UTF-8' -outOfOrder=true -locations=classpath:db/migration,classpath:uk.ac.bbsrc.tgac.miso.db.migration,filesystem:/flyway/migrations -placeholders.filesDir='${MISO_FILES_DIR}'"
+
+eval "${command} $1"
+
+rc=$?; if [[ $rc != 0 ]]
+then
+    eval "${command} clean"
```
djcooke: Clean doesn't roll back; it wipes the entire DB. I think that's probably not a good idea here. If a migration fails, it should be up to the person who ran it to clean it up how they want, which should usually mean restoring a backup

morgantaschuk: fixed

morgantaschuk:
I removed the clean command in `run-flyway`, added a docker-compose.prod.yml file with instructions in the installation doc, and updated the readme.

.docker/detailed_sample_config/V0793__detailed_sample_data.sql
```diff
@@ -0,0 +1,14 @@
...
...(117,'Ga','Gastric Margin',1,'2018-09-28 16:53:27',1,'2018-09-28 16:53:27'),(118,'Ep','Epidural mass',1,'2018-09-28 16:53:27',1,'2018-09-28 16:53:27'),(119,'Lg','Leg',1,'2018-10-18 14:51:31',1,'2018-10-18 14:51:31'),(120,'Ck','Cheek',1,'2018-10-18 14:51:31',1,'2018-10-18 14:51:31'),(121,'Ss','Sinus',1,'2018-10-18 14:51:31',1,'2018-10-18 14:51:31'),(122,'Ic','ileocecum',1,'2018-11-22 22:19:35',1,'2018-11-22 22:19:35'),(123,'Py','periampullary',1,'2018-11-22 22:19:35',1,'2018-11-22 22:19:35'),(124,'Ao','Anorectal',1,'2018-12-11 16:49:28',1,'2018-12-11 16:49:28'),(125,'Di','Diaphragm',1,'2019-01-22 17:43:44',1,'2019-01-22 17:43:44'),(126,'Il','Ileum',1,'2019-01-22 17:56:50',1,'2019-01-22 17:56:50'),(127,'Or','Orbit',1,'2019-02-04 18:27:25',1,'2019-02-04 18:27:25'),(128,'La','Lacrimal Sac',1,'2019-02-04 18:27:25',1,'2019-02-04 18:27:25'),(129,'Cj','Conjunctiva',1,'2019-02-04 18:27:25',1,'2019-02-04 18:27:25'),(130,'Tu','Tongue',1,'2019-02-04 18:27:25',1,'2019-02-04 18:27:25'),(131,'No','Nose',1,'2019-02-05 22:01:48',1,'2019-02-05 22:01:48'),(132,'Cs','Cul-de-sac',1,'2019-02-27 16:31:25',1,'2019-02-27 16:31:25'),(133,'Gt','Genital Tract',1,'2019-02-27 16:31:25',1,'2019-02-27 16:31:25');
```
apmasell: All the dates should probably be `CURRENT_TIMESTAMP`, but probably not worth fixing.

PR #1867, opened 2019-03-26
GLT-2428: converted Edit Container page to use form builder

djcooke:


miso-web/src/main/webapp/scripts/form_container.js
```diff
@@ -0,0 +1,110 @@
...
+            return container.model.alias;
+          }
+        }, {
+          title: 'Description',
+          data: 'description',
+          type: 'text',
+          maxLength: 255
+        }, {
+          title: 'Clustering Kit',
+          data: 'clusteringKitId',
```
callunity: I think this is `clusterKitId` in the DTO

miso-web/src/main/webapp/scripts/form_container.js
```diff
@@ -0,0 +1,110 @@
+if (typeof FormTarget === 'undefined') {
+  FormTarget = {};
+}
+FormTarget.container = (function($) {
```
callunity: jQuery isn't used in this function

djcooke:
Thanks @callunity. All fixed

PR #1873, opened 2019-03-28
Process Oxford Nanopore data from run scanner

callunity:
GLT-2778


callunity:
We discussed and agreed to merge this now.

PR #1874, opened 2019-03-28
fix names of PromethION flowcell models

djcooke:


sqlstore/src/main/resources/db/migration/V8502__fix_promethion_models.sql
```diff
@@ -0,0 +1,21 @@
+-- StartNoTest
+DELIMITER //
+CREATE PROCEDURE TempFixModel(iOldName varchar(10), iNewName varchar(10)) BEGIN
+  IF EXISTS (SELECT 1 FROM SequencingContainerModel WHERE alias = iOldName) THEN
+    UPDATE SequencingContainerModel SET alias = iNewName WHERE alias = iOldName;
+  ELSE
+    INSERT INTO SequencingContainerModel (alias, partitionCount, platformType) VALUES (iNewName, 1, 'ILLUMINA');
```
callunity: Is `ILLUMINA` correct here?

djcooke: No. :slightly_frowning_face:

PR #1880, opened 2019-04-03
Now everything is purple to match the new logo

morgantaschuk:
At least there are fewer clashing colours?

image


miso-web/src/main/webapp/styles/style-full.css
```diff
@@ -65,7 +65,11 @@ pre.wrap {
 }
 
 a {
-    color: #0066cc;
+    color: #404087;
+}
+
+a:hover { 
+    text-decoration: none;
```
djcooke: Disappearing underline on hover seems strange to me. Maybe reverse this - `a { text-decoration: none} a:hover { text-decoration: underline }`?

miso-web/src/main/webapp/WEB-INF/pages/editSample.jsp
```diff
@@ -228,7 +228,7 @@
     </tr>
     <c:choose>
     <c:when test="${not detailedSample}">
-	  <tr style="background-color:yellow;">
+	  <tr class="highlightfield">
```
djcooke: https://github.com/TGAC/miso-lims/pull/1881 removes the two changed elements on this page - the synthetic box and the QC status stripe, so these changes will likely be overwritten

morgantaschuk: OK.

miso-web/src/main/webapp/styles/style-full.css
morgantaschuk: Fixed.

PR #1881, opened 2019-04-04
GLT-2425: converted Edit Sample page to use form builder

djcooke:
Also changed ghost sample indication to an info-level warning, made header warnings show in different colours depending on level (info/error), and made it possible to show warnings in individual places (e.g. ghost sample only shown in header warnings, not in table)


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -325,12 +325,21 @@ private Sample save(Sample sample, boolean validateAliasUniqueness) throws IOExc
       // post-save field generation
       boolean needsUpdate = false;
       if (hasTemporaryName(sample)) {
-        created.setName(namingScheme.generateNameFor(created));
+        try {
+          created.setName(namingScheme.generateNameFor(created));
+        } catch (MisoNamingException e) {
```
apmasell: Does `MisoNamingException` serve a purpose anymore? Can it just be replaced with `ValidationError` in all cases? (Not for this PR. Just musing)

djcooke: We'd have to either move all of the naming scheme stuff into the service module, or the validation stuff into the core module/ The former probably makes more sense and I think would be ok. The naming schemes should only be used by the service and miso-web modules I think

djcooke:
Now with prettier warnings:

Screenshot from 2019-04-05 12-50-57

Screenshot from 2019-04-05 12-51-50

PR #1883, opened 2019-04-05
GLT-2799: symlink /storage/files/freezermaps/ to be served statically

djcooke:
Anything you put in `/storage/files/freezermaps/` (depending on your `miso.properties` settings - this is the default location) will be statically served at `<host>/freezermaps/`. Note that similar to the `scripts` and `styles` directories, it will not be under `/miso`

### DIFF
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/context/MisoAppListener.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/context/MisoAppListener.java
@@ -24,6 +24,9 @@
 package uk.ac.bbsrc.tgac.miso.webapp.context;
 
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Date;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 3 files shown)


djcooke:
For this to work, you must add the following to the context xml (`${CATALINA_HOME}/conf/Catalina/localhost/ROOT.xml`) within the Context element:

`<Resources allowLinking="true"/>`

This should be noted in the release notes.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/context/MisoAppListener.java
```diff
@@ -123,6 +127,22 @@ public void contextInitialized(ServletContextEvent event) {
...
 
+  private void linkMapsDir(ServletContext application, String fileStoragePath) {
+    try {
+      Path target = Paths.get(fileStoragePath, "freezermaps");
+      if (!Files.exists(target)) {
+        Files.createDirectory(target);
+      }
+      Path link = Paths.get(application.getRealPath("/"), "freezermaps");
+      if (Files.exists(link)) {
+        Files.delete(link);
```
avarsava: Won't this delete any freezer maps from previous versions of miso when an update happens?

djcooke: It's deleting the symlink; not the actual directory/files

PR #1884, opened 2019-04-05
GLT-2801: Updated to Flyway 5.2.4

djcooke:
Must run included script manually before you can migrate an existing DB via Flyway 5.2.4 (command-line tool or the updated Maven plugin). This should be noted in the release notes


djcooke:
This has to be updated since there's a new migration. It can wait though

PR #1886, opened 2019-04-08
Change links to point to new miso-lims organization

callunity:


miso-web/src/main/webapp/scripts/jquery/css/smoothness/jquery-ui-1.8.custom.css
```diff
@@ -39,7 +39,7 @@
 
 /*
 * jQuery UI CSS Framework
-* Copyright (c) 2010 AUTHORS.txt (http://jqueryui.com/about)
+* Copyright (c) 2019 AUTHORS.txt (http://jqueryui.com/about)
```
avarsava: Is this ours to change?

.docker/detailed_sample_config/miso.properties
```diff
@@ -1,6 +1,6 @@
 #
-# Copyright (c) 2012. The Genome Analysis Centre, Norwich, UK
-# MISO project contacts: Robert Davey @ TGAC
+# Copyright (c) 2019. The Genome Analysis Centre, Norwich, UK
```
avarsava: Should TGAC's full name be replaced by us as well or does this need to stay?

pom.xml
```diff
@@ -125,29 +114,29 @@
...
+      <name>OICR - MISO Releases</name>
+      <url>https://artifacts.oicr.on.ca/artifactory/miso-release</url>
     </repository>
     <snapshotRepository>
-      <id>miso-snapshots</id>
-      <name>MISO snapshot repository</name>
-      <url>scp://repos.tgac.ac.uk/storage/html/maven/miso/snapshots</url>
+      <id>deploy.artifacts.oicr</id>
+      <name>OICR - MISO Snapshots</name>
+      <url>https://artifacts.oicr.on.ca/artifactory/miso-snapshot</url>
```
djcooke: We don't push snapshots, so this can be removed

PR #1896, opened 2019-04-15
Avoid sending lists of IDs by query parameters

apmasell:
This includes spreadsheet downloads and printing

JIRA: GC-6778


callunity:
Is it possible to add some sort of "Working..." indicator,maybe by keeping the dialog open until the request finishes? I tried downloading a Tracking List of 1000 dilutions and during the 22 seconds it took for the response to complete, there was no indication that anything was actually happening.

apmasell:
Added a dialog.

PR #1898, opened 2019-04-17
GLT-2818: fixed container serial number uniqueness validation

djcooke:


callunity:
Could the release notes please be updated to indicate that this could be a breaking change if your SequencerPartitionContainer table has duplicate identificationBarcodes?

djcooke:
@callunity: added

djcooke:
@sklages: Since you have some pretty old data, you'll definitely want to make sure you don't have any duplicate container serial numbers before upgrading to the release that includes this change. You can check with this query:

```
SELECT identificationBarcode, COUNT(containerId)
FROM SequencerPartitionContainer
GROUP BY identificationBarcode
HAVING COUNT(containerId) > 1;
```

If you have duplicates, it would be ideal to delete the duplicates. If you want to keep the data for both containers, you could also change the serial number for one of them. If you have any trouble with this, please open an issue and we'd be glad to help

sklages:
`0 row(s) returned` :+1: 

Thanks for the hint.

PR #1901, opened 2019-04-25
GLT-2834: added single Edit Dilution page

djcooke:


docs/user_manual/library_dilutions.md
```diff
@@ -50,10 +50,40 @@ adjustments and click the "Save" button to try again.
...
+
+{% include userman-toplink.md %}
+
+
+
+{% assign subsub = 1 %}
+{% include userman-heading.md section=page.section sub=sub subsub=subsub title="Editing a Single Dilution" %}
+
+To get to the single Edit Dilution page, click on the dilution's name in the Dilutions list. Links to the Edit
+Dilution page also appaear on the Edit Pool page in the Included and Available Dilutions lists.
```
avarsava: appaear -> appear

miso-web/src/main/webapp/scripts/list_dilution.js
```diff
@@ -83,11 +83,14 @@ ListTarget.dilution = {
   createColumns: function(config, projectId) {
     return [{
       "sTitle": "Name",
-      "mData": "id",
+      "mData": "name",
```
callunity: Can this be changed back to `id`? Using `name` causes the column to be sorted alphabetically (1,10,100,1000,10000,10001, etc.)

docs/user_manual/library_dilutions.md
djcooke: fixed

miso-web/src/main/webapp/scripts/list_dilution.js
djcooke: Fixed. Added comment to explain

djcooke:
@callunity:

* Dilution search terms are outside the scope of this ticket. Created new ticket to address that: https://jira.oicr.on.ca/browse/GLT-2835
* Saving dilutions with targeted sequencing works fine. The specific dilution that you used belongs to a library with a kit that requires targeted sequencing. Another of the library's dilutions does not have a targeted sequencing value set. When a dilution is saved, its library is also saved in case of volume-related changes. The library fails to save in this case. This behaviour is the same as on the bulk edit page before this patch, though there it shows a validation error for the `libraryDesignCode` field which doesn't really make sense since that's not a dilution field. I've modified the dilution service to rethrow the library ValidationExceptions in a way that causes the messages to show up as general validation errors (at the top) and include "Library" at the start for clarity. e.g. "Library libraryDesignCode: Targeted sequencing must be assigned to the affected dilutions (LDI9319) to use this library design code: EX"

PR #1913, opened 2019-05-02
GLT-2856: fixed reverse compliment indices in Pinery

djcooke:


djcooke:
@callunity I think you did the diff backwards from what you intended? That output shows that the index is properly mangled in the PR (top), but not in prod (bottom)

callunity:
Ah thanks, I had reverse-complemented my understanding of what Pinery should represent when.

PR #1918, opened 2019-05-06
GLT-2847: converted Edit Kit Descriptor page to use form builder

djcooke:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultKitService.java
```diff
@@ -113,26 +114,46 @@ public long saveKit(Kit kit) throws IOException {
...
     kitDescriptor.setChangeDetails(authorizationManager.getCurrentUser());
     return kitStore.saveKitDescriptor(kitDescriptor);
   }
 
+  private void validateChange(KitDescriptor kitDescriptor, KitDescriptor beforeChange) throws IOException {
+    List<ValidationError> errors = new ArrayList<>();
+
+    if (isSetAndChanged(KitDescriptor::getName, kitDescriptor, beforeChange)) {
+      if (kitStore.getKitDescriptorByName(kitDescriptor.getName()) != null) {
+        errors.add(new ValidationError("name", "There is already a kit descriptor with this name"));
```
callunity: What do you think about removing the term "kit descriptor" in user-facing text sicne that distinction was originally for inventory purposes, which is now being handled by RAMEN?

djcooke: We still haven't decided what the interaction between MISO and RAMEN will look like, but it will likely require making the distinction between kits and kit descriptors

PR #1922, opened 2019-05-08
Add instructions on enabling SSL

morgantaschuk:
GLT-2831


.docker/nginx/self-sign-cert.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
+
+# only install openssl if generating a self-signed certificate
+apk add openssl
+
+# relies on the file at /run/secrets/ssl_password that has the SSL password
+mkdir -p /etc/nginx/ssl && cd /etc/nginx/ssl
+openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365 \
+    -passout file:/run/secrets/ssl_password \
+    -subj "/C=CA/ST=ON/L=Toronto/O=OICR/OU=Genome Sequence Informatics"
```
djcooke: Should these be configurable, or non-oicr/gsi-specific? Or maybe just document that "you might want to modify the openssl subject parameters"

morgantaschuk: Fixed.

PR #1933, opened 2019-05-15
GLT-2846: converted Edit Instrument page to use form builder

djcooke:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultInstrumentService.java
```diff
@@ -58,25 +54,43 @@ public Instrument getByUpgradedInstrumentId(long instrumentId) throws IOExceptio
...
   }
 
+  private void validateChange(Instrument instrument, Instrument beforeChange) throws IOException {
+    List<ValidationError> errors = new ArrayList<>();
+
+    if (ValidationUtils.isSetAndChanged(Instrument::getName, instrument, beforeChange) && getByName(instrument.getName()) != null) {
+      errors.add(new ValidationError("name", "There is already an instrument with this name"));
+    }
+    if (ValidationUtils.isSetAndChanged(Instrument::getUpgradedInstrument, instrument, beforeChange)
+        && getByUpgradedInstrumentId(instrument.getUpgradedInstrument().getId()) != null) {
```
callunity: This line throws a NPE when creating a new instrument

djcooke: Fixed. `ValidationUtils#isSetAndChanged` was returning true when the field was _not_ set if `beforeChange` was null (on create)

PR #1935, opened 2019-05-16
GLT-2850: converted Edit Study page to use form builder

djcooke:

PR #1938, opened 2019-05-17
GLT-2850: converted Edit Study page to use form builder (#1935)

djcooke:


djcooke:
Had weird rebase issues... new PR is https://github.com/miso-lims/miso-lims/pull/1939

PR #1939, opened 2019-05-17
GLT-2844: converted Edit Freezer page to use form builder

djcooke:


djcooke:
@callunity: fixed and tests added

PR #1940, opened 2019-05-17
GLT-2851: converted Edit Submission page to use form builder

djcooke:


RELEASE_NOTES.md
```diff
@@ -5,6 +5,8 @@ Changes:
   * Changed Edit Study page to show validation messages instead of error page when saving fails
   * Changed Edit Instrument page to show validation messages instead of error page when saving
     fails
+  * Changed Edit Submission page to show validation messages instead of error page when saving
+    fails
```
avarsava: Maybe these could just be one long Change? eg 'Changed Edit Submission, Instrument, Study, [...] to show validation messages instead of error page when saving fails'

djcooke: Yeah, would be better that way, though it makes rebasing a little more painful. I'm lead next week and will condense during release

miso-web/src/main/webapp/scripts/form_submission.js
```diff
@@ -0,0 +1,68 @@
...
+          title: 'Description',
+          data: 'description',
+          type: 'text',
+          maxLength: 255
+        }, {
+          title: 'Accession',
+          data: 'accession',
+          type: 'read-only',
+          getLink: function(sample) {
+            return 'http://www.ebi.ac.uk/ena/data/view/' + sample.accession;
```
apmasell: MISO was originally designed for the archive to be swappable with another one. I don't think there are any others in existence in the real world and I ripped out most the code that, so, I think it's fine to hard code these URLs. If you see anything that relates to a swappable archive, I say, rip it out.

djcooke: There's also no way of setting the accession on anything via the UI. This is how it was before too though.

PR #1945, opened 2019-05-22
Pick up MinKNOW and Protocol Versions

avarsava:


miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -2172,6 +2175,14 @@ public static Run to(@Nonnull NotificationDto from) {
     return to;
   }
 
+  private static void setOxfordNanoporeRunValues(@Nonnull OxfordNanoporeNotificationDto from,
+                                                 @Nonnull OxfordNanoporeRun to){
+    // software version is "minknowVersion + protocolVersion"
+    String[] softwareVersions = from.getSoftware().split("\\ \\+\\ ");
```
callunity: Rather than splitting strings in the client, might it make sense to change the run scanner's ONT DTO to have protocol version and minknow version as separate fields?

avarsava: It's grouped together in the RS DTO because 'software [version]' is a field of the abstract NotificationDto class, and the sync between the two version numbers is unclear. (e.g.,  if two DTOs have the same minknow version but different protocol versions, and someone groups only by minknow version, what differences in the data might go unnoticed?) We could add just a protocol version field to the ONT DTO and use 'software' to store the minKnow version, but I'd want to ask whether anyone could foresee any consequences to doing that

apmasell: There are no consequences. That single software version was mostly meant to be informative for a human looking at the DTO to know which version it was and possibly compare results between different versions. If we really care about these fields, it makes sense to have them as separate fields.

callunity: The `RunOxfordNanopore` table in the db has `minKnowVersion` and `protocolVersion` as separate fields, which seems like a 'we care about this' decision to me.. but perhaps we could report these separately and leave the `software` field empty? In the IlluminaRun DTO it's there to mark the version of the software (InterOp) that run scanner is using to generate its numbers, as the reported numbers could vary even if the data on disk stays the same. As I understand it, we're parsing the Oxford Nanopore data with the HDF5Viewer and not an Oxford Nanopore-specific library, so the data we report won't change unless the data on disk changes.

avarsava:
As per discussion, MinKNOW and Protocol versions will be split in Run Scanner. See PR: https://github.com/miso-lims/runscanner/pull/56

PR #1951, opened 2019-05-27
Fix Travis badge

djcooke:


djcooke:
Will use "Squash and merge" and improve commit message during merge

PR #1952, opened 2019-05-27
GLT-2813: added interface for modifying stains and stain categories

djcooke:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleTypeService.java
```diff
@@ -70,7 +70,8 @@ private void validateChange(SampleType sampleType, SampleType beforeChange) thro
       }
       long usage = sampleTypeDao.getUsage(beforeChange);
       if (usage > 0L) {
-        errors.add(new ValidationError("name", "Cannot change name of sample type because it is already in use by " + usage + " samples"));
+        errors
+            .add(new ValidationError("name", "Cannot change name of sample type because it is already in use by " + usage + " sample(s)"));
```
avarsava: DefaultStainCategoryService calculates whether to add the 's' or not, could this one too?

djcooke: fixed

PR #1953, opened 2019-05-28
GLT-2813: added interface for modifying tissue types

djcooke:


djcooke:
@callunity: rebased and added tests

PR #1957, opened 2019-05-30
GLT-2815: added interface for modifying box sizes and uses

djcooke:


migration/src/main/java/uk/ac/bbsrc/tgac/miso/migration/destination/MisoServiceManager.java
```diff
@@ -301,1407 +439,935 @@ public void setUpSecurityContext(User migrationUser) {
...
-  // TODO: Add naming scheme config instead of hard-coding
   public NamingScheme getNamingScheme() {
-    if (namingScheme == null) {
-      namingScheme = new OicrNamingScheme();
+    return getService(NamingScheme.class);
+  }
+
+  private <T> T getService(Class<T> clazz) {
+    @SuppressWarnings("unchecked")
+    T service = (T) services.get(clazz);
```
apmasell: You can remove the unchecked warning by doing:
```
T service = clazz.cast(services.get(clazz));
```

djcooke: Interesting... I feel like `Class#cast` should cause the same compiler warning. Changed anyway :man_shrugging:

PR #1978, opened 2019-06-13
GLT-2814: added interface for modifying library types

djcooke:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryTypeService.java
```diff
@@ -0,0 +1,107 @@
...
+        || ValidationUtils.isSetAndChanged(LibraryType::getDescription, object, beforeChange))
+        && libraryTypeDao.getByPlatformAndDescription(object.getPlatformType(), object.getDescription()) != null) {
+      errors.add(new ValidationError("description",
+          "There is already a library type with this description for platform: " + object.getPlatformType().getKey()));
+    }
+    long libUsage = libraryTypeDao.getUsageByLibraries(beforeChange);
+    long tempUsage = libraryTypeDao.getUsageByLibraryTemplates(beforeChange);
+    if ((libUsage > 0L || tempUsage > 0L) && ValidationUtils.isSetAndChanged(LibraryType::getPlatformType, object, beforeChange)) {
+      errors.add(new ValidationError("platform",
+          String.format("Cannot be changed because library type is already used by %d librar%s and %d library template%s", libUsage,
```
apmasell: I feel like there should be a better way to build these strings. Maybe there should be a static method in every interface that takes a bool and returns the singular or plural or an enum with both values and then a function that can generate the string from count+noun maker.

Just a thought.

callunity: Or something like Rails's `pluralize`, where `Library.pluralize(numLibs)` gets you the correct form...
```
String.format("Library type is already associated with %d %s and %d %s",
 libUsage, Library.pluralize(libUsage), tempUsage, LibraryTemplate.pluralize(tempUsage));
```

djcooke: I'd rather it wasn't all over the place (in all of the interfaces/models). How about a `Pluralizer` class with static methods like `Pluralizer.libraries(count)`?

djcooke:
@callunity: That works. Changed

Issue #1976, opened 2019-06-13
ERROR: no such image: misolims/miso-lims-migration:: invalid reference format


AKWB:
Hi there,

I am not a docker specialist (yet). I followed the instructions:
wget https://github.com/miso-lims/miso-lims/archive/master.zip
unzip master.zip 'miso-lims-master/.docker/*'
mv miso-lims-master/.docker miso-lims-compose
rm -r master.zip miso-lims-master/
cd miso-lims-compose
export MISO_DB_USER=tgaclims && export MISO_DB=lims && export MISO_DB_PASSWORD_FILE=./.miso_db_password && MISO_TAG=latest
echo "changeme" > ./.miso_db_password
docker-compose -f demo.plain.yml up

However it ends with:
WARNING: The MISO_TAG variable is not set. Defaulting to a blank string.
Creating network "miso-lims-compose_default" with the default driver
Pulling db (mysql:5.7.25)...
5.7.25: Pulling from library/mysql
27833a3ba0a5: Pull complete
864c283b3c4b: Pull complete
cea281b2278b: Pull complete
8f856c14f5af: Pull complete
9c4f38c23b6f: Pull complete
1b810e1751b3: Pull complete
5479aaef3d30: Pull complete
1d924ec3d520: Pull complete
1ab7ae63ac60: Pull complete
08aa5f3680e9: Pull complete
a832d0a0972a: Pull complete
Digest: sha256:dba5fed182e64064b688ccd22b2f9cad4ee88608c82f8cff21e17bab8da72b81
Status: Downloaded newer image for mysql:5.7.25
ERROR: no such image: misolims/miso-lims-migration:: invalid reference format

Any suggestion?

djcooke:
Looks like an error in our instructions. The `MISO_TAG` variable should be exported along with the rest. Replace the export line with this:

```
export MISO_DB_USER=tgaclims && export MISO_DB=lims && export MISO_DB_PASSWORD_FILE=./.miso_db_password && export MISO_TAG=latest
```

AKWB:
seems to work...

Now I have a warning:
flyway_1  | WARNING: DB: Unknown table 'lims.CompletedPartitions' (SQL State: 42S02 - Error Code: 1051)
miso-lims-compose_flyway_1 exited with code 0

I can open http://localhost/miso, login with admin/admin and get:
HTTP Status 404 – Not Found
Type Status Report
[...]

PR #1979, opened 2019-06-13
#1976: added redirect from /miso/ to mainMenu; fixed docs

djcooke:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/MenuController.java
```diff
@@ -248,6 +248,12 @@ public ModelAndView myAccountMenu(ModelMap model) {
     }
   }
 
+  @GetMapping("/")
+  public ModelAndView redirectMisoRoot(ModelMap model) {
+    model.clear();
+    return new ModelAndView("redirect:mainMenu", model);
+  }
```
callunity: Can this also be applied to endpoint `""`? `http://localhost/miso` without the trailing slash still throws the 404.

djcooke: I tried, but it doesn't work. I don't think there's a way to map `/miso` because of the weird way that that portion of the url is forced in. I do think that a redirect from `/miso/` makes more sense than redirecting from `/miso` anyway though

callunity: Yeah, that weird URL forcing is regrettable. I do think that people are more likely to try to go to `/miso` than `/miso/`, but I don't think there's anything to be done about that. :disappointed:

PR #1981, opened 2019-06-17
GLT-2752: renamed orders to sequencing orders

djcooke:


miso-web/src/main/webapp/scripts/list_completion.js
```diff
@@ -22,12 +22,12 @@
  */
 
 ListTarget.completion = {
```
apmasell: This should get renamed to `sequencingordercompletion` for symmetry with everything else.

djcooke: fixed

PR #1983, opened 2019-06-20
removed Maven 'external' profile

djcooke:


djcooke:
Forgot to delete commented lines. Will do before merging

miso-web/pom.xml
```diff
@@ -316,10 +316,27 @@
...
+              </excludes>
+            </resource>
+          </webResources>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
   <profiles>
-    <profile>
+    <!-- <profile>
```
apmasell: :fire:

djcooke: removed

PR #1986, opened 2019-06-21
GLT-2882: added poolId to Pinery RunPosition

djcooke:
Pinery version needs updated to a release before merging


djcooke:
Moved to https://github.com/miso-lims/miso-lims/pull/1994

PR #2000, opened 2019-07-10
GLT-2905: Fix LDAP Things

djcooke:


djcooke:
@avarsava Not really, but changing it would involve renaming the LDAP role and require anyone else who uses MISO with LDAP/AD auth to do the same, which I don't think is worth the effort

PR #2009, opened 2019-07-19
[GLT-2916] Patch 'No Scientific Name' bug

avarsava:
Putting periods in field names should cause Handsontable to create nested objects and fields. In rare (and as of yet unknown) circumstances, it fails to do this, and consistently does so on flat.sample.scientificName, creating a situation where on save there is an extra field flat["sample.scientificName"] which contains a string but flat.sample.scientificName is null. Detects when this happens and forces the value back to the correct place to enable saving. Does not fix the underlying issue. Can't seem to pinpoint it, this broken field seems to pop up at the call to pack.

Also corrects a typo in an unrelated function. Don't know if that's ever called.

### DIFF
--- a/miso-web/src/main/webapp/scripts/hot_library_receipt.js
+++ b/miso-web/src/main/webapp/scripts/hot_library_receipt.js
@@ -34,13 +34,23 @@ HotTarget.libraryReceipt = (function() {
           if (!flat.sample) {
             flat.sample = {};
           }
+
           col.libraryUnpack(lib.sample, flat.sample, setCellMeta);
+
+
         };
         col.libraryPack = col.pack;
         col.pack = function(lib, flat, errorHandler) {
           if (!flat.sample) {
             flat.sample = {};
           }
+
+          // Sometimes sample properties don't go into the nested subobject, but become a field called
+          // "sample.[name]". If this happens, it breaks validation on save. Move it back.
+          if(Object.keys(flat).filter(function(s){return s === col.data}).length > 0){
+            eval("flat.sample." + col.data.substring(7) + " = '" + flat[col.data] + "';");
+          }
+
           col.libraryPack(lib.sample, flat.sample, errorHandler);
         };
         if (col.update) {
--- a/miso-web/src/main/webapp/scripts/hot_library_receipt.js
+++ b/miso-web/src/main/webapp/scripts/hot_library_receipt.js
@@ -30,30 +30,28 @@ HotTarget.libraryReceipt = (function() {
             col.depends = 'sample.' + col.depends;
           }
         }
+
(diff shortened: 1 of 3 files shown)


miso-web/src/main/webapp/scripts/hot_library_receipt.js
```diff
@@ -34,13 +34,23 @@ HotTarget.libraryReceipt = (function() {
...
         };
         col.libraryPack = col.pack;
         col.pack = function(lib, flat, errorHandler) {
           if (!flat.sample) {
             flat.sample = {};
           }
+
+          // Sometimes sample properties don't go into the nested subobject, but become a field called
+          // "sample.[name]". If this happens, it breaks validation on save. Move it back.
+          if(Object.keys(flat).filter(function(s){return s === col.data}).length > 0){
```
callunity: `flat.hasOwnProperty[col.data]` is more idiomatic for JavaScript

miso-web/src/main/webapp/scripts/hot_library_receipt.js
```diff
@@ -34,13 +34,23 @@ HotTarget.libraryReceipt = (function() {
...
+            eval("flat.sample." + col.data.substring(7) + " = '" + flat[col.data] + "';");
```
callunity: Rather than `eval` and substringing, does it work if you:
1) save the initial value of `col.data` up at line 22 to a variable? `var initialColData = Object.assign(col.data).toString();`
2) use that for the assignment here on line 51? `flat.sample[initialColData] = flat[col.data];`

miso-web/src/main/webapp/scripts/hot_library_receipt.js
avarsava: Fixed

miso-web/src/main/webapp/scripts/hot_library_receipt.js
avarsava: Fixed

miso-web/src/main/webapp/scripts/hot_library_receipt.js
```diff
@@ -34,13 +35,23 @@ HotTarget.libraryReceipt = (function() {
           if (!flat.sample) {
             flat.sample = {};
           }
+
           col.libraryUnpack(lib.sample, flat.sample, setCellMeta);
+
+
```
callunity: Stray whitespace?

avarsava: evened it out

djcooke:
Please add to release notes and include ticket number in commit message when squashing

PR #2016, opened 2019-07-26
Make minimum edit distance editable

callunity:
GLT-2931
*  Allows the "duplicate indices" and "near-duplicate indices" bases to be set by an administrator. OICR will soon be erroring on edit distance <= 2; this allows other administrators to set their own thresholds.
*  The Index Distance Tool informs the user of the default edit distance for duplicate & near-duplicate indices, but uses only the user-supplied edit distance when doing its calculations.
*  Moves (near-)duplicate indices calculations into the DTO conversions since it's only used to display information to the user.


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditPoolController.java
```diff
@@ -99,6 +100,10 @@
   private ContainerService containerService;
   @Autowired
   private BoxService boxService;
+  @Value("${miso.error.edit.distance:2}")
```
apmasell: I kind of hate that the defaults get smeared everywhere, but no suggestions for a better way, so, :man_shrugging:

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditPoolController.java
```diff
@@ -128,20 +133,24 @@ public ModelAndView setupForm(@PathVariable Long poolId, ModelMap model) throws
     }
 
     if (pool == null) throw new NotFoundException("No pool found for ID " + poolId.toString());
-
+    PoolDto poolDto = Dtos.asDto(pool, true, false, errorEditDistance, warningEditDistance);
```
apmasell: The original didn't call this when it was unsaved. I'm not sure it matters either way... I'm assuming you've checked the behaviour.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/IndexDistanceToolController.java
```diff
@@ -87,20 +134,50 @@ public ModelAndView getTool(ModelMap model) {
...
     }
 
-    return results;
+    IndexDistanceResponseDto response = new IndexDistanceResponseDto(results, shortestIndexLength);
+    return response;
+  }
+
+  public static int getShortestIndexLength(List<String> indices) {
+    String initialValue = "REALLY_LONG_VALUE_THAT_WILL_SURELY_GET_REPLACED_BUT_HERE_ARE_SOME_MORE_CHARACTERS_JUST_IN_CASE";
+    String shortestIndex = indices.stream()
```
apmasell: You can do `indicies.stream().filter(Objects::nonNull).mapToInt(String::length).min().getAsInt()`.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/IndexDistanceToolController.java
```diff
@@ -87,20 +134,50 @@ public ModelAndView getTool(ModelMap model) {
...
+          if (current != null && current.length() > 0 && current.length() < shortest.length()) return current;
+          return shortest;
+        });
+    if (shortestIndex == null || initialValue.equals(shortestIndex)) return 0;
+    return shortestIndex.length();
+  }
+
+  public static String flattenIndices(List<Index> indices) {
+    return indices.stream()
+        .sorted((i1, i2) -> Integer.compare(i1.getPosition(), i2.getPosition()))
```
apmasell: `.sorted(Comparator.comparingInt(Index::getPosition))`

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditPoolController.java
callunity: I did check it, but I checked again and found a bug in the "Pool custom" functionality so thank you :laughing:

callunity:
I'll be refactoring this PR to do the calculations outside of the Dtos class, as we'll need to be able to calculate this prior to serialization for thingsl ike Pool Orders.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditPoolController.java
callunity: I'm going to extract it into a microcontroller because we need to use the logic outside of just the display.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/PoolImpl.java
```diff
@@ -170,6 +169,12 @@
   @Transient
   private List<FileAttachment> pendingAttachmentDeletions;
 
+  @Transient
+  private Set<String> duplicateIndicesSequences;
+
+  @Transient
+  private Set<String> nearDuplicateIndicesSequences;
```
djcooke: It doesn't seem like these should be properties of the pool (or pool view) itself, as that will lead to confusion since they're usually not set. In every case where they're used in this PR, the pool's near/duplicate indices are set, and then the pool is passed to something else (DTO conversion) that actually cares about them. Instead, you could just include the DuplicateIndicesChecker as a parameter of the DTO conversion method. This would also prevent a lot of duplicated code and having to remember to set the near/duplicate indices every time they're going to be needed by something else

miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/PoolPageIT.java
```diff
@@ -291,9 +291,9 @@ public void testRemoveAliquots() {
 
   @Test
   public void testLibraryAliquotTableWarnings() {
-    testLibraryAliquotTableWarningOnPoolWithError(801L, "(NO INDEX)");
-    testLibraryAliquotTableWarningOnPoolWithError(802L, "(NEAR-DUPLICATE INDEX)");
-    testLibraryAliquotTableWarningOnPoolWithError(803L, "(DUPLICATE INDEX)");
+    testLibraryAliquotTableWarningOnPoolWithError(801L, "(MISSING INDEX)");
+    testLibraryAliquotTableWarningOnPoolWithError(802L, "(NEAR-DUPLICATE INDICES)");
```
djcooke: This was changed to lower-case in some places, but not others?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/IndexDistanceToolController.java
```diff
@@ -87,20 +137,51 @@ public ModelAndView getTool(ModelMap model) {
...
+  }
+
+  public static String flattenIndices(List<Index> indices) {
+    return indices.stream()
+        .sorted(Comparator.comparingInt(Index::getPosition))
+        .map(Index::getSequence)
+        .collect(Collectors.joining("-"));
+  }
+
+  public static String truncate(String index, int shortestLength) {
```
djcooke: These 3 methods should be private, or if useful outside of this controller, they should be in a utilities class

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/component/DuplicateIndicesChecker.java
```diff
@@ -0,0 +1,107 @@
...
+    return getIndexSequencesWithMinimumEditDistance(indices, warningEditDistance);
+  }
+
+  public static Set<String> getIndexSequencesWithMinimumEditDistance(Pool pool, int minAllowableEditDistance) {
+    if (pool == null) return Collections.emptySet();
+    List<List<Index>> indices = getIndexSequences(pool);
+    return getIndexSequencesWithMinimumEditDistance(indices, minAllowableEditDistance);
+  }
+
+  public static Set<String> getIndexSequencesWithMinimumEditDistance(List<List<Index>> indices, int minAllowableEditDistance) {
```
djcooke: Do these need to be public?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ExperimentRestController.java
```diff
@@ -124,23 +124,26 @@
...
-    return RestUtils.getObject("Experiment", experimentId, experimentService, Dtos::asDto);
+    return RestUtils.getObject("Experiment", experimentId, experimentService,
+        expt -> Dtos.asDto(expt));
   }
 
   @GetMapping(produces = "application/json")
   public @ResponseBody List<ExperimentDto> list() throws IOException {
     Collection<Experiment> experiments = experimentService.list();
-    return experiments.stream().map(Dtos::asDto).collect(Collectors.toList());
+    return experiments.stream().map(expt -> Dtos.asDto(expt)).collect(Collectors.toList());
```
djcooke: You've made changes similar to this in a few places... If your compiler was complaining, you might want to update your Java version. I was previously using a version that had a similar bug too

miso-web/src/main/webapp/WEB-INF/pages/indexDistanceTool.jsp
```diff
@@ -62,7 +62,7 @@
     </div>
     <div style="display: table-cell; width: 100px; vertical-align: bottom; padding-left: 12px; padding-right: 12px;">
       <label>Min. Distance:<br>
-        <input id="minDistance" type="text" value="2"
+        <input id="minDistance" type="text" value="3"
```
djcooke: This should default to the warning distance +1

miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -696,12 +696,14 @@
...
-      if (prefix == "SAM" || prefix == "LIB" || prefix.substring(1, 3) == "PO") {
+      if (prefix == "SAM" || prefix == "LIB" || prefix == "LDI" || prefix.substring(1, 3) == "PO") {
         if (prefix == "SAM") {
           path += "sample/";
         } else if (prefix == "LIB") {
           path += "library/";
-        } else {
+        } else if (prefix == "LDI") {
+          path += "libraryaliquot/";
+        } else if (prefix == "IPO") {
```
djcooke: Surrounding `if` condition should be changed to match this (`IPO` instead of `?PO`)... Or refactor to get rid of the surrounding `if` statement

miso-web/src/main/webapp/scripts/warning_partition.js
```diff
@@ -36,21 +36,21 @@ WarningTarget.partition = {
           level: "important"
         }, {
           include: pool.duplicateIndices,
-          tableMessage: "(DUPLICATE INDICES)"
+          tableMessage: Constants.warningMessages.duplicateIndices
         }, {
           include: pool.nearDuplicateIndices && !pool.duplicateIndices,
-          tableMessage: "(NEAR-DUPLICATE INDICES)"
+          tableMessage: Constants.warningMessages.nearDuplicateIndices
```
djcooke: Parentheses have been removed from some warning messages, but not others. I'd be in favour of removing them from all - I think it's enough that they're coloured differently, and are usually in their own column too. Either way, they should be consistent

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ExperimentRestController.java
callunity: Ah, this is left over from when I had to expand it to pass in the error and warning distances. Will revert.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/PoolImpl.java
callunity: I'd prefer not to include the DuplicateIndicesChecker as a parameter of the PoollDto conversion method, as it results in something like [this[(https://github.com/miso-lims/miso-lims/pull/2016/commits/e57bd26a5e725882c10ce86cc02e3b092d66020a) where the indexChecker gets wired into 12 different controllers, several of which don't even need it. Can we brainstorm about alternatives?

RELEASE_NOTES.md
```diff
@@ -2,6 +2,11 @@
 
   * Fixed bug that broke List Freezers page
   * Show _Assignee_ column in _Issue_ tables
+  
+`miso.properties` changes:
```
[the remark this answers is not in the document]
callunity: Done

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/PoolImpl.java
callunity: Added an IndexChecker as a parameter of the DTO conversion method.

miso-web/src/main/webapp/scripts/box_visualization.js
callunity: Done.

miso-web/src/main/webapp/scripts/warning_partition.js
callunity: Updated all to remove parentheses.

RELEASE_NOTES.md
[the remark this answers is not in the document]
djcooke: Please use styling consistent to the rest of the document

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -1687,14 +1685,18 @@ public static PoolDto asDto(@Nonnull Pool from, boolean includeContents, boolean
       }
       dto.setPooledElements(pooledElements);
 
-      dto.setDuplicateIndicesSequences(from.getDuplicateIndicesSequences());
+      dto.setDuplicateIndicesSequences(indexChecker.getDuplicateIndicesSequences(from));
       dto.setDuplicateIndices(dto.getDuplicateIndicesSequences() != null && !dto.getDuplicateIndicesSequences().isEmpty());
-      dto.setNearDuplicateIndicesSequences(from.getNearDuplicateIndicesSequences());
+      dto.setNearDuplicateIndicesSequences(indexChecker.getNearDuplicateIndicesSequences(from));
```
djcooke: Missing null checks on `indexChecker` here and above

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/component/DuplicateIndicesChecker.java
```diff
@@ -0,0 +1,126 @@
...
+import uk.ac.bbsrc.tgac.miso.core.data.impl.view.ListPoolView;
+import uk.ac.bbsrc.tgac.miso.core.data.impl.view.ListPoolViewElement;
+import uk.ac.bbsrc.tgac.miso.core.data.impl.view.PoolElement;
+import uk.ac.bbsrc.tgac.miso.core.data.impl.view.PoolableElementView;
+import uk.ac.bbsrc.tgac.miso.dto.IndexChecker;
+
+@Component
+public class DuplicateIndicesChecker implements IndexChecker {
+
+  @Value("${miso.pools.error.index.mismatches:0}")
```
djcooke: Default to 1 as in `miso.properties`

miso-web/src/main/webapp/scripts/indexDistance_ajax.js
```diff
@@ -36,14 +36,19 @@
         data: JSON.stringify(data)
       }).success(function(data) {
         var results = '';
-        if (!data || !data.length) {
+        if (data.warningEditDistance > data.errorEditDistance) {
+          results += 'Indices with edit distance of less than ' + Constants.warningEditDistance + ' are considered near-matches.\n';
```
djcooke: Should be "...distance of less than x+1" or "...distance of x or less". Same below

miso-web/src/main/webapp/scripts/indexDistance_ajax.js
```diff
@@ -36,14 +36,19 @@
...
+        }
+        results += 'Indices with edit distance of less than ' + Constants.errorEditDistance + ' are considered DUPLICATES.\n\n';
+        results += 'Indices were compared at the length of the shortest Index 1, which is ' + data.shortestIndexLength + ' bp.\n';
```
djcooke: Add an extra line break to separate from near/duplicate listing

miso-web/src/main/webapp/scripts/indexDistance_ajax.js
callunity: I find the "edit distance" terminology to be confusing, and I'd like to reword it to "Indices with x or fewer mismatches" and change the page's "minimum edit distance" to "minimum mismatches". What do y'all think?

djcooke: I don't have a strong opinion either way. Just make sure the User Manual gets updated too

RELEASE_NOTES.md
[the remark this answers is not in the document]
callunity: Done.

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
callunity: Added

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/component/DuplicateIndicesChecker.java
callunity: Done in IndexChecker.

miso-web/src/main/webapp/scripts/indexDistance_ajax.js
callunity: Updated. PR: https://github.com/miso-lims/walkthroughs/pull/90

miso-web/src/main/webapp/scripts/indexDistance_ajax.js
callunity: Updated.

miso-web/src/main/webapp/scripts/indexDistance_ajax.js
djcooke: That's good too, but I meant the User Manual which is in `/docs` of this repo

djcooke: Ah, I see you got that already too though :+1:

RELEASE_NOTES.md
```diff
@@ -13,11 +13,18 @@ Changes:
 
 Upgrade Notes:
 
-`miso.properties` changes:
-  * `miso.pools.strictIndexChecking` now present in `miso.properties`. Default value is false, if
-    property is not found 'false' is assumed.
-  * Number of mismatches to trigger duplicate & near-duplicate indices messages are now configurable
-    * Messages for duplicate & near-duplicate indices are now configurable
+  * There have been changes to the settings in `security.properties`. You may need to make
```
djcooke: `security.properties`?

callunity: Copypasta. Fixed.

PR #2017, opened 2019-07-26
GLT-2752: added pool orders

djcooke:


miso-web/src/main/webapp/scripts/poolorder_ajax.js
```diff
@@ -0,0 +1,50 @@
...
+
+    removeAliquots: function(aliquotIds) {
+      var aliquots = PoolOrder.getAliquots().filter(function(orderAli) {
+        return aliquotIds.indexOf(orderAli.aliquot.id) === -1;
+      });
+      PoolOrder.setAliquots(aliquots);
+    },
+
+    updateProportions: function() {
+      // TODO
```
apmasell: ?

miso-web/src/main/webapp/scripts/list_orderaliquot.js
```diff
@@ -0,0 +1,149 @@
...
+    }, "aliquot.sampleAlias", 0, true), {
+      'sTitle': 'Indices',
+      'mData': 'aliquot.indexIds',
+      'include': true,
+      'bSortable': false,
+      'iSortPriority': 0,
+      'mRender': function(data, type, full) {
+        var indices = Constants.indexFamilies.reduce(function(acc, family) {
+          return acc.concat(family.indices.filter(function(index) {
+            return data.indexOf(index.id) != -1;
```
apmasell: If you try to add a library aliquot from a library that has no indices, it will explode here and the jQuery table locks up. Maybe there should be a filter at the previous step that disallows any indexless entries to be added.

miso-web/src/main/webapp/scripts/poolorder_ajax.js
djcooke: Was a place-holder that I didn't end up using. Removed

miso-web/src/main/webapp/scripts/list_orderaliquot.js
djcooke: It should be possible to create a pool order for a single aliquot with no indices so I've just changed it to show "(none)" instead. There is a separate ticket for disallowing index collisions in a pool order, which should also prevent multiple libraries with missing indices

docs/user_manual/pool_orders.md
```diff
@@ -0,0 +1,119 @@
...
+---
+
+Pool orders are used to track sets of library aliquots that are ready to be pooled. It may be useful for people who are
+responsible for creating libraries to then create a pool orders to pass requirements on to other people who are
+responsible for creating pools. Pool orders are intended to help organize separate teams, and using them is completely
+optional.
+
+A pool order specifies which library aliquots should be pooled, and at what proportions. To fulfill a pool order, a
+pool containing all of the specified aliquots must be linked to it. If the pool contains an aliquot derived from the
+specified aliquot, that is also accepted. A pool can be used even if it contains additional aliquots, or if the
```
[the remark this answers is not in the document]
djcooke: Changed to "A pool can be used to fulfill the pool order even if..."

djcooke:
@callunity Apparently Hibernate uses an inner join by default for any relationship for which you create an alias on the `Criteria`. You can use a different `createAlias` method to specify the join type, but since we've abstracted that, it's not straightforward for us to do so. This alias was only needed for sorting purposes though, and I don't think sorting on the sequencing parameters column is real useful, so I made it not sortable and removed the alias. The lists now populate correctly.

miso-web/src/main/webapp/WEB-INF/log4j.miso.properties
```diff
@@ -34,3 +34,6 @@ log4j.appender.debugFileAppender.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss
 
 #log4j.rootLogger=INFO, stdout
 log4j.rootLogger=INFO, debugFileAppender
+
+#log4j.logger.org.hibernate.SQL=DEBUG
+#log4j.logger.org.hibernate.type=TRACE
```
callunity: Stray commented-out properties?

djcooke: No, left intentionally so I don't have to look it up every time

PR #2022, opened 2019-07-31
[GLT-2923] Prevent creation of sequencing orders when pool contains bad indices

avarsava:


avarsava:
> If we're allowing merging pools that result in near/duplicate indices, should creating a sequencing order for them also be allowed? If so, then I don't think we can really prevent creating sequencing orders due to near/duplicate indices. A warning would have to be sufficient

Not creating sequencing orders on Pools with duplicate indices is one of the conclusions we came to in our discussion on 16/07/2019. We could discuss this again as a group

avarsava:
Holding off on merging until further discussion at next MISO Meeting

avarsava:
Discussed 01/08/2019, ahead of MISO Meeting: Go ahead with development of this issue, add marking aliquots as skipped after

avarsava:
> Could you please rebase on develop? My recent PR removed the Pool.getDuplicateIndicesSequences method.

Done!

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultPoolService.java
```diff
@@ -298,7 +298,8 @@ public void validateIndices(Pool pool, Pool beforeChange, Collection<ValidationE
     Set<String> bcIndices = getAllBadIndices(beforeChange);
 
     if(indices.size() > bcIndices.size()){ // If this change introduces new conflicts
-      String errorMessage = "Pools may not contain Library Aliquots with indices with 2 or fewer positions of difference, please address the following conflicts: ";
+      String errorMessage = "Pools may not contain Library Aliquots with indices with 2 or fewer positions of " +
```
callunity: Since the near-duplicate indices level is now configurable, can `2` be changed to `indexChecker.getWarningMismatches()`?

avarsava: Done, and updated release notes

PR #2023, opened 2019-07-31
GLT-2903: allow moving items between worksets

djcooke:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultWorksetService.java
```diff
@@ -197,4 +200,33 @@ public void authorizeDeletion(Workset workset) throws IOException {
...
+  }
+
+  @Override
+  public void moveLibraryAliquots(Workset from, Workset to, Collection<LibraryAliquot> items) {
+    moveItems(from, to, items, Workset::getLibraryAliquots, "library aliquots");
+  }
+
+  public <T extends Identifiable> void moveItems(Workset from, Workset to, Collection<T> items, Function<Workset, Set<T>> getter,
+      String pluralTypeLabel) {
+    Set<T> sourceItems = getter.apply(from);
```
apmasell: i think this will be okay, but I would test what happens if you move stuff from a workset to itself. It might be worth throwing a validation exception if `to.getId() == from.getId()` or just doing an early return.

PR #2028, opened 2019-08-08
Display index checking on pool orders

apmasell:
JIRA: GLT-2931

### DIFF
--- a/miso-web/src/main/webapp/scripts/warning_poolorder.js
+++ b/miso-web/src/main/webapp/scripts/warning_poolorder.js
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2012. The Genome Analysis Centre, Norwich, UK
+ * MISO project contacts: Robert Davey @ TGAC
+ * *********************************************************************
+ *
+ * This file is part of MISO.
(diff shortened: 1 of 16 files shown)


miso-web/src/main/webapp/scripts/warning_poolorder.js
```diff
@@ -0,0 +1,34 @@
...
+ * *********************************************************************
+ */
+
+WarningTarget.poolorder = {
+  getWarnings: function(poolorder) {
+    return [{
+      include: poolorder.duplicateIndices,
+      tableMessage: Constants.warningMessages.duplicateIndices
+    }, {
+      include: poolorder.nearDuplicateIndices,
```
callunity: Can a clause get added here: `&& (Constants.warningEditDistance > Constants.errorEditDistance)` ?
(When we get to the glorious only-automatic-analysis future, <=2 mismatches will be error and there will be no separate warning threshold; in that case, it should only display the duplicates warning)

apmasell: Makes sense. Will do.

PR #2037, opened 2019-08-16
Make widget to order pools for sample sheet creation

apmasell:
JIRA: GLT-2902


apmasell:
Realistically, these are only useful for non-MiSeq sequencers, which no one is going to need a sample sheet for. :man_shrugging:

PR #2048, opened 2019-08-30
Allow selecting all things from all pages in a table

apmasell:
JIRA: GLT-2970

### DIFF
--- a/miso-web/src/main/webapp/scripts/list.js
+++ b/miso-web/src/main/webapp/scripts/list.js
@@ -234,7 +234,7 @@ ListUtils = (function($) {
         $('<p>').attr('id', 'headerMessage').addClass('big big-' + (level || 'info')).css('margin-top', '.25em').text(text));
   };
 
-  var initTable = function(elementId, target, projectId, config, optionModifier) {
+  var initTable = function(elementId, target, projectId, config, optionModifier, selectAll) {
     var staticActions = target.createStaticActions(config, projectId);
     var bulkActions = target.createBulkActions(config, projectId);
     var columns = target.createColumns(config, projectId).filter(function(x) {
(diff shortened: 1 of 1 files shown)


miso-web/src/main/webapp/scripts/list.js
```diff
@@ -298,6 +298,29 @@ ListUtils = (function($) {
           updateSelectedLabel(state);
         }
       });
+      staticActions.push({
+        "name": "🗹",
```
avarsava: I don't know what this is supposed to be, but on my browser it shows up as one of those unknown character blocks?
image

miso-web/src/main/webapp/scripts/list.js
```diff
@@ -298,6 +298,29 @@ ListUtils = (function($) {
...
+        "title": "Select all on all pages",
+        "handler": function() {
+          if (jqTable.fnSettings().fnRecordsDisplay() > 10000) {
+            Utils.showOkDialog('Select all', [ 'Too many items selected.', 'Please use stricter filtering.' ]);
```
avarsava: Could we communicate the 10 000 item limit to users in the error message?

miso-web/src/main/webapp/scripts/list.js
apmasell: I � Unicode

miso-web/src/main/webapp/scripts/list.js
apmasell: Done.

apmasell:
I talked to Ilinca and she doesn't really want to see all those things in the table, so this is what she's after.

apmasell:
I mean, I can... :laughing:

PR #2049, opened 2019-09-04
GLT-2979 Pool Order changelogs

avarsava:
Since Create/Link Sequencing Orders buttons appear to be broken, change logs can be added for them when they get fixed, probably by me (GLT-2983, GLT-2984)

### DIFF
--- a/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultPoolOrderService.java
+++ b/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultPoolOrderService.java
@@ -2,6 +2,7 @@ package uk.ac.bbsrc.tgac.miso.service.impl;
 
 import java.io.IOException;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Consumer;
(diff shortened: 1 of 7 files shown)


sqlstore/src/main/resources/db/migration/V8001__pool_order_create_trigger.sql
```diff
@@ -0,0 +1,44 @@
...
+
+DROP TRIGGER IF EXISTS BeforeInsertPoolOrder//
+
+DROP TRIGGER IF EXISTS PoolOrderChange//
+CREATE TRIGGER PoolOrderChange BEFORE UPDATE ON PoolOrder
+FOR EACH ROW
+  BEGIN
+  DECLARE log_message varchar(500) CHARACTER SET utf8;
+  SET log_message = CONCAT_WS(', ',
+        CASE WHEN (NEW.alias IS NULL) <> (OLD.alias IS NULL) OR NEW.alias <> OLD.alias THEN CONCAT('alias: ', COALESCE(OLD.alias, 'n/a'), ' → ', COALESCE(NEW.alias, 'n/a')) END,
```
callunity: Alias isn't nullable, so you could eliminate those comparisons.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultPoolOrderService.java
```diff
@@ -202,29 +205,96 @@ private static boolean allMatch(Collection<OrderLibraryAliquot> one, Collection<
...
     to.setDraft(from.isDraft());
 
     // remove from TO if library isn't in FROM
     // add to TO if library isn't in TO
     // update proportion in TO if different in FROM
-    to.getOrderLibraryAliquots().removeIf(toOrderAli -> from.getOrderLibraryAliquots().stream()
-        .noneMatch(fromOrderAli -> toOrderAli.getAliquot().getId() == fromOrderAli.getAliquot().getId()));
+    Set<Long> fromIds = new HashSet<>();
+    for (OrderLibraryAliquot fromOrderAli : from.getOrderLibraryAliquots()){
+      fromIds.add(fromOrderAli.getAliquot().getId());
```
apmasell: I wondering if, for all the add/remove/change operations, there should be 1 aggregate message. That's how we do it for boxes... @callunity?

callunity: I was debating that, and looked at boxes as well -- it looks to my eye like they do individual messages though? (I'm looking at box 1 on staging). Although aggregating would lead to fewer messages, which is a positive, I find it more understandable to scan vertically rather than horizontally...

callunity: (And ultimately, Pool Orders will have fewer changelogs than boxes)

PR #2050, opened 2019-09-04
Don't permit Create/Link Seq Orders when Fulfilled

avarsava:


miso-web/src/main/webapp/scripts/form_poolorder.js
```diff
@@ -216,7 +216,7 @@ FormTarget.poolorder = (function($) {
           controls.push(makeButton('Unlink Sequencing Order', unlinkSequencingOrder, form));
         } else {
           controls.push(makeButton('Unlink Pool', unlinkPool, form));
-          controls.push(makeButton('Create Sequencing Order', createSequencingOrder, form), makeButton('Link Sequencing Order',
+          if (object.status !== "Fulfilled") controls.push(makeButton('Create Sequencing Order', createSequencingOrder, form), makeButton('Link Sequencing Order',
```
callunity: Can you please use the PoolOrder.Status enum for this?

PR #2051, opened 2019-09-05
Generate sample sheets like the Illumina Experiment Manager does

apmasell:
JIRA: GLT-2902

### DIFF
--- a/miso-web/src/main/webapp/scripts/hot_pool.js
+++ b/miso-web/src/main/webapp/scripts/hot_pool.js
@@ -293,6 +293,12 @@ HotTarget.pool = (function() {
                ]);
                return;
              }
+             if (platformTypes[0] != "ILLUMINA") {
+               Utils.showOkDialog("Error", [
+                 "Can only create sample sheets for Illumina sequencers."
+               ]);
+               return;
+             }
              var instrumentModels = Constants.instrumentModels.filter(function(model) {
                return (
                  model.instrumentType == "SEQUENCER" &&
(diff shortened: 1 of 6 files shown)


miso-web/src/main/webapp/scripts/hot_pool.js
```diff
@@ -307,45 +313,77 @@ HotTarget.pool = (function() {
...
+                     required: true,
+                     type: "select",
+                     getLabel: function(type) {
+                       return type.description;
+                     },
+                     values: Constants.illuminaExperimentTypes
+                   },
+                   {
+                     property: "sequencingParameters",
+                     label: "Instrument Model",
```
callunity: This should say `Sequencing Parameters`

miso-web/src/main/webapp/scripts/hot_pool.js
```diff
@@ -307,45 +313,77 @@ HotTarget.pool = (function() {
...
+                     label: 'Genome Folder',
+                     value: Constants.genomeFolder,
+                     required: true
+                   }, {
+                     property: "pools",
+                     label: "Lanes Configuration",
+                     required: true,
+                     type: "order",
+                     getLabel: Utils.array.getAlias,
+                     values: pools
```
callunity: Just realized that this doesn't allow you to load two lanes of the same pool... it's only one lane per pool. That's probably a separate ticket though.

apmasell: Yeah. It also can't deal with knowing about an S4 flowcell with joined/split lanes. For right now, they are only planning on using this on the MiSeq, so I'm okay with it being limited. I need to see sample sheets for the other platforms with those cases.

miso-web/src/main/webapp/scripts/hot_pool.js
apmasell: Fixed.

PR #2053, opened 2019-09-06
Allow specifying two different read lengths in Illumina runs

apmasell:
JIRA: GLT-2902

### DIFF
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/service/RunScannerClient.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/service/RunScannerClient.java
@@ -110,16 +110,48 @@ public class RunScannerClient {
           Predicate<SequencingParameters> isMatchingSequencingParameters;
           switch (dto.getPlatformType()) {
           case ILLUMINA:
-            isMatchingSequencingParameters = params -> params.getInstrumentModel().getPlatformType() == PlatformType.ILLUMINA &&
-                Math.abs(params.getReadLength() - ((IlluminaNotificationDto) dto).getReadLength()) < 2
-                && params.isPaired() == ((IlluminaNotificationDto) dto).isPairedEndRun()
-                && params.getChemistry() == Dtos.getMisoIlluminaChemistryFromRunscanner(((IlluminaNotificationDto) dto).getChemistry());
+            isMatchingSequencingParameters = params -> {
+              IlluminaNotificationDto illuminaDto = (IlluminaNotificationDto) dto;
+              if (params.getInstrumentModel().getPlatformType() != PlatformType.ILLUMINA) {
+                return false;
+              }
+              if (params.getChemistry() != Dtos.getMisoIlluminaChemistryFromRunscanner(illuminaDto.getChemistry())) {
+                return false;
+              }
+              // If we are talking to an old Run Scanner that doens't provide read lengths, use the old logic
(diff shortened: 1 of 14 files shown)


pinery-miso/src/main/resources/ca/on/oicr/pinery/lims/miso/queryAllOrderSamples.sql
```diff
@@ -2,7 +2,7 @@ SELECT o.sequencingOrderId orderId
         ,lib.NAME libraryId 
         ,bc1.sequence barcode 
         ,bc2.sequence barcode_two 
-        ,sp.paired paired 
+        ,sp.readLength = sp.readLength2 paired
```
callunity: Doesn't this mean that 10X runs will be described as single-end? I don't think that's accurate.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/service/RunScannerClient.java
```diff
@@ -110,16 +110,48 @@ private void processResults(List<NotificationDto> results) {
...
+              }
+              if (Math.abs(params.getReadLength() - illuminaDto.getReadLengths().get(0)) > 1) {
+                return false;
+              }
+              // If no second read is required, make sure none is provided
+              if (params.getReadLength2() == 0) {
+                return illuminaDto.getReadLengths().size() == 1;
+              }
+              // Otherwise, check the second read matches the right length
+              return Math.abs(params.getReadLength2() - illuminaDto.getReadLengths().get(1)) < 2;
```
callunity: ``
2019-09-09 10:03:56 ERROR: Failed to save run: 140103_D00353_0031_BC3E4NACXX
java.lang.IndexOutOfBoundsException: Index: 1, Size: 1
  at java.util.ArrayList.rangeCheck(ArrayList.java:657)
  at java.util.ArrayList.get(ArrayList.java:433)
  at uk.ac.bbsrc.tgac.miso.webapp.service.RunScannerClient.lambda$processResults$0(RunScannerClient.java:148)
```

apmasell: Fixed. I had a condition backwards.

pinery-miso/src/main/resources/ca/on/oicr/pinery/lims/miso/queryAllOrderSamples.sql
apmasell: We define them as single ended right now, so it's consistent with the legacy. I don't know what actually looks at information. CellRanger and BCL2FASTQ look at the bases mask, so I don't know if this matters. I could make it `readLength2 != 0`, which I guess is more accurate, but might cause all the 10X runs to go stale.

callunity: Half of our runs have `paired = 1` so I think we're already losing on consistency...
`paired` is used to make the `readLength`, which is independent of the `runBasesMask`. It doesn't look like `pinery_ius` even includes the `read_length`, so I don't think it'll be an analysis issue.

apmasell: Okay, fixed. Even if `pinery_ius` doesn't use it, it still becomes part of the hash in the LIMS key, so we'll just have to see.

PR #2058, opened 2019-09-16
Allow setting the AD domain DN

apmasell:


miso-web/src/main/resources/security.properties
```diff
@@ -38,6 +38,12 @@ security.ldap.stripRolePrefix:MISO_
 
 security.ad.emailDomain:your_email_domain
 security.ad.url:ldap://your.active_directory.server:389
+# Normally, the email domain can be converted to the base DN for a user. So, if
+# your email domain is example.com, then the inferred domain DN will be
+# dc=example,dc=com. If your AD is set up strangely where this is not the case,
+# override the base DN in which the query for
+# userPrinicpalName={username}@{emailDomain} will occur.
+security.ad.domainDn:
```
callunity: Can you give an example of what a correct value would look like?

apmasell: Not really. You have to know it from your LDAP server.

djcooke: Does this cause an error if the property is missing from `security.properties`? In release notes, please add a change if no, or an upgrade note if yes

apmasell: Added a note. It should be fine if not set.

PR #2062, opened 2019-09-19
GLT-2806: added new options for searching by dates

djcooke:


RELEASE_NOTES.md
```diff
@@ -1,13 +1,25 @@
...
+    `security.properties`. This should only be necessary if you have an unusual
+    ActiveDirectory configuration.
+  * New options for searching by date: year, month, fiscal year, fiscal quarter,
+    before/after specified range
+
+Upgrade notes:
+
+  * To enable searching by fiscal year and quarter, you must add the property
+    `miso.fiscalYearStartMonth` to your `miso.properties` file, and set the value to
+    the month number (1-12) of the start of your fiscal year.
```
callunity: Do you think it's worth noting in the release notes that the `miso.properties` file no longer lives in an `external` directory, on the off chance that there are people who copy that file and `sed` it or something as part of their release process?

djcooke: Not really? The docs have been updated, and it should be easy to find if you had to look for it knowing the previous location

djcooke:
@callunity: fix pushed. "FY" portion was meant to be optional when a quarter is specified, and now is. Also fixed it to use current fiscal year instead of current calendar year.

PR #2064, opened 2019-09-23
GLT-2807: search by user group

djcooke:


callunity:
There are some failing unit tests.

PR #2066, opened 2019-09-23
Move special information out of sample class alias

apmasell:
JIRA: GLT-2823

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/LimsUtils.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/LimsUtils.java
@@ -263,7 +263,7 @@ public class LimsUtils {
   }
 
   public static boolean isSampleSlide(DetailedSample sample) {
-    return sample instanceof SampleSlide || sample.getSampleClass().getAlias().equals(SampleSlide.SAMPLE_CLASS_NAME);
+    return sample instanceof SampleSlide || SampleSlide.SUBCATEGORY_NAME.equals(sample.getSampleClass().getSampleSubcategory());
   }
 
   public static boolean isDetailedLibrary(Library library) {
--- a/miso-web/src/main/webapp/scripts/form_sample.js
+++ b/miso-web/src/main/webapp/scripts/form_sample.js
@@ -359,7 +359,7 @@ FormTarget.sample = (function($) {
             }]
           }, {
             title: 'Slide',
-            include: config.detailedSample && object.sampleClassAlias === 'Slide',
+            include: config.detailedSample && object.sampleSubcategory === 'Slide',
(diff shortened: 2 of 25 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/LimsUtils.java
```diff
@@ -263,7 +263,7 @@ public static boolean isAliquotSingleCellSample(Sample sample) {
   }
 
   public static boolean isSampleSlide(DetailedSample sample) {
-    return sample instanceof SampleSlide || sample.getSampleClass().getAlias().equals(SampleSlide.SAMPLE_CLASS_NAME);
+    return sample instanceof SampleSlide || sample.getSampleClass().getAlias().equals(SampleSlide.SUBCATEGORY_NAME);
```
djcooke: `sample.getSampleClass().getAlias()` -> `sample.getSampleClass().getSampleSubcategory()`

sqlstore/src/main/resources/db/migration/V8000__sample_class_alias.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE SampleClass ADD COLUMN sampleSubcategory varchar(50) NOT NULL DEFAULT '';
```
djcooke: Can we use null instead of empty string?

apmasell: We can. I didn't because most of the logic is set up as `getSampleCategory().equals(whatever)` and I didn't want to make all of those null-safe.

djcooke: We use null everywhere else though, and since we often have objects not fully populated (often only the ID set), making it all null safe will likely prevent NPEs at some point

apmasell: Okay. Fixed.

miso-web/src/main/webapp/scripts/form_sample.js
```diff
@@ -359,7 +359,7 @@ FormTarget.sample = (function($) {
             }]
           }, {
             title: 'Slide',
-            include: config.detailedSample && object.sampleClassAlias === 'Slide',
+            include: config.detailedSample && object.sampleSubcategory === 'Slide',
```
djcooke: Need to add the `sampleSubcategory` field to `DetailedSampleDto`. Currently, all of the subcategory fields are missing from the Edit Sample page

apmasell: Okay. I think this is fixed.

PR #2069, opened 2019-09-24
Edit Sequencing Parameters

djcooke:


djcooke:
@callunity Yes, defaulting both reads to 0 makes sense. I'll also add validation that ensures they are both 0 for ONT, and that read 1 is not 0 for Illumina. I'll add some field information in the quick help dropdown too. I'll have to look into why the options aren't appearing in the run page dropdown immediately.

And yes, if we want to change the sequencing parameters model, that can definitely happen in a different ticket please 😆

djcooke:
@callunity I've made those changes. The new options do not show up _immediately_ after saving because refreshing constants now happens asynchronously. They show up shortly after saving

avarsava:
the ONT run type can be set to anything, will bad values have any kind of side effects?

djcooke:
@avarsava bad values would be values that don't exist in the ONT software and so aren't reported by Run Scanner. This means the parameters wouldn't be detected correctly when saving to MISO. I don't think the values should be constrained in MISO, as that would mean the code needs updated if ONT options change, rather than just updating the parameters in MISO to match. That would also kind of defeat the purpose of sequencing parameters for ONT.

avarsava:
Great, thanks. Are the CI test failures related to this change?

djcooke:
@avarsava I don't think so, but I'm looking into them now

djcooke:
The integration test failures were caused by this change and should now be fixed

PR #2070, opened 2019-09-25
[GLT-2996] Make library aliquot alias link to library aliquot

avarsava:


avarsava:
@callunity fixed! :)

callunity:
:heart_eyes_cat:

PR #2071, opened 2019-09-26
GLT-2812: added interface for editing container models

djcooke:


avarsava:
I was able to create a sequencing container with 0 partitions, does that make sense?

Also, looks like there's some IT failures

djcooke:
@avarsava working on fixing the tests currently. Will add validation to ensure partition count is positive

djcooke:
But zero is not positive.. Though I see I've used it myself when 0 is acceptable. 😿 

Updated validator with parameters to specify whether required and optional min/max values

PR #2072, opened 2019-09-26
[GLT-2961] Prevent changes which desync pools & pool orders

avarsava:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultPoolService.java
```diff
@@ -311,6 +315,21 @@ public void validateIndices(Pool pool, Pool beforeChange, Collection<ValidationE
     }
   }
 
+  @Override
+  public boolean checkMismatchedWithOrders(Pool pool, List<PoolOrder> poolOrders) throws IOException {
+    Set<Pair<LibraryAliquot, Integer>> poolAliquots = new HashSet<>(), poolOrderAliquots = new HashSet<>();
```
callunity: I'm not fond of this style of multiple variable declaration on a single line as I find it requires more cognitive overhead to parse, especially since the variables have such similar names. Do you have strong objections to doing the more verbose one-variable-per-line declaration?

avarsava: Changed to two separate declarations

callunity:
If you've edited a pool and caused the "mismatch with pool order" validation error, there's no way to easily compare between the pool and the pool order to see where the mismatch is. (There's no link to the Pool Order from the Pool page, as far as I can tell.) 
Would it be very difficult to give the user a more informative message about where the mismatch is happening? Maybe something that catches all the aliquots that are mismatched and indicates "Pool order requires $aliquotAlias ($aliquotName) to have proportion $proportion" or "Pool order requires $aliquotAlias to be absent" (and maybe make the pool order a link?)

djcooke:
@callunity @avarsava

I agree that the validation errors should state the specific problem aliquots. The pool order alias can be included as well.

I'm not sure about adding links in validation messages. It might be better to keep them simple rather than adding some kind of markup. Would adding a Pool Orders table to the Edit Pool page suffice?

callunity:
I think that a combination of pool order alias in the validation message + Pool Orders table on the Edit Pool page would work well.

PR #2075, opened 2019-09-30
Change order thigns are present in generated samples sheet

apmasell:
JIRA: GLT-2902


djcooke:
Please add release notes

PR #2076, opened 2019-10-02
GLT-2812: added interface for editing instrument models

djcooke:


miso-web/src/main/webapp/WEB-INF/pages/editInstrumentModel.jsp
```diff
@@ -0,0 +1,83 @@
+<%--
+  ~ Copyright (c) 2012. The Genome Analysis Centre, Norwich, UK
+
+  ~ MISO project contacts: Robert Davey @ TGAC
```
avarsava: Do we include this for historical reasons or is this inaccurate?

avarsava:
I've checked out this branch to test out the page but when I `mvn install`, 

HibernateSequencingContainerModelDaoTest.testGetUsagePlatform:151 expected:<4> but was:<0>

miso-web/src/main/webapp/WEB-INF/pages/editInstrumentModel.jsp
djcooke: Was there because copy+paste. Removed

avarsava:
I cannot set 'Sequencing Containers Per Run' to anything other than 0 - on both new and existing models, it says 'Should be 1 or greater for sequencer models'. The instrument type is set to Sequencer. Not sure I understand the error.
Adding a Run Position with an alias of >10 characters causes 'Something has gone terribly wrong, please report this to your MISO administrator'. Also can't remove the offending alias without refreshing the page and losing all my work - 'Nothing selected'.

avarsava:
Does being able to set the 'Index Sequencing' value make sense for Array Scanner and Other instrument types?

djcooke:
@avarsava: all fixed

avarsava:
I tried editing a Sequencer to use 128 Sequencing Containers per Run and this causes 'Something has gone terribly wrong'. 127 is fine?

djcooke:
@avarsava fixed this case, though similar things will probably happen when overflowing any datatype since we've never had validation to prevent it.

Also found and fixed an SQLException that was happening when lowering the number of containers if there were no instruments of the model being edited.

PR #2083, opened 2019-10-09
"Missing Index" Warning Fixes

djcooke:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultPoolService.java
```diff
@@ -300,11 +311,7 @@ public void validateIndices(Pool pool, Pool beforeChange, Collection<ValidationE
               "fewer positions of difference, please address the following conflicts: ",
               indexChecker.getWarningMismatches());
       indices.removeAll(bcIndices);
-
-      for (String index : indices){
-        errorMessage += index + " ";
-      }
-
+      errorMessage += indices.stream().map(index -> index.length() == 0 ? "(no indices)" : index).collect(Collectors.joining(", "));
```
avarsava: ""Pools may not contain Library Aliquots with indices with 2 or fewer positions of difference, please address the following conflicts: (no indices)"" is a possible error message? I'm not sure I understand how that is possible

djcooke: If multiple libraries are missing indices, they are considered duplicates

djcooke: Before this change, the message is "Pools may not contain Library Aliquots with indices with 2 or fewer positions of difference, please address the following conflicts: " (empty string is included where the sequence should be)

PR #2085, opened 2019-10-10
Rebuild constants.js every 15 min

apmasell:


apmasell:
Yeah, I labelled it as bad, but I like the idea known issues. Not all of them are worth of rollback. Let's do that in the future.

PR #2086, opened 2019-10-10
Use tissue piece type in tissue piece alias

apmasell:
JIRA: GLT-2769


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/naming/generation/OicrSampleAliasGenerator.java
```diff
@@ -105,7 +107,14 @@ private String addSiblingTag(String parentAlias, DetailedSample sample) throws I
     if (sc == null || sc.getSuffix() == null) {
       throw new InvalidParameterException("Unexpected null SampleClass or suffix");
     }
```
djcooke: The check for null SampleClass should stay here, but the check for null suffix should be moved into the else block below since sample classes of the Tissue Piece subcategory don't need a suffix as it wouldn't be used

apmasell: Right. Fixed.

PR #2091, opened 2019-10-15
Delete all the things!

djcooke:
:fire: :fire: :fire:

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/DeleterService.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/DeleterService.java
@@ -5,9 +5,16 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
+import org.hibernate.exception.ConstraintViolationException;
+import org.slf4j.LoggerFactory;
+import org.springframework.dao.DataIntegrityViolationException;
+
+import com.eaglegenomics.simlims.core.User;
+
 import uk.ac.bbsrc.tgac.miso.core.data.Deletable;
 import uk.ac.bbsrc.tgac.miso.core.security.AuthorizationException;
 import uk.ac.bbsrc.tgac.miso.core.security.AuthorizationManager;
+import uk.ac.bbsrc.tgac.miso.core.service.exception.ValidationException;
 import uk.ac.bbsrc.tgac.miso.core.service.exception.ValidationResult;
 import uk.ac.bbsrc.tgac.miso.core.store.DeletionStore;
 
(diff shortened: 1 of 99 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/DeleterService.java
```diff
@@ -82,7 +86,18 @@ public default void bulkDelete(Collection<T> objects) throws IOException {
     result.throwIfInvalid();
     for (T object : managedObjects) {
       beforeDelete(object);
-      getDeletionStore().delete(object, getAuthorizationManager().getCurrentUser());
+      try {
```
apmasell: I think this nested try can be flattened.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultKitDescriptorService.java
```diff
@@ -161,4 +165,38 @@ public long count(Consumer<String> errorHandler, PaginationFilter... filter) thr
...
+
+  @Override
+  public AuthorizationManager getAuthorizationManager() {
+    return authorizationManager;
+  }
+
+  @Override
+  public ValidationResult validateDeletion(KitDescriptor object) throws IOException {
+    ValidationResult result = new ValidationResult();
+    switch (object.getKitType()) {
```
apmasell: Does it make sense to do all the checks in case there's some database shenanigans? Worst case, they'll be zero.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultProjectService.java
```diff
@@ -162,4 +172,42 @@ public void setAuthorizationManager(AuthorizationManager authorizationManager) {
...
+  }
+
+  @Override
+  public void authorizeDeletion(Project object) throws IOException {
+    authorizationManager.throwIfNonAdminOrMatchingOwner(object.getCreator());
+  }
+
+  @Override
+  public uk.ac.bbsrc.tgac.miso.core.service.exception.ValidationResult validateDeletion(Project object) throws IOException {
+    uk.ac.bbsrc.tgac.miso.core.service.exception.ValidationResult result = new uk.ac.bbsrc.tgac.miso.core.service.exception.ValidationResult();
```
apmasell: I think you need to also check for studies or have them cascade delete. Given how tied to a project they are, cascade makes the most sense to me.

Also, does there need to be explicit cascading for `LibraryTemplate_Project` and `Project_Attachment`?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/DeleterService.java
djcooke: Flattened. Also added better logging for when this happens except for `User` objects, where this is expected since they're connected to way too many things to bother checking

PR #2092, opened 2019-10-17
Edit indices

djcooke:


miso-web/src/main/webapp/scripts/form_indexfamily.js
```diff
@@ -0,0 +1,119 @@
+if (typeof FormTarget === 'undefined') {
+  FormTarget = {};
```
callunity: Could replace this block with `FormTarget = FormTarget || {};`

miso-web/src/main/webapp/scripts/hot_index.js
```diff
@@ -0,0 +1,66 @@
...
+      pack: function(index, flat, errorHandler) {
+        if (!index.family) {
+          index.family = config.indexFamily;
+        }
+      }
+    }, HotUtils.makeColumnForInt("Position", true, 'position', HotUtils.validator.integer(true, 1, 2)),
+        HotUtils.makeColumnForText('Name', true, 'name', {
+          validator: HotUtils.validator.requiredTextNoSpecialChars
+        }), HotUtils.makeColumnForText('Sequence', !config.indexFamily.fakeSequence, 'sequence', {
+          validator: HotUtils.validator.requiredTextNoSpecialChars
```
callunity: When adding unique dual indices to a library, the code currently expects that the correct Index 2 will have the same name as the Index 1. Is there a (sensible) way we could add a constraint to the create/update indices HoT to have it check that each Index 2's name corresponds to an Index 1's name? If not, then a text box indicating this expectation would probably do.

miso-web/src/main/webapp/scripts/hot_index.js
```diff
@@ -0,0 +1,66 @@
...
+        }), {
+          header: 'Sequences',
+          data: 'realSequences',
+          description: 'Values should be comma-separated. e.g. "ACACAC, GTGTGT"',
+          include: config.indexFamily.fakeSequence,
+          validator: HotUtils.validator.requiredTextNoSpecialChars,
```
callunity: Can this prohibit commas if it's not a fake sequence family?

miso-web/src/main/webapp/scripts/hot_index.js
```diff
@@ -0,0 +1,66 @@
...
+      }
+    }, HotUtils.makeColumnForInt("Position", true, 'position', HotUtils.validator.integer(true, 1, 2)),
+        HotUtils.makeColumnForText('Name', true, 'name', {
+          validator: HotUtils.validator.requiredTextNoSpecialChars
+        }), HotUtils.makeColumnForText('Sequence', !config.indexFamily.fakeSequence, 'sequence', {
+          validator: HotUtils.validator.requiredTextNoSpecialChars
+        }), {
+          header: 'Sequences',
+          data: 'realSequences',
+          description: 'Values should be comma-separated. e.g. "ACACAC, GTGTGT"',
```
callunity: Does this description get displayed anywhere?

miso-web/src/main/webapp/scripts/hot_index.js
djcooke: I think it's fair to expect administrators to enter values that make sense, but we could... If we're prohibiting commas, then we may as well prohibit a lot of other things though. Is there any reason to allow characters other than `[ACGT]`?

miso-web/src/main/webapp/scripts/hot_index.js
djcooke: Column descriptions are added to the quick help section at the top of the page

miso-web/src/main/webapp/scripts/hot_index.js
djcooke: This would mean that index 1 always needs to be created or updated before index 2, which has more potential to be annoying than helpful. I'll add a column description and also detail that better in the user manual

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultIndexService.java
```diff
@@ -10,51 +9,111 @@
...
+  protected void collectValidationErrors(Index object, Index beforeChange, List<ValidationError> errors) throws IOException {
+    if (ValidationUtils.isSetAndChanged(Index::getName, object, beforeChange)
+        && indexStore.getByFamilyPositionAndName(object.getFamily(), object.getPosition(), object.getName()) != null) {
+      errors.add(new ValidationError("name", "This index family already contains an index with this name and position"));
+    }
+    if (ValidationUtils.isSetAndChanged(Index::getPosition, object, beforeChange)
+        && indexStore.getByFamilyPositionAndName(object.getFamily(), object.getPosition(), object.getName()) != null) {
+      errors.add(new ValidationError("position", "This index family already contains an index with this name and position"));
+    }
+    if (beforeChange != null && object.getPosition() != beforeChange.getPosition()) {
```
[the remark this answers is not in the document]
djcooke: We add duplicate sequences intentionally so that we can have the same index entered multiple times with different box/plate positions in the name :crying_cat_face:

miso-web/src/main/webapp/scripts/hot_index.js
callunity: I'm not seeing them at the top of the Create Indices page, can you look into why that might be?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultIndexFamilyService.java
```diff
@@ -0,0 +1,88 @@
...
+    if (ValidationUtils.isSetAndChanged(IndexFamily::getName, object, beforeChange)
+        && indexFamilyDao.getByName(object.getName()) != null) {
+      errors.add(new ValidationError("name", "There is already an index family with this name"));
+    }
+  }
+
+  @Override
+  protected void applyChanges(IndexFamily to, IndexFamily from) throws IOException {
+    to.setName(from.getName());
+    to.setFake(from.hasFakeSequence());
```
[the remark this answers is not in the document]
djcooke: No, you're right; it's not safe. I'll prevent it from happening

miso-web/src/main/webapp/scripts/hot_index.js
djcooke: @apmasell those are "fake sequence families" though, so they wouldn't be validated here

miso-web/src/main/webapp/scripts/hot_index.js
djcooke: Are you working with a "fake sequence"/multi-sequence family? It's showing up for me

miso-web/src/main/webapp/scripts/hot_index.js
callunity: Ah yes, 10X indices need to have both real sequences and fake sequences. The fake sequence is the name that the *Ranger software uses to demultiplex (SI-GA-A1 -> SI-GA-H12, not zero-padded) and the real sequences are what we use to avoid index collisions. Both will need to be able to be entered.

miso-web/src/main/webapp/scripts/form_indexfamily.js
djcooke: I feel like there was a reason I didn't use that syntax originally on all of the form scripts, but it doesn't seem to break anything, so :man_shrugging:

miso-web/src/main/webapp/scripts/hot_index.js
djcooke: Ah, I didn't know the fake sequence was actually used for anything. I'll make that column always appear and stop copying the name into the sequence. Any reason to prohibit commas specifically then?

callunity: Relatedly, right now for "multi-sequence indices" the value in the HoT "Name" column is being stored in the `Indices.sequence` db column which doesn't look quite right.

miso-web/src/main/webapp/scripts/hot_index.js
callunity: Ah, that was the issue. Better now :+1:

miso-web/src/main/webapp/scripts/hot_index.js
callunity: I think it would work for non-ACGT characters to be prohibited in the `Sequence`/`sequence` column above, and commas should be kept for this `Sequences`/`realSequences` column. I'm trying to think of good names to distinguish between the two columns though, if they'll both be displayed for the `fakeSequences` indices.

djcooke: * Changed the column name to "Fake Sequence" if a fake sequence family
* For non-fake-sequence families, validator is changed to only allow ACGT characters
* For fake-sequence families, "Sequences" field now only allows comma-separated sets of ACGT characters
* Added more info to field descriptions

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultIndexFamilyService.java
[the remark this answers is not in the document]
djcooke: Prevented

docs/user_manual/type_data.md
```diff
@@ -352,22 +352,30 @@ that are intended to be used together. Index sequences should all be unique with
...
+In other index families, any index 1 may be matched with any index 2.
 
 Some Illumina sequencers, including the NextSeq, sequence the index 2 as the
 reverse complement of other sequencers. MISO has an option to set the
 sequencing platform to use the normal index in the interface and provide the
 reverse complement in sample sheets and though the Pinery data export
 interface.
 
+In some index families such as 10X Genomics kits, multiple sequences are associated with a single index. These are
+referred to as multi-sequence indices in MISO.
```
callunity: Could the Fake/Demultiplexing sequence thing get a few sentences here?

miso-web/src/main/webapp/scripts/hot_index.js
```diff
@@ -0,0 +1,71 @@
...
+              index.family = config.indexFamily;
+            }
+          }
+        },
+        HotUtils.makeColumnForInt("Position", true, 'position', HotUtils.validator.integer(true, 1, 2)),
+        HotUtils.makeColumnForText('Name', true, 'name', {
+          description: 'For unique dual index families, the name of index 1 and index 2 must match for automatic selection to work',
+          validator: HotUtils.validator.requiredTextNoSpecialChars
+        }),
+        HotUtils.makeColumnForText(config.indexFamily.fakeSequence ? 'Fake Sequence' : 'Sequence', true, 'sequence', {
```
callunity: What do you think about changing `Fake Sequence` to `Demultiplexing Sequence`, or something similar/descriptive?

docs/user_manual/type_data.md
djcooke: Added

miso-web/src/main/webapp/scripts/hot_index.js
djcooke: Changed to "Demultiplexing Name" since it is a name rather than a sequence

PR #2093, opened 2019-10-22
GLT-2959: improved interface for editing library templates

djcooke:
Open to opinions on whether the bulk edit page, and the bulk add/remove project actions should be removed

### DIFF
--- a/miso-web/src/main/webapp/scripts/list_project.js
+++ b/miso-web/src/main/webapp/scripts/list_project.js
@@ -63,7 +63,7 @@ ListTarget.project = {
             }), null, function(response) {
               Utils.showWizardDialog("Add Project", response.map(function(project) {
                 return {
-                  name: project.alias,
+                  name: project.name + ': ' + project.alias + (project.shortName ? ' (' + project.shortName + ')' : ''),
                   handler: function() {
                     LibraryTemplate.addProject(project);
                   }
--- a/miso-web/src/main/webapp/scripts/list_project.js
+++ b/miso-web/src/main/webapp/scripts/list_project.js
@@ -28,23 +28,59 @@ ListTarget.project = {
   },
   getQueryUrl: null,
   createBulkActions: function(config, projectId) {
-    return [ListUtils.createBulkDeleteAction("Projects", "projects", function(project) {
-      if (!project.alias) {
-        return project.shortName;
-      } else if (!project.shortName) {
-        return project.alias;
-      } else {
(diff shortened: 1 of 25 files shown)


sqlstore/src/main/resources/db/migration/V8510__libraryTemplate_volumeUnits.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE LibraryTemplate ADD COLUMN volumeUnits varchar(30);
+UPDATE LibraryTemplate SET volumeUnits = 'MICROLITRES' WHERE defaultVolume IS NOT NULL;
```
callunity: Can you help me understand what this is doing? I'm confused about why it's changing existing volumeUnits to `MICROLITRES`, and leave NULL volumeUnits alone (and it looks like that's intended, from the lack of default on the volumeUnits column).

miso-web/src/main/webapp/scripts/list_project.js
```diff
@@ -28,23 +28,59 @@ ListTarget.project = {
...
+            label: "Search",
+            property: "query",
+            value: ""
+          }], function(results) {
+            Utils.ajaxWithDialog('Getting Projects', 'GET', '/miso/rest/projects/search?' + jQuery.param({
+              q: results.query
+            }), null, function(response) {
+              Utils.showWizardDialog("Add Project", response.map(function(project) {
+                return {
+                  name: project.alias,
```
callunity: Since the user might be searching by name, alias, or shortName, could this return text that includes all of these so they can confirm they've got the correct one? Something like `PRO13: Really Long Name For This Project (RLNFTP)` ?

sqlstore/src/main/resources/db/migration/V8510__libraryTemplate_volumeUnits.sql
djcooke: If default volume is set, then volume units should also be set. Microlitres is the only volume unit that currently exists in MISO, so I'm setting that on all the templates that already specify a volume

callunity: Makes sense, thanks!

miso-web/src/main/webapp/scripts/list_project.js
djcooke: Updated here and also on the library templates list

PR #2103, opened 2019-10-29
GLT-2817: allow editing sample classes

djcooke:
Will update user manual after the documentation makeover PR is merged


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SampleIdentity.java
```diff
@@ -11,6 +12,8 @@
   
   public static final String CATEGORY_NAME = "Identity";
 
+  public static final List<String> SUBCATEGORIES = Collections.unmodifiableList(Collections.emptyList());
```
apmasell: The empty list is implicitly unmodifable.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SampleClass.java
```diff
@@ -4,68 +4,65 @@
...
+  public void setArchived(boolean archived);
 
-  boolean isArchived();
+  public boolean isDirectCreationAllowed();
 
-  void setArchived(boolean archived);
+  public void setDirectCreationAllowed(boolean directCreationAllowed);
 
-  boolean isDirectCreationAllowed();
+  public boolean getDNAseTreatable();
```
avarsava: It looks like everything else is using 'is__' for boolean getters, is there a reason this method (and others named the same thing) still use 'get'?

miso-web/src/main/webapp/scripts/form_sampleclass.js
```diff
@@ -0,0 +1,145 @@
...
+          maxLength: 255
+        },
+        {
+          title: 'Category',
+          data: 'sampleCategory',
+          type: 'dropdown',
+          required: true,
+          getSource: function() {
+            return Constants.sampleCategories;
+          },
```
apmasell: Is this not going to be bad for the DB if you change it?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SampleClass.java
djcooke: It hadn't annoyed me enough to rename it? We don't consistently use one or the other throughout MISO

avarsava: Alright! Just noticed getArchived() became isArchived() and this one didn't :+1:

miso-web/src/main/webapp/scripts/form_sampleclass.js
djcooke: There's validation in the service layer that prevents changing it if the class is used by any samples, or if relationships would become invalid

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SampleClass.java
djcooke: `getArchived` had annoyed me enough 😆

PR #2108, opened 2019-10-31
Mkdocs styling and more documentation updates

morgantaschuk:
- styled readthedocs to look like MISO
- updated urls to point to the right locations
- Documentation badge on the readme
- Updated copyright

Screen Shot 2019-10-30 at 2 23 21 PM


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/util/form/TagUtils.java
```diff
@@ -54,7 +54,8 @@ private static Authentication getAuthentication() {
 
   public static String docsVersion() {
     if (Version.VERSION.matches("^\\d*\\.\\d*\\.\\d*$")) {
-      return Version.VERSION.substring(0, Version.VERSION.lastIndexOf(".")) + ".x";
+      //FIXME once we know how readthedocs handles version tags
+      return Version.VERSION;
```
djcooke: It seems like it'll build versions that match our tag names, like "v0.2.190" and create a "stable" version for our most recent release, and "latest" pointing to current develop branch. Probably safest to just return "latest" for now so we know it won't be a broken link in the next release though?

PR #2114, opened 2019-11-04
Enforce unique run aliases

callunity:
GLT-3052


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -309,7 +309,7 @@ private void validateChanges(Run before, Run changed) throws IOException {
     }
     if (isSetAndChanged(Run::getAlias, changed, before)) {
       Run maybeDupe = getRunByAlias(changed.getAlias());
-      if (changed.getId() != maybeDupe.getId()) {
+      if (maybeDupe != null && changed.getId() != maybeDupe.getId()) {
```
djcooke: You can just check

```
if (isSetAndChanged(Run::getAlias, changed, before) && getRunByAlias(changed.getAlias() != null)
```

If the alias is new or changed, then any run you get from the database for that alias can't possibly be the one being saved

callunity: Thanks, I've updated it to include this.

PR #2115, opened 2019-11-05
GLT-2900: cleaned up OicrLibraryAliasValidator code

djcooke:


djcooke:
@callunity: Changed alias-related exceptions to validation exception

PR #2122, opened 2019-11-08
GLT-3010: add search by freezer

djcooke:


callunity:
Can boxes be filtered by freezer as well?

djcooke:
@callunity Yes, I suppose that makes sense! Added, and put the advanced search help on the Boxes list page as well.

PR #2129, opened 2019-11-20
Protect against warning for null version

avarsava:


miso-web/src/main/webapp/scripts/list_kits.js
```diff
@@ -49,7 +49,9 @@ ListTarget.kit = {
       "sTitle": "Version",
       "include": true,
       "iSortPriority": 0,
-      "mData": "version"
+      "mData": function(data){
+        return data.version? data.version : null;
```
djcooke: Should probably change the condition to `data.hasOwnProperty('version')` in-case of version set to `0`

avarsava: Fixed!

PR #2139, opened 2019-12-05
GLT-3086: added run data approval fields

djcooke:


miso-web/src/main/webapp/scripts/form_run.js
```diff
@@ -205,6 +207,47 @@ FormTarget.run = (function($) {
...
+          getSource: function() {
+            return [{
+              label: 'True',
+              value: true
+            }, {
+              label: 'False',
+              value: false
+            }];
+          },
+          convertToBoolean: true,
```
callunity: What do you think about `Approved`/`Not Approved`?

djcooke: For the dropdown labels you mean? "Data Approved: Not Approved" seems kind-of weird

Maybe "Yes"/"No" would be better than "True"/"False"?

callunity: I like "Yes"/"No"!

djcooke:
@callunity: updated

PR #2144, opened 2019-12-19
Export intialVolume in Pinery samples

callunity:
GR-978


djcooke:
Please add to release notes

PR #2145, opened 2020-01-02
Table Redesign

djcooke:


avarsava:
Would it be possible to make the warning icons a bit bolder in line width and colour contrast, and force them to not wrap the line around? I barely noticed the warning icons at first, and the tooltip cuts off under the sidebar 
image

miso-web/src/main/webapp/scripts/warning.js
```diff
@@ -66,9 +66,23 @@ var Warning = (function($) {
...
+            level = 'error';
+          } else if (warnings.some(function(warning) {
+            return !warning.level || warning.level === 'important';
+          })) {
+            level = 'important';
+          }
+          html += ' <div class="tooltip"><span class="message-' + level + '">⚠</span><span class="tooltiptext">'
+              + warnings.map(function(warning) {
+                return warning.tableMessage;
+              }).join(';<br>') + '</span></div>';
```
avarsava: This looks as though it draws 1 '⚠' for all the warnings, should we consider having 1 '⚠' per warning instead?

djcooke: I don't think additional icons will be useful as they already use the colour of the most severe warning, and the point of this change is to save space. All full warning messages are included in the tooltip

djcooke:
@avarsava: bolded and increased the size of warning icons, so they're more visible now: 
image

djcooke:
@callunity: The only change on the Pools list is moving the warnings from full text in the description column into an icon in the name column, which is saving a bit of space.

The request for this change was only for the Samples, Libraries, and Library Aliquots lists. It would probably be a good idea to solicit feedback for the Pools list too. Maybe there are some columns that aren't useful, or that could be moved further to the right if less useful.

PR #2152, opened 2020-01-09
GLT-3062: allow changing sample project

djcooke:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
```diff
@@ -274,6 +277,8 @@ public ModelAndView setupForm(@PathVariable Long sampleId, ModelMap model) throw
         .map(Dtos::asDto)
         .collect(Collectors.toList()));
 
+    model.put("projects", mapper.writeValueAsString(projectService.list().stream().map(Dtos::asDto).collect(Collectors.toList())));
```
apmasell: Navel gazing question: should we limit to active projects?

djcooke:
Going to add a warning of "This project change will affect X libraries" per sample when changing project on aliquots

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
djcooke: Yeah, I'll change it to only display active projects + current project if not active

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/EditSampleController.java
```diff
@@ -511,7 +517,7 @@ private BulkPropagateSampleBackend(SampleClass targetSampleClass, BoxDto newBox,
     }
 
     @Override
-    protected SampleDto createDtoFromParent(Sample item) {
+    protected SampleDto createDtoFromParent(Sample item) { // TODO: identityConsentLevel
```
callunity: Is this TODO still needed? Looks like it's implemented below in 530-532.

djcooke: removed

djcooke:
Added library warning and tests should be fixed now

PR #2154, opened 2020-01-13
GLT-3102: allow changing sequencing container model

djcooke:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultContainerService.java
```diff
@@ -122,6 +129,32 @@ private void validateChange(SequencerPartitionContainer container, SequencerPart
...
+        Stream<InstrumentModel> requiredInstrumentModels = beforeChange.getRunPositions().stream()
+            .map(RunPosition::getRun)
+            .map(Run::getSequencer)
+            .map(Instrument::getInstrumentModel);
+        if (requiredInstrumentModels.anyMatch(required -> after.getInstrumentModels().stream()
+            .map(InstrumentModel::getId)
+            .noneMatch(id -> id == required.getId()))) {
+          errors.add(new ValidationError("model.id",
+              String.format("Can only change to a model compatible with the linked runs' instrument models (%s)",
+                  LimsUtils.joinWithConjunction(requiredInstrumentModels.map(InstrumentModel::getAlias).collect(Collectors.toSet()),
```
apmasell: Streams aren't supposed to be reused. Maybe create a list or set of instrument models?

djcooke: fixed

miso-web/src/main/webapp/scripts/form_container.js
```diff
@@ -40,6 +36,10 @@ FormTarget.container = (function() {
           getDisplayValue: function(container) {
             return container.model.alias;
           }
+        }, {
+          title: 'Change Model',
+          type: 'special',
```
callunity: (Just leaving a note here that we spoke earlier about converting the container model field to a dropdown instead.)

djcooke: Changed. Dropdown will sometimes contain a mix of models from multiple instruments when the container isn't connected to any runs, but it should never be too long of a list

PR #2159, opened 2020-01-17
New Pinery Attributes

djcooke:


pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/MisoClient.java
```diff
@@ -920,7 +924,46 @@ public String extractStringValueFrom(ResultSet rs) throws SQLException {
           return extractBigDecimalString(rs, getSqlKey());
         }
       }, //
-      REERENCE_SLIDE_ID("reference_slide_id", "Reference Slide");
+      REERENCE_SLIDE_ID("reference_slide_id", "Reference Slide"), TARGET_CELL_RECOVERY("target_cell_recovery", "Target Cell Recovery") {
```
callunity: Could add the terminal `//` after REFERENCE_SLIDE_ID to make it clearer the below function is for TARGET_CELL_RECOVERY (if you like)

djcooke: Did that and fixed the spelling on `REFERENCE_SLIDE_ID` too

pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/MisoClient.java
```diff
@@ -935,6 +935,34 @@ public String extractStringValueFrom(ResultSet rs) throws SQLException {
...
+      }, //
+      SPIKE_IN("spike_in", "Spike-In"), //
+      SPIKE_IN_DILUTION("spike_in_dilution_factor", "Spike-In Dilution Factor") {
+        @Override
+        public String extractStringValueFrom(ResultSet rs) throws SQLException {
+          String raw = rs.getString(getSqlKey());
+          if (raw == null) {
+            return null;
+          }
+          switch (raw) {
```
apmasell: Maybe these return values should just be integers? I would ask Peter and/or Larry how these get used.

djcooke: :+1: Changed to a decimal after discussing with Peter

djcooke: :+1: Changed to a decimal after discussing with Peter

PR #2169, opened 2020-01-24
Add null check for group description

apmasell:
JIRA: GC-7493


apmasell:
Updated.

djcooke:
`DetailedLibraryAliquot` too

apmasell:
Oops. Fixed.

PR #2171, opened 2020-01-24
Upgrade Pinery to 2.9.0

apmasell:


djcooke:
Oh, should also be noted under Upgrade Notes in release notes

PR #2177, opened 2020-01-30
GLT-3126: improved performance of Edit Sample page

djcooke:


avarsava:
Andre is in a meeting but says the change looks good and to merge it

PR #2185, opened 2020-02-06
GLT-3152: fixed timestamps saving in wrong time zone

djcooke:
This fixes the issues, though there are still ugly bits...

1. User enters time in UI time
2. Controller converts it to DB time
3. Saves to DB in DB time
4. Loads from DB as DB time
5. Interceptor converts to UI time
6. Displays to user as UI time

UI->DB time happens in Controller and DB->UI time happens in Interceptor (when loading via Hibernate). Ideally these would happen at the same layer. Unfortunately, there doesn't seem to be a good way to convert to DB time on insert/update in the interceptor - there's an `onSave` method for inserts, but not a method that is consistently invoked once per update. Moving all conversion to happen at the controller layer would be error-prone, as we'd have to handle every timestamp field of every entity type.

1. Service layer generates created/modified times in DB time
2. Saves to DB in DB time
3. Loads from DB as DB time
4. Interceptor converts to UI time
5. Displays to user as UI time
6. When re-saving, UI time values are still there, though they won't get saved
7. When regenerating, they will be generated in DB time

Step 6 in this flow is another ugly part and again is due to the conversions happening at separate layers.

This solution is basically returning it to how it was before the issue started, and also setting the JVM time zone to ensure that it works as intended when configured correctly


djcooke:
@avarsava: do you have time corrections configured in your miso.properties?

```
miso.timeCorrection.dbZone:UTC
miso.timeCorrection.uiZone:Canada/Eastern
```

If not, both will default to UTC, and unless you've changed it, your MySQL will actually be running in Eastern. `dbZone` needs to match the MySQL's time zone (Can check with `SELECT @@global.time_zone;`)

The best way to test the changes would be to add these settings to miso.properties if they're missing, and also change MySQL to use UTC, which you can do by adding this to `my.cnf` (probably in`/etc/mysqld/`) and then restarting mysql (probably via `systemctl restart mysql`)

```
[mysqld]
default-time-zone=UTC
```

avarsava:
Okay, adding both dbZone and uiZone to my miso.properties fixed it. The upgrade notes only suggest the existence of dbZone though, could that be expanded upon?

djcooke:
@avarsava: The properties are not new, but I added a note to clarify their renewed importance. Also applied the timestamp correction to sample receipt and fixed an issue where you could set items' volume or add them to boxes after distributing them.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/component/TimeZoneCorrector.java
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  public void toDbTime(Date date, Consumer<Date> setter) {
+    if (date == null) {
+      return;
+    }
+    long time = date.getTime();
+    int fromOffset = uiZone.getOffset(time);
+    int toOffset = dbZone.getOffset(time);
+    long shifted = time - fromOffset + toOffset;
```
apmasell: This logic might not work around leap days, but I'm not sure a better way without the new `java.time` stuff. So, :man_shrugging:

djcooke: It's only operating on the long value (ms since epoch), so leap days don't really have any effect

PR #2186, opened 2020-02-07
GLT-3151: fix deleting sequencing orders

djcooke:
Note: the actual fix to the error is in `DefaultPoolService#refreshPoolElements`


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultPoolService.java
```diff
@@ -299,16 +294,15 @@ private void refreshPoolElements(Pool pool) throws IOException {
...
     Set<String> indices = getAllBadIndices(pool);
     Set<String> bcIndices = getAllBadIndices(beforeChange);
 
-    if(indices.size() > bcIndices.size()){ // If this change introduces new conflicts
+    if (indices.size() > bcIndices.size()) { // If this change introduces new conflicts
       String errorMessage = String.format("Pools may not contain Library Aliquots with indices with %d or " +
-              "fewer positions of difference, please address the following conflicts: ",
-              indexChecker.getWarningMismatches());
+          "fewer positions of difference, please address the following conflicts: ",
+          indexChecker.getWarningMismatches());
```
avarsava: When adding Library Aliquots to a Pool, isn't this error supposed to pop up if there's an index conflict? It's just giving me the red ! triangle in the list then saying the form is 'invalid' when I try to save, but they can be added to the list. But it also appears to work the same way on Staging, so either this has been changed for a while or I've forgotten how this works ¯\_(ツ)_/¯

djcooke: This PR did not affect that, but fixed. There was no error container to display the message in

PR #2188, opened 2020-02-11
GLT-3144: allow configuring a secondary naming scheme

djcooke:


miso-web/src/main/webapp/scripts/form_project.js
```diff
@@ -94,6 +101,18 @@ FormTarget.project = (function($) {
...
+          data: 'secondaryNaming',
+          type: 'checkbox',
+          include: !!config.naming.secondary,
+          onChange: function(newValue, form) {
+            var scheme = config.naming[newValue ? 'secondary' : 'primary'];
+            form.updateField('shortName', {
+              required: scheme.shortNameRequired,
+              disabled: !scheme.shortNameModifiable
+            });
+          }
```
callunity: Would it make sense to move this before the "Short Name" column, as that'll be affected by changes to this checkbox (and could lead to users having previously-valid data be marked invalid, or have to enter in invalid data which will then be validated by checking this box as they fill in left-to-right)?

RELEASE_NOTES.md
```diff
@@ -4,6 +4,9 @@ Changes:
 
   * Added configuration for library aliquot alias generator and validator
   * Added V2 naming scheme
```
callunity: Might be worth changing this to `v2` to match the case expected in `miso.properties`

miso-web/src/main/webapp/scripts/form_project.js
djcooke: Yeah, makes sense. Will move

djcooke:
@callunity: My intention was that when we switch over to using the V2 scheme for all new projects, we'd flip them so that V2 is configured as the primary scheme and OICR is secondary (flipping the bit on all projects at the same time of course)

callunity:
@djcooke that makes sense. It could be helpful to others to add that suggestion to the baremetal-installation-guide in case someone else goes the same route but doesn't think of that possibility (but I think the chances of that happening are realistically pretty small).

RELEASE_NOTES.md
djcooke: Changed

miso-web/src/main/webapp/scripts/form_project.js
djcooke: Moved

djcooke:
@callunity: added to the guide

RELEASE_NOTES.md
callunity: Looks good to me

PR #2193, opened 2020-02-19
GLT-3168: fixed adding too many containers to a run

djcooke:


sqlstore/src/main/resources/db/migration/V8501__unique_constraints.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE Run_SequencerPartitionContainer ADD CONSTRAINT uk_run_position UNIQUE (Run_runId, positionId);
+ALTER TABLE User ADD CONSTRAINT uk_user_loginname UNIQUE (loginName);
```
apmasell: :+1:

PR #2206, opened 2020-03-05
GLT-2822: updated type data docs

djcooke:


docs/user_manual/type_data.md
```diff
@@ -350,6 +354,10 @@ creating new libraries.
 
 MISO administrators can add, edit, and delete indices using the standard interface within the Edit Index Family page.
 
+When creating indices, keep in mind that they are sorted alphabetically in dropdowns. If your indices are named "Index 1,
+Index 2... Index 10, Index 11..." then it may be ideal to add zeros before the numbers so that they sort in the expected
+order. e.g. "Index 01, Index 02... Index 10, Index 11"
+
```
avarsava: Is there a ticket for fixing this sort behaviour?

djcooke: No, and I'm not sure there should be? There are all sorts of naming patterns for indices besides the example given here, and if we sort based on what would work well for one pattern, it won't work well for another pattern. Sorting alphabetically is easy to explain, understand, and adjust.

PR #2207, opened 2020-03-05
GLT-3129: make index checker available externally

djcooke:


djcooke:
@callunity: fixed the NaN issue, and also an error where the green message wasn't being shown when there were no near/duplicates.

I wasn't concerned about the menu getting cut off because it was like that before, but when deployed to readthedocs, there's an ad at the bottom of the menu that prevents it from happening. Fixed that now too though.

PR #2215, opened 2020-03-11
Allow custom printer labels

apmasell:
JIRA: GLT-3123


apmasell:
I will fix the test error. It is relatively minor.

This is really ugly right now. I want to get this into this release, but I will make another PR with an editor for the labels and documentation for that UI.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/printing/PrintableField.java
```diff
@@ -0,0 +1,304 @@
...
+  },
+  SECONDARY_ID {
+
+    @Override
+    public String text(Barcodable barcodable) {
+      return barcodable.visit(getParentSampleField(SampleTissue.class, t -> t.getSecondaryIdentifier()));
+    }
+
+  },
+  SIZE {
```
djcooke: This seems intended to replace `Barcodable#getBarcodeSizeInfo`, which was previously returning concentration and volume information rather for samples, libraries, library aliquots, and pools. I imagine the `dnaSize` field from libraries and aliquots will be wanted on labels for the v2 naming scheme since it's not included in the alias, but the old layouts should still have concentration and volume

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/PrinterRestController.java
```diff
@@ -220,7 +223,7 @@ public long boxPositions(@PathVariable("printerId") Long printerId, @RequestBody
   @PostMapping(headers = { "Content-type=application/json" })
   @ResponseBody
   @ResponseStatus(code = HttpStatus.CREATED)
-  public PrinterDto create(@RequestBody PrinterDto dto) throws IOException {
+  public PrinterDto create(@RequestBody PrinterDto dto) throws Exception {
```
djcooke: It looks like this should still only throw `IOException`?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/printing/PrintableField.java
apmasell: Illinca is fine with the change and no one from GRP has responded, so forgiveness if it comes to it.

PR #2221, opened 2020-03-17
Fix Children button for libraries on samples

apmasell:


RELEASE_NOTES.md
```diff
@@ -4,6 +4,7 @@ Changes:
 
   * Show included projects in Runs list
   * Allow archiving labs and institutes
+  * Fix that caused libraries to be missing from some sample contexts (including Children button and edit Sample page)
```
djcooke: "Fixed error" (otherwise does not sound like a very good fix 😆)?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -1075,11 +1079,16 @@ public ValidationResult validateDeletion(Sample object) {
...
+    try {
+    final List<Library> libraries = libraryService.listBySampleId(object.getId());
+      if (libraries.isEmpty()) {
+        result.addError(new ValidationError(object.getName() + " has " + libraries.size() + " librar"
+            + (libraries.size() > 1 ? "ies" : "y")));
+      }
+    } catch (IOException e) {
+       log.error("Failed to fetch libraries for " + object.getName(), e);
+       result.addError(new ValidationError("Failed to check for libraries."));
     }
```
djcooke: * Condition should be negated, which will fix the failing test too
* Use `ValidationError#forDeletionUsage` to construct standard error message
* Use `Pluralizer#libraries` for `referrer` parameter of `forDeletionUsage`
* throw the `IOException` (already declared in `DeleterService#validateDeletion`) as that shouldn't be reported as a validation error)

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/RelationFinder.java
```diff
@@ -158,9 +163,13 @@ public SampleDto asDto(DetailedSample model) {
...
     }
 
-    public static Stream<Library> searchChildrenLibraries(DetailedSample model) {
-      return model.getChildren().stream()
-          .flatMap(child -> Stream.concat(child.getLibraries().stream(), searchChildrenLibraries(child)));
+    public static Stream<Library> searchChildrenLibraries(DetailedSample model, LibraryService libraryService) {
+      try {
+        return Stream.concat(libraryService.listBySampleId(model.getId()).stream(), model.getChildren().stream().flatMap(child -> searchChildrenLibraries(child, libraryService)));
+      } catch (IOException e) {
+         log.error("Failed to get child libraries.", e);
```
djcooke: throw exception as that will be less confusing than returning nothing to both users and developers

PR #2230, opened 2020-03-24
LDAP Compatibility Fix

avarsava:
There appears to be no guarantee that the implementation of UserDetails we get will be an InetOrgPerson (external user seeing LdapUserDetailsImpl instead). The only methods we use outside of the interface are the Display Name and Email getters, so only use those if instanceof InetOrgPerson.

Still setting up LDAP Server locally to test against, but requesting thoughts on this approach to ensure there aren't obvious pitfalls

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/security/util/LimsSecurityUtils.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/security/util/LimsSecurityUtils.java
@@ -93,6 +93,8 @@ public class LimsSecurityUtils {
       InetOrgPerson person = (InetOrgPerson) ldapUserDetails;
       target.setFullName(person.getDisplayName());
       target.setEmail(person.getMail());
+    } else {
+      target.setFullName(ldapUserDetails.getUsername());
     }
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/security/util/LimsSecurityUtils.java
```diff
@@ -88,8 +89,11 @@ public static void updateFromLdapUser(User target, InetOrgPerson ldapUserDetails
     target.setInternal(roles.contains(MisoAuthority.ROLE_INTERNAL.name()));
 
     target.setPassword(ldapUserDetails.getPassword());
-    target.setFullName(ldapUserDetails.getDisplayName());
-    target.setEmail(ldapUserDetails.getMail());
+    if(ldapUserDetails instanceof InetOrgPerson) {
+      InetOrgPerson person = (InetOrgPerson) ldapUserDetails;
+      target.setFullName(person.getDisplayName());
+      target.setEmail(person.getMail());
+    }
```
djcooke: Will fail to save because fullname isn't nullable. Email is nullable in the DB, but I'm not sure that's intentional as it's required on the Edit User page and there previously hasn't been a way for it to be null. You'll have to check the call hierarchy to see if null email would cause any problems. Maybe you could just use the username as the fullname too, but I'm not sure there's any way to get an email

avarsava: Using the username as full name works. Email is harder since UserDetails doesn't assume an email address is associated with a user. Searching for 'email' across all of MISO it looks as though it's tracked but no actual features use it? Should we make emails optional?

avarsava:
Wasn't able to get LDAP server up, confirmed at least that this does not break AD signon

PR #2235, opened 2020-04-02
GLT-3225: switch to single table Hibernate inheritance

djcooke:


sqlstore/src/main/resources/db/migration_beforeMigrate/00_functions.sql
```diff
@@ -32,6 +32,38 @@ BEGIN
...
+  IF (val1 IS NULL) <> (val2 IS NULL) THEN
+    RETURN TRUE;
+  ELSEIF val1 IS NULL AND val2 IS NULL THEN
+    RETURN FALSE;
+  ELSE
+    RETURN val1 <> val2;
+  END IF;
+END//
+
+DROP FUNCTION IF EXISTS makeChangeMessage//
```
apmasell, line 47: I really like this.

PR #2260, opened 2020-05-19
GLT-3238: added new bulk table builder; implemented for library aliquots

djcooke:


miso-web/src/main/webapp/scripts/bulk.js
```diff
@@ -0,0 +1,1053 @@
+BulkUtils = (function($) {
+
+  /*
+   * BulkTarget structure: {
+   *   getSaveUrl: required function() returning url for bulk save. POST is used for create and PUT for edits
+   *   getUserManualUrl: optional (recommended) function() returning url of specific user manual page to set on 'Help' link
+   *   getCustomActions: optional function() returning array of Custom Actions (see below) to make available while editing
```
apmasell, line 7: > while editing

On the bulk edit page while editing? I'm not sure why I find this weird.

miso-web/src/main/webapp/scripts/lims.js
```diff
@@ -552,7 +552,35 @@ var Utils = Utils
...
+          width: 350,
+          title: title,
+          modal: true,
+          buttons: {},
+          closeOnEscape: false,
+          open: function(event, ui) {
+            jQuery(this).parent().children().children('.ui-dialog-titlebar-close').hide();
+          }
+        });
+        // dialog doesn't appear without this delay for some reason...
```
apmasell, line 576: I've seen similar problems elsewhere. I think it's an event-loop thing.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultLibraryAliquotService.java
```diff
@@ -169,6 +175,44 @@ public long create(LibraryAliquot aliquot) throws IOException {
     return savedId;
   }
 
+  @Override
+  public List<LibraryAliquot> bulkCreate(List<LibraryAliquot> aliquots) throws IOException {
+    List<Long> savedIds = new ArrayList<>();
```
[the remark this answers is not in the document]
djcooke: Ideally, all of the services with bulk save would extend `AbstractBulkSaveService`, but the ones that don't already extend `AbstractSaveService` aren't straight-forward to refactor. The sample, library, and library aliquot services in-particular have a lot of extra complexity.

miso-web/src/main/webapp/scripts/bulk.js
djcooke: Updated. Also fixed incomplete documentation of the API object's `updateField` options

PR #2263, opened 2020-05-21
misc documentation updates

djcooke:


docs/user_manual/barcode_label_printers.md
```diff
@@ -1,8 +1,8 @@
 # Barcode Label Printers
 
 MISO pages can be printed to any printer that is installed on the computer you are using. No extra configuration is
-required to do this. Printing barcode labels to label printers does require some configuration, however.
-
+required to do this. Printing barcode labels to label printers does require some configuration, however. Brady and
+Zebra label printers are supported.
```
avarsava, line 5: Specific model #s?

djcooke: To my knowledge, all should work

avarsava: In that case I think it still might be more accurate list the models we've verified to work and write that the rest should work but have not been tested, but if you do not have that information this is fine

djcooke: Yeah, good idea. Added

avarsava: :+1:

PR #2264, opened 2020-05-22
GLT-3242: allow deselect when adding aliquots to pools/orders

djcooke:


djcooke:
@avarsava: fixed, thanks

PR #2276, opened 2020-06-05
GLT-3270: fixed scientific name required and default values

djcooke:


djcooke:
Closing because most of this fix is no longer needed as the bulk column `initial` property was changed to take the item label instead of value (id). Will open separate PR to fix required field

PR #2291, opened 2020-06-19
GLT-3264: fixed requesting bulk page with many items

djcooke:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/util/MisoWebUtils.java
```diff
@@ -164,4 +165,60 @@ private static boolean checkFile(File path) throws IOException {
...
+   * @param formData
+   * @param required
+   * @return the value mapped to the provided key if it is set and not empty; otherwise null
+   * @throws ClientErrorException if the value is required and missing or empty, or if the value is set and non-empty, but is not a valid
+   *           Integer
+   */
+  public static Integer getIntegerInput(String key, Map<String, String> formData, boolean required) {
+    String stringValue = getStringInput(key, formData, required);
+    try {
+      return stringValue == null ? null : new Integer(stringValue);
```
apmasell, line 218: This constructor is deprecated in newer versions of Java: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Integer.html#%3Cinit%3E(java.lang.String)

djcooke:
@avarsava: fixed. I'd moved a function and missed updating a couple references, so there was a Javascript error

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/util/MisoWebUtils.java
djcooke: changed to `Integer#parseInt`

djcooke: Changed to `Integer#valueOf`

PR #2296, opened 2020-06-30
GLT-3285: added progress bar to bulk save

djcooke:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/component/AsyncOperationManager.java
```diff
@@ -0,0 +1,148 @@
...
+    ObjectNode json = mapper.createObjectNode();
+    json.put("operationId", uuid);
+    json.put("status", status);
+    json.put("completedUnits", operation.getProgress());
+    json.put("totalUnits", operation.getTotalCount());
+    return json;
+  }
+
+  private <R extends Identifiable> String addAsyncOperation(BulkSaveOperation<R> operation) {
+    String uuid = null;
```
apmasell, line 129: You could also allocate a static `AtomicLong` and use `getAndIncrement()`. Not really necessary, but probably faster than even generating a single UUID.

miso-web/src/main/webapp/styles/style-full.css
```diff
@@ -1,3 +1,10 @@
+:root {
```
apmasell, line 1: MISO can't survive on so few colours! :laughing:

PR #2310, opened 2020-07-16
Export UMI status from library for library aliquots in Pinery

apmasell:


djcooke:
Why? It's a library attribute, and already available on the library

apmasell:
In Pinery, everything in sample provenance has UMIs set to false, even things that have it set to true in MISO. I'm assuming that's because provenance is joining against the aliquots and not the libraries.

PR #2311, opened 2020-07-17
[GLT-3286] workset search

avarsava:
Getting 'Library Aliquot cannot be filtered by workset.' on Workset 138 despite all the code being the same, don't understand it

### DIFF
--- a/miso-web/src/main/webapp/scripts/urls.js
+++ b/miso-web/src/main/webapp/scripts/urls.js
@@ -377,9 +377,7 @@ Urls = (function() {
     spreadsheet: libraryRestBase + '/spreadsheet',
     bulkSave: libraryRestBase + '/bulk',
     bulkSaveProgress: idUrlFunction(libraryRestBase + '/bulk'),
-    worksetDatatable: function(worksetId) {
-      return libraryRestBase + '/dt/workset/' + worksetId;
-    },
+    worksetDatatable: idUrlFunction(libraryRestBase + '/dt/workset')
   };
 
   // Library Aliquots
@@ -415,9 +413,7 @@ Urls = (function() {
     children: idUrlFunction(libraryAliquotRestBase + '/children'),
     bulkSave: libraryAliquotRestBase + '/bulk',
     bulkSaveProgress: idUrlFunction(libraryAliquotRestBase + '/bulk'),
-    worksetDatatable: function(worksetId) {
-      return libraryAliquotRestBase + '/dt/workset/' + worksetId;
-    },
+    worksetDatatable: idUrlFunction(libraryAliquotRestBase + '/dt/workset')
   };
 
   // Library Designs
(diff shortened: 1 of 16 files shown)


djcooke:
The `jQueryBackend` in `LibraryAliquotRestController` uses `PoolableElementViewService` rather than `LibraryAliquotService`. The `PaginationFilter` needs to go in `HibernatePoolableElementViewDao` instead of `HibernateLibraryAliquotDao`.

`PoolableElementView` is an alternate view of `LibraryAliquot` that has a more shallow graph and is used to improve performance in some places

avarsava:
> The `jQueryBackend` in `LibraryAliquotRestController` uses `PoolableElementViewService` rather than `LibraryAliquotService`. The `PaginationFilter` needs to go in `HibernatePoolableElementViewDao` instead of `HibernateLibraryAliquotDao`.
> 
> `PoolableElementView` is an alternate view of `LibraryAliquot` that has a more shallow graph and is used to improve performance in some places

Done, thanks!

miso-web/src/main/webapp/scripts/urls.js
```diff
@@ -376,7 +376,10 @@ Urls = (function() {
     bulkDelete: libraryRestBase + '/bulk-delete',
     spreadsheet: libraryRestBase + '/spreadsheet',
     bulkSave: libraryRestBase + '/bulk',
-    bulkSaveProgress: idUrlFunction(libraryRestBase + '/bulk')
+    bulkSaveProgress: idUrlFunction(libraryRestBase + '/bulk'),
+    worksetDatatable: function(worksetId) {
+      return libraryRestBase + '/dt/workset/' + worksetId;
+    },
```
djcooke, line 382: Can shorten to `idUrlFunction(libraryRestBase + '/dt/workset')` (and same for library aliquots and samples below)

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateLibraryAliquotDao.java
```diff
@@ -18,6 +18,7 @@
 import org.springframework.stereotype.Repository;
 import org.springframework.transaction.annotation.Transactional;
 
+import uk.ac.bbsrc.tgac.miso.core.data.Workset;
```
djcooke, line 21: Added unused import

miso-web/src/main/webapp/scripts/urls.js
avarsava: I'm not sure I understand, these URLs need the workset ID, not the sample/library/libraryaliquot ID. will idUrlFunction still work? I'm not sure I understand where it gets 'id' from

djcooke: `idUrlFunction` returns a function that takes the workset ID as a parameter and appends it to the provided string (with a slash in-between), so it has the exact same effect as the code you wrote. It's defined at the top of `urls.js`. The same pattern is repeated many times, so these convenience functions keep the script much more tidy.

djcooke: (`idUrlFunction` doesn't specify or care that the parameter of the returned function is a workset ID. That's just the intention this case)

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateLibraryAliquotDao.java
avarsava: Fixed

miso-web/src/main/webapp/scripts/urls.js
avarsava: Fixed

PR #2316, opened 2020-07-27
[GLT-3300] Fix typo in Printer layout

avarsava:
2barcode -> 2dbarcode


djcooke:
Looks like the typo was made in 3 places - twice [here](https://github.com/miso-lims/miso-lims/blob/develop/sqlstore/src/main/resources/db/migration/V1040__auto_main.sql#L95) and once in the line above. You could replace all instances of `"2barcode"` in the column with `"2dbarcode"`? That would cover cases where the layouts have been modified too.

avarsava:
> Looks like the typo was made in 3 places - twice [here](https://github.com/miso-lims/miso-lims/blob/develop/sqlstore/src/main/resources/db/migration/V1040__auto_main.sql#L95) and once in the line above. You could replace all instances of `"2barcode"` in the column with `"2dbarcode"`? That would cover cases where the layouts have been modified too.

Updated the migration to affect the BASE64 layout too. Not sure how we could affect layouts if they've been modified?

djcooke:
```
UPDATE Printer
SET layout = REPLACE(layout, '"2barcode"', '"2dbarcode"');
```

avarsava:
> ```
> UPDATE Printer
> SET layout = REPLACE(layout, '"2barcode"', '"2dbarcode"');
> ```

Updated

PR #2331, opened 2020-08-05
GLT-3310: fixed inconsistent sorting in paginated lists

djcooke:
Apparently Hibernate always resolves "id" to the primary key, regardless of how it's defined on the entity model


djcooke:
@apmasell: There should be no change in which you can search by ID. It was allowed on anything implementing `HibernatePaginatedDatasource` that returned a non-null value from `propertyForId`, which was everything except `SequencingOrderCompletionDao`. This PR makes it so that everything uses the "id" field to enable this by default (in the interface), and then `SequencingOrderCompletionDao` overrides the default method to disable it.

PR #2340, opened 2020-08-25
GLT-3245: added library batches

djcooke:


djcooke:
@apmasell: I haven't included it in sample provenance at all. That would be a Pinery PR if necessary. Is it required in provenance for Dashi? Or something else?

apmasell:
I'm not sure, but it probably will be in the future. We can wait until we have a reason.

PR #2358, opened 2020-09-22
Long changelog messages

avarsava:
Applying these migrations and setting runpath or description on a run from a 255-char string to a different 255-char string still causes `Caused by: com.mysql.jdbc.MysqlDataTruncation: Data truncation: Data too long for column 'makeChangeMessage('description', OLD.description, NEW.description)' at row 1` or similar

### DIFF
--- a/sqlstore/src/main/resources/db/migration_afterMigrate/05_trigger_array.sql
+++ b/sqlstore/src/main/resources/db/migration_afterMigrate/05_trigger_array.sql
@@ -11,7 +11,7 @@ DROP TRIGGER IF EXISTS ArrayChange//
 CREATE TRIGGER ArrayChange BEFORE UPDATE ON Array
 FOR EACH ROW
   BEGIN
-  DECLARE log_message varchar(500) CHARACTER SET utf8;
+  DECLARE log_message longtext CHARACTER SET utf8;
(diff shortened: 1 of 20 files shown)


djcooke:
I just tried updating the `createChangeMessage` function and `RunChange` trigger, then updating a run alias to a 255-character string, and then updating it again to another 255-character string. It worked, and the changelog entry was generated as expected. My database may be in a weird state though. I can try restoring and and migrating properly tomorrow

djcooke:
This works fine for me. Here's what I did to be sure:

* Copied production to my local DB
* Checked out develop branch
* built develop branch, ran flyway:migrate (via Maven), deployed MISO locally
* attempted to change a run alias to a 255-character string via the UI. It failed with the expected "Data too long" error
* checked out this branch
* built, ran flyway:migrate
* attempted to change a run alias to a 255-character string via the UI. It worked and the changelog entry was generated as expected
* attempted to change a run alias to a different 255-character string via the UI. It worked and the changelog entry was generated as expected

avarsava:
I must have missed a step, rebuilding and _then_ migrating also made it work for me.

djcooke (approved):

apmasell (approved):

PR #2367, opened 2020-10-06
GLT-3358: added dataManglingPolicy to runs

djcooke:
Added as enum instead of boolean to be consistent with how it is on instrument model

### DIFF
--- a/miso-web/src/main/webapp/scripts/form_run.js
+++ b/miso-web/src/main/webapp/scripts/form_run.js
@@ -134,6 +134,18 @@ FormTarget.run = (function($) {
             return item.value;
           }
         }, {
+          title: 'Index Sequencing',
+          data: 'dataManglingPolicy',
+          type: 'dropdown',
+          source: Constants.dataManglingPolicies,
+          getItemLabel: function(item) {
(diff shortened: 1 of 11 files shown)


apmasell (approved):

miso-web/src/main/webapp/scripts/form_run.js
```diff
@@ -133,6 +133,18 @@ FormTarget.run = (function($) {
           getItemValue: function(item) {
             return item.value;
           }
+        }, {
+          title: 'Index Sequencing',
+          data: 'dataManglingPolicy',
+          type: 'dropdown',
+          source: Constants.dataManglingPolicies,
```
apmasell, line 140: It might make sense to be able to set this back to `null` as _Instrument Default_ or something like that.

mlaszloffy (approved):

miso-web/src/main/webapp/scripts/form_run.js
djcooke: You already can - the field is not marked `required` and `nullLabel` is "Instrument Default"

PR #2379, opened 2020-10-27
Allow editing label layouts for printers

apmasell:
JIRA: GLT-3123

### DIFF
--- a/miso-web/src/main/webapp/scripts/labeldesign.js
+++ b/miso-web/src/main/webapp/scripts/labeldesign.js
@@ -1,0 +1,606 @@
+/*
+ * Copyright (c) 2012. The Genome Analysis Centre, Norwich, UK
+ * MISO project contacts: Robert Davey @ TGAC
+ * *********************************************************************
+ *
+ * This file is part of MISO.
+ *
+ * MISO is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MISO is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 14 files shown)


miso-web/src/main/webapp/scripts/labeldesign.js
```diff
@@ -0,0 +1,606 @@
+/*
+ * Copyright (c) 2012. The Genome Analysis Centre, Norwich, UK
+ * MISO project contacts: Robert Davey @ TGAC
```
avarsava, line 3: Is this still an accurate contact for new files?

avarsava (approved):

miso-web/src/main/webapp/scripts/labeldesign.js
apmasell: Good question. Dillan, what are we doing for new files?

djcooke: Mostly not including it at all

miso-web/src/main/webapp/scripts/list_printer.js
```diff
@@ -35,15 +35,78 @@ ListTarget.printer = {
...
+          property: "width",
+          value: items[0].width
+        }, {
+          type: "float",
+          label: "Height (mm)",
+          property: "height",
+          value: items[0].height
+        }], function(settings) {
+          if (!settings.name) {
+            Utils.showOkDialog('Create Printer', ['A printer needs a name.']);
```
djcooke, line 67: Can set `required: true` in the field specifications above instead. For width and height too

miso-web/src/main/webapp/scripts/list_printer.js
```diff
@@ -84,13 +143,6 @@ ListTarget.printer = {
               Utils.showOkDialog('Create Printer', ['A printer needs a name.']);
```
djcooke, line 143: The fields here should all be required too

miso-web/src/main/webapp/scripts/labeldesign.js
apmasell: I like this solution best.

djcooke (approved):

PR #2381, opened 2020-10-30
[GLT-3372] Write Design Code to Tracking List spreadsheet

avarsava:


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/spreadsheet/LibraryAliquotSpreadSheets.java
```diff
@@ -25,6 +21,7 @@
       Column.forString("Library Alias", libraryAliquot -> libraryAliquot.getLibrary().getAlias()), //
       Column.forString("Library Barcode", libraryAliquot -> libraryAliquot.getLibrary().getIdentificationBarcode()), //
       Column.forString("Library Type", libraryAliquot -> libraryAliquot.getLibrary().getLibraryType().getDescription()), //
+      Column.forString("Library Design", libraryAliquot -> LimsUtils.isDetailedLibraryAliquot(libraryAliquot)? ((DetailedLibraryAliquot)libraryAliquot).getLibraryDesignCode().getCode(): ""), //
```
djcooke, line 24: There's another method that takes a boolean indicating whether the column is for detailed sample mode only (`Column.forString(String, boolean, Function)`). Should use that instead so that there aren't extra blank columns added for plain sample mode. Same in `LibrarySpreadsheets`

avarsava: Updated!

djcooke (approved):

apmasell (approved):

PR #2407, opened 2020-12-02
GLT-3370: replaced project clinical flag with pipeline

djcooke:


djcooke:
Forgot to make Pinery changes

PR #2418, opened 2021-01-05
GLT-3414: allow longer times received and tube number

djcooke:


apmasell (approved):

avarsava (approved):
Values larger than 2147483647 cause JsonParseException that causes 'Oh snap!' on the Edit Sample page without indicating what the issue is

djcooke:
@avarsava: added validation for max values

PR #2429, opened 2021-01-14
[GLT-3421] Change unique alias to unique alias FOR PROJECT in Subproject

avarsava:
UK_n88wwxd7kv4q0m2xhfek9xl70 is a very ugly name but it comes from V0010 so it should be consistent across deployments


djcooke:
The uniqueness is validated in`DefaultSubprojectService` too

apmasell (approved):

avarsava:
> The uniqueness is validated in`DefaultSubprojectService` too

I think I fixed it. Local MISO still won't deploy properly so I can't test

djcooke:
The service should still validate the field for uniqueness so that the user gets a useful message. It just needs to check project+alias instead of only alias. `SubprojectDao#getByAlias` can be modified as it's not used by anything else.

There is also already a V8505 migration, so this probably will fail to build.

Let me know if you need help getting MISO deployed

djcooke (approved):

PR #2443, opened 2021-02-01
[GLT-3436][GLT-3271] Allow wildcard and null searches

djcooke:


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/util/DbUtils.java
```diff
@@ -41,44 +48,108 @@
...
+    String sanitized = sanitizeQueryString(query.getText());
+    MatchMode matchMode = getMatchMode(query);
+    if (searchProperties.length == 1) {
+      return textRestriction(searchProperties[0], sanitized, matchMode);
+    } else {
+      Criterion[] criteria = new Criterion[searchProperties.length];
+      for (int i = 0; i < searchProperties.length; i++) {
+        criteria[i] = textRestriction(searchProperties[i], sanitized, matchMode);
+      }
+      if (sanitized == null) {
```
apmasell, line 105: This is slightly confusing...if the query is `IS NULL` and we have multiple properties, then it becomes _all are null_, but if we're searching for text, then _any field has this text_?

apmasell (approved):

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/util/DbUtils.java
djcooke: For searches that check multiple fields, if you're looking for nulls, it should be null in all those fields. e.g. "external" checks both `externalName` and `secondaryIdentifier`. Searching "external:" finds items with neither specified this way. If it were "any is null" then the query would always return everything.

apmasell: Makes sense. Can you stick this in a comment?

djcooke: done

avarsava (approved):

PR #2471, opened 2021-03-17
[GLT-3467] Lengthen kitLot

avarsava:


apmasell (approved):

djcooke (commented):
Can we keep the limit a bit lower? 255 maybe?

avarsava:
> Can we keep the limit a bit lower? 255 maybe?

done

djcooke (approved):

PR #2475, opened 2021-03-30
[GLT-3474] updated to Java 11

djcooke:


apmasell (approved):
Great. This means we can modularise some of MISO dependencies (Pinery comes to mind).

avarsava (approved):

PR #2482, opened 2021-04-08
[GLT-3458] revised advanced search syntax

djcooke:


callunity (approved):
That is indeed a lot of slashes!

apmasell (approved):

PR #2489, opened 2021-04-16
[GLT-3494] create receipt transfers

djcooke:


miso-web/src/main/webapp/WEB-INF/pages/editTransfer.jsp
```diff
@@ -18,6 +18,34 @@
...
+  </p>
+  <table>
+    <thead>
+      <tr>
+        <th>Type</th><th>Sender</th><th>Recipient</th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr>
+        <td>Receipt</td><td>External Lab</td><td>Internal Group</td>
```
callunity, line 37: This is not a blocking question, merely a curiosity: Is the distinction between "External Lab" here and "External Entity" (in Distribution below) meaningful?

djcooke: Yes, "external lab" (for receipt) must be selected from a defined list, whereas "external entity" is a free text entry and could be a person or lab or raccoon, etc. I think this becomes clearer when you look at the controls below as well.

I'm always open to suggestions for wording to make things clearer though! 😉

avarsava (approved):

callunity (approved):

miso-web/src/main/webapp/WEB-INF/pages/editTransfer.jsp
callunity: _adds "set default text in the external entity field to :raccoon: " to my MISO April Fool's wishlist_

PR #2492, opened 2021-04-20
Update dependencies

djcooke:


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/context/MisoAppListener.java
```diff
@@ -86,16 +89,22 @@ public void contextInitialized(ServletContextEvent event) {
...
-    PropertyConfigurator.configure(application.getRealPath("/") + "/WEB-INF/log4j.miso.properties");
-
     XmlWebApplicationContext context = (XmlWebApplicationContext) WebApplicationContextUtils.getRequiredWebApplicationContext(application);
 
     // resolve property file configuration placeholders
     MisoPropertyExporter exporter = (MisoPropertyExporter) context.getBean("propertyConfigurer");
     Map<String, String> misoProperties = exporter.getResolvedProperties();
 
+    // load logging system manually so that we can inject the baseDirectory property
+    ThreadContext.put("baseDirectory", misoProperties.get("miso.baseDirectory"));
```
apmasell, line 99: Why is this going in a thread local?

apmasell (approved):

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/context/MisoAppListener.java
djcooke: log4j2 has several methods of [property lookup](https://logging.apache.org/log4j/2.x/manual/lookups.html). It wasn't my first choice, but ThreadContext map is the one I got to work. Spring lookup seems to only work for Spring Boot.

djcooke: Actually, I could set a system property and use that instead. Probably makes more sense

apmasell: Worth putting a comment about it too.

djcooke: Beyond the existing comment about injecting the property?

apmasell: Yeah. It's unclear that the property files are the consumers.

djcooke: Updated

avarsava (approved):

PR #2493, opened 2021-04-20
[GLT-3495] allow QC notes for any status

djcooke:


apmasell (approved):

callunity (approved):
I wonder how long til a user requests to filter by QC note text...

PR #2496, opened 2021-04-21
[GLT-3320] fixed select all on all pages for static lists

djcooke:


apmasell (approved):
I can't believe we had Flash.

avarsava (approved):

PR #2500, opened 2021-04-26
[GLT-3503] Revised release notes generation

djcooke:


apmasell (approved):

compact-changelog.sh
```diff
@@ -0,0 +1,62 @@
...
+ echo "ERROR: Bad command. Usage example: ${0} v1.2.3"
+ exit 1
+fi
+
+CHANGE_DIR=changes
+CHANGELOG=RELEASE_NOTES.md
+RELEASE_VERSION=$1
+DATE=$(date +%Y-%m-%d)
+
+BAD_FILES=$(find changes -depth 1 -not -name "README.md" -and -not -name "add_*" \
```
callunity, line 16: My Ubuntu 18 version of bash is taking issue with the `-depth` argument, telling me 
```
find: paths must precede expression: `1'
```

I've added an `add_change_note.md` file to the `changes` dir and tried to run it as `./compact-changelog.sh v1.2.3`

changes/README.md
```diff
@@ -0,0 +1,50 @@
...
+underscore. e.g. `fix_saving_samples.txt` or `add_GLT-1234.txt`. The portion after the underscore
+is not important, but something descriptive makes them easier to find if they need to be modified.
+
+## Message Format
+
+Each change must be in a separate file, but a single change may contain multiple lines of text. A
+bullet point will be added before the first line when writing to the changelog, so the first line
+in a change file should not have a bullet point. Subsequent lines may use bullet points, which will
+be indented as a second-level list. For example, the following is a well-formatted change, which
+could be named `add_field_x.txt`:
```
callunity, line 32: Worth mentioning that change messages can be written in Markdown (though there's no guarantee that the script will correctly parse more complex features)?

djcooke: 👍 Updated to say that it's markdown

compact-changelog.sh
djcooke: Changed to use mindepth+maxdepth instead, which should work everywhere

callunity (approved):

PR #2503, opened 2021-04-26
[GLT-3505] fixed toolbar display bug

djcooke:
Once approved, I'll hotfix stage and prod by adding this to the main `style.css`:

```
.paging_full_numbers {
    height: auto !important;
}
```

Also going to make a ticket to sort out all the js/css imports so they're not linked in so many different JSPs. The removed css imports in this PR are already included on every page via `header.jsp`, so they were being included twice and overriding anything in the main `style.css`, thus `!important` being necessary for the simpler hotfix.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,6 +10,9 @@ Changes:
 
 * Fix issue saving Oxford Nanopore containers
 
+Known Issues:
+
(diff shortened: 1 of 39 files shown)


callunity (approved):
Yikes that is a lot of JSP imports

avarsava (approved):

PR #2506, opened 2021-04-27
removed commons IO dependency

djcooke:
Instead of dependabot PR that updates the library


avarsava (approved):

callunity (approved):

callunity:
Those failing tests look like dependency issues...

djcooke:
> Those failing tests look like dependency issues...

Other dependencies used conflicting versions of commons-io. Removed/excluded them, so it should be fixed now.

PR #2508, opened 2021-05-04
[GLT-3190] added bulk controls to Identity Search Tool

djcooke:
Open to suggestions for renaming the "Show Children" button, as the table's options are "Show Children" and "Children." I think "Show" is too vague, and using a synonym for children such as "Show Descendants" isn't really any better.


callunity (approved):

callunity:
The "Identity Results" table is falling off the right side of the page; perhaps the "Search for X' buttons and friends can go on top of or below the search box, to pull the Identity Results table in beside the search box?
What about "Filter Children on table below" and "Act on Children"? A bit wordy, but would help indicate to the user that the results for the current "Show Children" button are displayed elsewhere. (I'm hesitant to suggest a jump to the table below, but would serve the same purpose..)

avarsava (approved):

djcooke:
@callunity I moved all of the inputs into a single column within a flexbox, so it scales better to a narrower window now. Renamed the "Show Children" button to "Filter Below," which I think helps to differentiate the buttons, and described them better in the help text in addition. I don't want to rename the "Children" button, as that would make it inconsistent with the same function on other pages

callunity:
That displays much nicer on my narrower window. I like "Filter Below".

PR #2520, opened 2021-05-25
[GLT-3532] fix NPE when generating sample sheets

djcooke:


avarsava (approved):

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/SampleSheet.java
```diff
@@ -274,34 +274,33 @@ public String createSampleSheet(Run run, User user) {
...
     }
     final boolean needsSuffix = indices.size() > 1;
     return indices.stream()
         .map(index -> makeColumns(needsSuffix, model, partitionNumber, partitionBarcode, aliquot, index, userName)
             .collect(Collectors.joining(",")));
+  }
 
+  private Set<String> getSequences(Index index, InstrumentModel model) {
+    Set<String> sequences = index.getFamily().hasFakeSequence() ? index.getRealSequences() : Collections.singleton(index.getSequence());
+    if (model.getDataManglingPolicy() == InstrumentDataManglingPolicy.I5_RC) {
```
apmasell, line 300: I think this also needs to check if the index is in position 2 since the position 1 shouldn't be reverse complemented.

PR #2521, opened 2021-05-26
[GLT-3509] added kits to library aliquots

djcooke:


avarsava (approved):

callunity (commented):
I edited some TS library aliquots and changed their kits and targeted sequencing. When I went to save them, I got the message "An unexpected error has occurred" (but the error didn't write to my Tomcat logs so I'm not really able to see what happened...)

callunity (approved):

callunity:
> I edited some TS library aliquots and changed their kits and targeted sequencing. When I went to save them, I got the message "An unexpected error has occurred" (but the error didn't write to my Tomcat logs so I'm not really able to see what happened...) Are you able to reproduce this or is it a local issue?

It was an issue with my single-barrelled naming scheme. Adding the second naming scheme fixed the issue.

PR #2534, opened 2021-06-07
[GLT-3509] allow archiving kits

djcooke:


callunity (approved):
Non-blocking request: tests for at least one of these paths?

avarsava (approved):

djcooke:
> Non-blocking request: tests for at least one of these paths?

If you mean to test archived items being excluded from the dropdowns, it would be ideal to have javascript tests for that. While I'd like to add javascript tests at some point, MISO does not have that set up as of yet. I don't think testing this type of thing with the UI integration tests is a good idea as those tests run very slowly.

callunity:
That makes sense; I'd forgotten that the dropdowns on the editX.jsp pages were also populated by javascript.

PR #2536, opened 2021-06-14
[GLT-3537] added storage labels

djcooke:


avarsava (approved):

callunity (approved):
I really like the convention of having `to` and `from` within the DTO classes themselves.

PR #2540, opened 2021-06-21
switch to GitHub actions for CI

djcooke:
I think having all the conditional steps is slightly nicer than having separate jobs with all of the setup steps repeated. There doesn't seem to be a good way to share the setup steps between multiple jobs.


avarsava (approved):

callunity (approved):

djcooke:
Separated into multiple workflows because it's not possible to rerun single jobs within a workflow...

PR #2563, opened 2021-08-12
GLT-3567 metrics expansion

djcooke:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultMetricService.java
```diff
@@ -67,21 +82,35 @@ protected void authorizeUpdate(Metric object) throws IOException {
...
-        && metricDao.getByAliasAndCategory(object.getAlias(), object.getCategory()) != null) {
-      errors.add(ValidationError.forDuplicate("metric", "alias"));
-    }
+    // no checks
   }
 
   @Override
   protected void applyChanges(Metric to, Metric from) throws IOException {
     to.setAlias(from.getAlias());
-    to.setCategory(from.getCategory());
```
callunity, line 82: Why remove the category assignment?

callunity (approved):

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultMetricService.java
djcooke: Prevents changing category when subcategory is already set. Incompatible changes could be prevented with validation instead, but I think it would also be confusing if the category changed for metrics that are already linked to assays. This way, you have to delete and re-create the metric, which forces you to look at the assays it's linked to as well

docs/user_manual/type_data.md
```diff
@@ -719,5 +721,25 @@ Examples:
...
 is not used by any existing assays.
+
+
+## Metric Subcategories
+
+Metric subcategories can be used to subdivide metrics within the major the metric categories. If your site uses
+[detailed sample mode](../site_configuration/#detailed-sample-mode), a library design code may be specified to indicate
+that the subcategory is only applicable to matching libraries.
+
+MISO administrators can add, edit, and delete metric subcategoriess using the standard interface. A metric subcategory
```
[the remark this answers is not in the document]
djcooke: fixed

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultMetricService.java
callunity: That sounds reasonable to me

avarsava (approved):

PR #2565, opened 2021-09-01
GLT-3569: changed partition QC type analysis/pipeline labels

djcooke:


miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/PartitionQCTypeDto.java
```diff
@@ -56,11 +56,11 @@ public String getDetailedLabel() {
     sb.append(getDescription());
     sb.append(" (order ");
     if (!isOrderFulfilled()) sb.append("not ");
-    sb.append("fulfilled, analysis ");
+    sb.append("fulfilled, ");
     if (isAnalysisSkipped()) {
```
avarsava, line 60: It looks like this is getting renamed everywhere except this method name?

avarsava (approved):

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/PartitionQCTypeDto.java
djcooke: The field name has not been changed - only the labels

callunity (approved):

Issue #2445, opened 2021-02-03
Show full text and abbreviation for "Code" field on the "Edit Libraries" page


peterk87:
Hello,

On the "Edit Libraries" page, the "Code" field only displays the abbreviation and not the full text for the library design. Would it be possible to show both? New MISO users may not be familiar with the abbreviations. 

Thanks!

djcooke:
Yes, that is a good suggestion, thanks! We already show both on the single Edit Library page, and should update the bulk page to be similar.

djcooke:
https://github.com/miso-lims/miso-lims/pull/2594

PR #2594, opened 2021-12-20
[#2445] show description in library design code column

djcooke:


callunity (approved):
Approved once test is fixed :)

avarsava (approved):

PR #2604, opened 2022-01-07
[GLT-3256] updated stains to new bulk interface

djcooke:


callunity (approved):
I'm really enjoying seeing all these service transition to the new bulk interface. This looks great!

Issue #2621, opened 2022-01-26
flyway migration issue( Migrating schema `lims` to version 0320 - auto main)


Owen2015:
Hi, in my first installation of miso, error occur in the database migration part.
I use mariadb 10.2

Flyway Community Edition 5.2.4 by Boxfuse
Database: jdbc:mariadb://106.53.98.28:3306/lims (MySQL 10.2)
Executing SQL callback: beforeValidate
Successfully validated 167 migrations (execution time 00:00.644s)
Executing SQL callback: beforeMigrate
Creating Schema History table: `lims`.`flyway_schema_history`
Current version of schema `lims`: << Empty Schema >>
WARNING: outOfOrder mode is active. Migration of schema `lims` may not be reproducible.
Migrating schema `lims` to version 0001 - base schema
Migrating schema `lims` to version 0002 - miso type data

and the error occurs at version 320
Migrating schema `lims` to version 0320 - auto main
ERROR: Migration of schema `lims` to version 0320 - auto main failed! Please restore backups and roll back database and code!
ERROR: 
Migration V0320__auto_main.sql failed
-------------------------------------
SQL State  : 42000
Error Code : 1064
Message    : (conn=20) You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'JSON), '$.host'), 'pin', JSON_EXTRACT(CAST(configuration AS JSON), '$.passwor...' at line 2
Location   : db/migration/V0320__auto_main.sql (/opt/data/work/software/flyway-5.2.4/file:/opt/data/work/software/flyway-5.2.4/jars/sqlstore-1.39.1.jar!/db/migration/V0320__auto_main.sql)
Line       : 67
Statement  : -- StartNoTest
UPDATE Printer SET backend = 'BRADY_FTP', configuration = JSON_OBJECT('host', JSON_EXTRACT(CAST(configuration AS JSON), '$.host'), 'pin', JSON_EXTRACT(CAST(configuration AS JSON), '$.password'))  WHERE backend = 'FTP'

djcooke:
I still need to investigate further, but I've just tried migrating a new database on MariaDB 10.2.41 and got the same error.

djcooke:
Closing the issue as the fix is in now. Let us know if you have any more trouble

PR #2624, opened 2022-01-27
[#2621] fix DB migration for MariaDB 10.2.41

djcooke:
MariaDB doesn't have a JSON datatype, but `JSON_EXTRACT` works on strings in MySQL too, so the `CAST` is not necessary


callunity (approved):
Works well on my fresh MariaDB install!

PR #2627, opened 2022-02-01
[GLT-3257] updated library templates to new bulk interface

djcooke:


callunity (approved):
Approved, and I know you'll fix the tests :)

PR #2630, opened 2022-02-04
[GLT-3632] set nonStandardAlias on libraries switching naming schemes

djcooke:


callunity (approved):

djcooke:
@avarsava Can check the database - `Project.secondaryNaming`, `TRUE`=secondary scheme, `FALSE`=primary scheme. The issue that this addresses happens when a sample is moved between projects with different naming schemes and that sample has libraries. There are two related warnings.

* A confirmation dialog is shown when moving samples between projects to make sure the user is aware that libraries will be affected
* A warning is displayed while editing and while propagating from items that have non-standard aliases. The warning will not be shown while editing a library for the first time after it was moved because the library itself was not edited in that process, and so it hasn't been validated and had the non-standard alias flag set yet. The warning will be shown after saving the library.

avarsava (approved):

PR #2634, opened 2022-02-14
[GLT-3635] improved Edit Requisition page performance

djcooke:


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultQcStatusService.java
```diff
@@ -158,12 +158,6 @@ public void update(QcStatusUpdate update) throws IOException {
       throwIfNull("Run-library", aliquotView);
       LibraryAliquot aliquot = libraryAliquotService.get(update.getIds()[2]);
       RunPartitionAliquot runLib = runPartitionAliquotService.get(run, partition, aliquot);
-      if (runLib == null) {
```
callunity, line 161: Why will runLib never be null here?

djcooke: Previously, the DAO would only return items that exist in the database. Every time runlibs were pulled from the DAO, the caller would have to construct any runlibs that didn't exist already. I changed this so that the DAO does the construction. This means that for any valid run+partition+aliquot relationship, the DAO will return a runlib, whether it has already been saved or not

avarsava (approved):

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultQcStatusService.java
djcooke: None in MISO - this is a special case. The problem is that the entity represents a relationship that exists whether or not the entity does. The consumer cares about the relationship though, and will never care whether the entity exists in the database. The options I see:

1. We could create/destroy the entity when the relationship is created/destroyed, but that would have to be handled in many places - adding/removing aliquots from a pool, adding/removing a pool from a container, adding/removing a container from a run, etc. I could see this causing bugs due to missed cases, and it would also create more overhead within all of those cases
2. We only create the entity when it is needed (e.g. setting QC or run purpose) and the DAO only returns saved entities, but the consumer will **always** want an entity for each existing relationship. This is how it was previously, and means that the consumer must always construct missing ones, which involves more lookups in separate database sessions and is a lot more confusing in my opinion
3. We only create the entity when it is needed (e.g. setting QC or run purpose) and the DAO returns an entity for each existing relationship, constructing any entity that does not yet exist. The caller gets what they want with no additional work, and the lookup is more optimized than the caller would be capable of. This allows you to assume the entity exists for any existing relationship, which I would argue is more intuitive everywhere except within the DAO itself

callunity: Thanks, I can see why you went with option 3. Would you be able to add a comment in the DAO explaining some of this reasoning?

djcooke: Yep, good idea - added. Also fixed missing check for valid relationship in `get` and added a couple more tests

callunity (approved):
This looks great!

PR #2644, opened 2022-03-08
[GLT-3642] added assay tests to Pinery

djcooke:
Requires Pinery release and updating `pinery.version` in `pinery-miso/pom.xml` before merging

### DIFF
--- a/pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/MisoClient.java
+++ b/pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/MisoClient.java
@@ -24,6 +24,7 @@ import java.util.stream.Collectors;
 
 import javax.sql.DataSource;
 
+import ca.on.oicr.pinery.lims.*;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.text.WordUtils;
 import org.springframework.jdbc.core.JdbcTemplate;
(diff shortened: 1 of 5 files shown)


avarsava (approved):

pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/MisoClient.java
```diff
@@ -1443,6 +1438,25 @@ public MisoBoxPosition mapRow(ResultSet rs, int rowNum) throws SQLException {
     return assay;
   };
 
+  private static final RowMapper<AssayTest> assayTestRowMapper = (rs, rowNum) -> {
+    AssayTest test = new DefaultAssayTest();
+    test.setName(rs.getString("name"));
+    test.setTissueType(rs.getString("tissueType"));
+    Boolean negate = rs.getBoolean("negateTissueType");
+    if (!rs.wasNull()) {
```
callunity, line 1446: It [looks like the column is not null](https://github.com/miso-lims/miso-lims/blob/develop/sqlstore/src/main/resources/db/migration/V8501__assay_tests.sql#L5), is this check still needed?

callunity (approved):

pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/MisoClient.java
djcooke: Ah right - thanks for noticing this! The check is wrong as written, but there should be a different check to handle plain sample mode as both of these are specific to detailed sample mode. I'll remove these `wasNull` checks and instead only set the fields if tissue type is not null (tissue type is required in detailed sample mode)

callunity: Ah, that makes sense!

PR #2680, opened 2022-06-16
Add test for WorkflowStep.throwUnexpectedInput

lacinoire:
Hey 😊
I want to contribute the following test:

Test that a `java.lang.IllegalArgumentException` and a `java.lang.IllegalArgumentException` is thrown when the parameter `stepNumber` of the method `processInput` is set to `1.0`.
This tests the method [`WorkflowStep.throwUnexpectedInput`](https://github.com/lacinoire/miso-lims/blob/16e88bc983c6b37be3eefb773f7cf6e24ed88813/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/workflow/WorkflowStep.java#L85).
This test is based on the test [`testChangeToExistingContainer`](https://github.com/miso-lims/miso-lims/blob/16e88bc983c6b37be3eefb773f7cf6e24ed88813/core/src/test/java/uk/ac/bbsrc/tgac/miso/core/data/workflow/impl/LoadSequencerWorkflowTest.java#L87).

Curious to hear what you think!

Also, is the description I provided of the test helping you to answer to this pull request? Why (not)?

(I wrote this test as part of a research study at TU Delft. [Find out more](https://github.com/lacinoire/lacinoire/blob/main/README.md))

---

Jira ticket or GitHub issue: I did not open one as this is a minor addition to the tests

- [ ] Includes a change file: no, as I would not call this a "notable" change relvant to users. Happy to add one if you disagree!


djcooke (commented):
Thanks for adding the test! The reason for this exception is that the data type (SequencerPartitionContainerProgressStep) is not expected for step number 1. Can you add a comment or revise the failure message to clarify this?

lacinoire:
Hey @djcooke,
Thank you for having a look!
I have added a comment, to explain why the behavior is unexpected and throws an exception:

`// for step 1, a container step is not expected and should be rejected with an IllegalArgumentException`

callunity (approved):

PR #2682, opened 2022-06-20
[GLT-3705] added password length and complexity requirements

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3705

- [x] Includes a change file

### DIFF
--- a/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultUserService.java
+++ b/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultUserService.java
@@ -3,7 +3,9 @@ package uk.ac.bbsrc.tgac.miso.service.impl;
 import java.io.IOException;
 import java.util.List;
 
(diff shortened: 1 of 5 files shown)


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultUserService.java
```diff
@@ -72,7 +98,7 @@ public long update(User user) throws IOException {
     }
     if (securityManager.isPasswordMutable()) {
       if (user.getPassword() != null) {
-        original.setPassword(user.getPassword());
+        user.setPassword(validateAndEncodePassword(user.getPassword(), false));
```
callunity, line 101: Should this be `original.setPassword()`? It's `original` that's saved below

djcooke: Yep! fixed

callunity (approved):

avarsava (approved):

Issue #2684, opened 2022-06-22
Spring beans vulnerability ROOT.war


jayramr:
Hi there,

Our IT found and raised the spring beans  vulnerability in the current docker image version of miso. 

CVE--> cve-2022-22965 -> https://nvd.nist.gov/vuln/detail/cve-2022-22965

Spring Beans 4.3.24 (/var/lib/docker/overlay2/9c56f858b4efc325b528ea5b5ae9be3e5558b0a4bbb5607358a75a03575f26ab/merged/usr/local/tomcat/webapps/ROOT.war)
Spring beans file discovered inside war file

Below is the resolution as per our IT

---
Upgrade to non vulnerable version of spring
Download and apply the patch from: https://spring.io/blog/category/releases
---

Kindly let us know, if this is not relevant. 

Thanks
Jay

PR #2685, opened 2022-06-27
[#2684] updated to latest Spring and Hibernate versions

djcooke:
Some notes...

* Disabled logging of Hibernate deprecation warnings. The Hibernate criteria API has long been deprecated in favour of the JPA criteria API. This is unfortunate, since the Hibernate criteria API was much much better... MISO uses the Hibernate criteria API extensively, and Hibernate now logs a warning every time you invoke it. If we ever update to the JPA criteria API, it will take months of effort.
* There was some remaining log4j1 config that I had to rip out. I tested to ensure that logging to `miso_debug.log` is still working as before
* Default behaviour of `GenerationType.AUTO` used to be `IDENTITY`, but that changed, so I had to change everything to specify `IDENTITY`

Jira ticket or GitHub issue: https://github.com/miso-lims/miso-lims/issues/2684

- [x] Includes a change file

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -447,17 +447,6 @@
         </exclusions>
       </dependency>
       <dependency>
-        <!-- Required to make Hibernate 4 logging compatible with log4j2 -->
-        <groupId>org.jboss.logging</groupId>
-        <artifactId>jboss-logging</artifactId>
-        <version>3.4.1.Final</version>
-      </dependency>
-      <!--<dependency>
-        <groupId>org.hibernate.javax.persistence</groupId>
-        <artifactId>hibernate-jpa-2.1-api</artifactId>
-        <version>1.0.2.Final</version>
-      </dependency>-->
-      <dependency>
         <groupId>javax.annotation</groupId>
         <artifactId>javax.annotation-api</artifactId>
         <version>1.2</version>
--- a/pom.xml
+++ b/pom.xml
@@ -111,10 +111,10 @@
   </distributionManagement>
   <properties>
     <skipUTs>${skipTests}</skipUTs>
-    <spring-version>4.3.24.RELEASE</spring-version>
+    <spring-version>5.3.21</spring-version>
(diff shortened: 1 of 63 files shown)


pom.xml
```diff
@@ -447,11 +452,11 @@
         <artifactId>jboss-logging</artifactId>
         <version>3.4.1.Final</version>
       </dependency>
-      <dependency>
+      <!--<dependency>
```
avarsava, line 455: why only commenting this one out?

avarsava (approved):

pom.xml
djcooke: Will remove. Thanks!

pom.xml
```diff
@@ -447,11 +452,11 @@
         <artifactId>jboss-logging</artifactId>
         <version>3.4.1.Final</version>
       </dependency>
-      <dependency>
```
callunity, line 450: line 450 "" - might be able to be removed?

pom.xml
```diff
@@ -111,10 +111,10 @@
   </distributionManagement>
   <properties>
     <skipUTs>${skipTests}</skipUTs>
-    <spring-version>4.3.24.RELEASE</spring-version>
+    <spring-version>5.3.21</spring-version>
     <spring-security-version>5.0.12.RELEASE</spring-security-version>
     <spring-integration-version>4.3.20.RELEASE</spring-integration-version>
```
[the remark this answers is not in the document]
djcooke: Not that I'm aware of. The reported vulnerability was in spring-beans. Spring integration is versioned separately from most Spring modules too

pom.xml
djcooke: Yep, a slightly newer version is now included as a dependency of hibernate core, so this no longer needs to be managed here. Removed and tested - ensured that hibernate messages still appeared in `miso_debug.log`

callunity (approved):
:+1: once tests pass

PR #2702, opened 2022-09-28
[GLT-3774] remove AD email domain from login name

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3774

- [x] Includes a change file

### DIFF
--- a/changes/note_change_ad_username.md
+++ b/changes/note_change_ad_username.md
@@ -31,5 +31,5 @@ new user instead of the old. e.g.
 UPDATE Sample SET creator = @newUser WHERE creator = @oldUser;
 ```
 
-Substitute `Sample` and `creator` for with the table and field names specified
-in the error. Try deleting again, and repeat until deletion is successful.
+Substitute `Sample` and `creator` with the table and field names specified in
+the error. Try deleting again, and repeat until deletion is successful.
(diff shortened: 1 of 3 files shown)


changes/note_change_ad_username.md
```diff
@@ -0,0 +1,35 @@
...
+SELECT userId INTO @oldUser FROM User WHERE loginName = 'person@example.com';
+SELECT userId INTO @newUser FROM User WHERE loginName = 'person';
+
+DELETE FROM User_Group WHERE users_userId = @oldUser;
+DELETE FROM User WHERE loginName = '@oldUser';
+```
+
+If there is anything associated with that user, you will see a foreign key
+violation error. You'll then have to update any records to associate with the
+new user instead of the old. e.g.
```
callunity, line 28: What do you think about this clarification of the order of operations?
"You'll then have to update" -> "First, associate any records with the new user instead of the old, e.g. {code} and then reattempt the deletion"
Alternatively, what do you think about doing the association first (are there instances where it would be undesirable to merge these user IDs?) and then the delete?

callunity (approved):

changes/note_change_ad_username.md
```diff
@@ -0,0 +1,35 @@
...
+
+```
+UPDATE Sample SET creator = @newUser WHERE creator = @oldUser;
+```
+
+Substitute `Sample` and `creator` for with the table and field names specified
```
avarsava, line 34: "for with"?

avarsava (approved):

PR #2719, opened 2022-11-16
[GLT-3803] auto-select identities based on single external name match

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3803

- [x] Includes a change file

### DIFF
--- a/miso-web/src/main/webapp/scripts/bulk_sample.js
+++ b/miso-web/src/main/webapp/scripts/bulk_sample.js
@@ -641,9 +641,10 @@ BulkTarget.sample = (function($) {
         includeSaved: targetCategory === 'Identity',
         required: true,
         onChange: function(rowIndex, newValue, api) {
+          var enteredNames = separateExternalNames(newValue);
           if (targetCategory === 'Identity') {
             return;
-          } else if (!newValue) {
+          } else if (!enteredNames || !enteredNames.length) {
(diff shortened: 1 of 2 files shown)


miso-web/src/main/webapp/scripts/bulk_sample.js
```diff
@@ -1337,4 +1339,25 @@ BulkTarget.sample = (function($) {
...
+      return [];
+    }
+    return names.split(',').map(function(name) {
+      return name.trim();
+    }).filter(function(name) {
+      return !!name;
+    });
+  }
+  function anyMatch(arr1, arr2) {
+    for (var i1 = 0; i1 < arr1.length; i1++) {
```
callunity, line 1353: What do you think about making this implementation? (I find it more semantic/readable, and want to know what you think)
```
let matches = arr1.filter(a => arr2.includes(a));
return matches.length > 0
```

callunity (approved):

avarsava (approved):

miso-web/src/main/webapp/scripts/bulk_sample.js
djcooke: Yes, that is better, thanks! I improved it further by returning `arr1.some(...)` instead of filtering and checking length

PR #2739, opened 2022-12-22
[GLT-3261] updated attachment categories to new bulk interface

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3261

- [x] Includes a change file


sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/RequisitionDao.java
```diff
@@ -11,6 +11,4 @@ public interface RequisitionDao extends PaginatedDataSource<Requisition>, SaveDa
 
   Requisition getByAlias(String alias) throws IOException;
 
-  List<Requisition> listByIdList(Collection<Long> ids) throws IOException;
```
avarsava, line 14: all the other files which from extending SaveDao to BulkSaveDao when removing this method, just confirming you don't want to do so here?

djcooke: Yeah, this was intentional. Requisitions don't have a bulk interface and the method was unused

avarsava (approved):

PR #2746, opened 2023-01-13
[GLT-3837] added VSCode settings

djcooke:
Java formatting should be very similar to what was set up in Eclipse. JavaScript formatting never worked well in Eclipse and formatting was pretty inconsistent, so I formatted all JS files with the new settings.

Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3837

- [ ] Includes a change file (n/a)


callunity (commented):
What do you think about adding a ` .prettierignore` file to exclude scripts in `miso-web/src/main/webapp/scripts/handsontable` ?

callunity (commented):
The `maven-minify-plugin` really doesn't like the prettier formatting, and it's what concatenates all our JS files into header-script.js . I don't recall what the actual space savings was of concatenating and minifying the files, but if we're going to remove the no-longer-supported maven-minify plugin, we'll need another mechanism for generating the header-script.js file

djcooke:
> The `maven-minify-plugin` really doesn't like the prettier formatting, and it's what concatenates all our JS files into header-script.js . I don't recall what the actual space savings was of concatenating and minifying the files, but if we're going to remove the no-longer-supported maven-minify plugin, we'll need another mechanism for generating the header-script.js file

I'm looking into this plugin: https://github.com/primefaces-extensions/resources-optimizer-maven-plugin

djcooke:
Had to remove `const` use and move a few block-scoped functions into variables to parse as ECMASCRIPT5. We should migrate to ECMASCRIPT_2015 or later, but that requires more fixes since it adds "use strict".

Bonus: the output file is being minified now, so there should be some minor speedup. A source map is generated, so we can still debug easily. It would be nice if it could map to the original files instead of the aggregate file, but I couldn't find a way to do that.

We were using the minify plugin for css too, but only one file was being included and it wasn't being minified, so it wasn't really doing anything.

callunity (approved):

PR #2756, opened 2023-02-22
[GLT-3513] Adding Service Record to Fridges/Freezers on MISO

hyeonjeongkwon:
Jira ticket or GitHub issue: GLT-3513

- [x] Includes a change file


miso-web/src/main/webapp/scripts/list_servicerecord.js
```diff
@@ -66,18 +78,28 @@ ListTarget.servicerecord = {
...
+                : "";
+            } else {
+              return data
+                ? '<a href="' +
+                    Urls.ui.freezers.editRecord(config.freezerId, full.id) +
+                    '">' +
+                    data +
+                    "</a>"
+                : "";
+            }
```
djcooke, line 97: Reduce duplication by setting url in a variable: `var url = config.instrumentId ? Urls.ui.instruments... : Urls.ui.freezers...`

miso-web/src/main/webapp/scripts/list_servicerecord.js
```diff
@@ -66,18 +78,28 @@ ListTarget.servicerecord = {
...
           }
           return data;
         },
       },
-      {
+      config.instrumentId && {
```
djcooke, line 102: This should already by excluded for freezers based on `include: config.hasPositions`

miso-web/src/main/webapp/scripts/list_storage_location.js
```diff
@@ -81,6 +81,9 @@ ListTarget.storage_location = {
         include: true,
         mRender: function (data, type, full) {
           if (type === "display") {
+            if (full.outOfService) {
```
djcooke, line 84: This is never set anywhere. Should be added to `StorageLocation` to determine whether out of service (similar to `InstrumentImpl`, but can ignore positions), then also added to `StorageLocationDto` and set in the relevant `Dtos` method so that it gets to the front end

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/StorageLocationRestController.java
```diff
@@ -306,10 +305,8 @@ public void bulkDelete(@RequestBody(required = true) List<Long> ids) throws IOEx
...
       @PathVariable(name = "locationId", required = true) Long locationId, @RequestBody ServiceRecordDto dto)
       throws IOException {
-    StorageLocation retrievedFreezer = RestUtils.retrieve("Storage location", locationId, storageLocationService);
-    if (retrievedFreezer == null) {
-      throw new NotFoundException("No freezer found for ID " + locationId.toString());
-    }
+    RestUtils.retrieve("Storage location", locationId, storageLocationService);
+
     ServiceRecord record = Dtos.to(dto);
     StorageLocation freezer = storageLocationService.get(locationId);
```
djcooke, line 311: Freezer is being fetched twice. Use `RestUtils#retrieve` only to get the freezer

miso-web/src/main/webapp/WEB-INF/pages/editFreezer.jsp
```diff
@@ -17,6 +17,10 @@
   <a id="addingStorageLink" target="_blank" rel="noopener noreferrer">Adding Storage to a Freezer</a>
 </div>
 
+<c:choose>
+      <c:when test="${freezer.isOutOfService()}"><p class="big-warning">Out of Service</p></c:when>
+    </c:choose>
```
djcooke, line 22: `choose` is meant for when there are multiple tests. More conventional to use `if` when there's only one

miso-web/src/main/webapp/scripts/list_storage_location.js
```diff
@@ -80,11 +80,12 @@ ListTarget.storage_location = {
         mData: "retired",
         include: true,
         mRender: function (data, type, full) {
-          if (type === "display") {
-            if (full.outOfService) {
-              return "Out of Service";
+          if (full.outOfService) {
+            return "Out of Service";
+          } else {
+            if (type === "display") {
```
djcooke, line 86: This `type` parameter is telling you why DataTables (table library we're using to create these lists) is calling the `mRender` function. `display` means to get the value to display in the table. The other options are related to sorting, filtering, and figuring out the datatype for the column.

Since you're checking for out of service before checking `type`, this will end up returning a string for out of service freezers, and a boolean for others, which may cause errors or strange behaviour. Since this will always display a string, I think it should be fine to use the same string for other purposes too. You can remove the `if (type === "display")` check and always return "Out of Service", "Retired", or "Production"

djcooke (approved):

miso-web/src/main/webapp/WEB-INF/pages/editInstrument.jsp
```diff
@@ -83,7 +83,7 @@
         <div id="records_arrowclick" class="toggleLeft"></div>
       </div>
       <div id="recordsdiv" class="expandable_section" style="display:none;">
-        <miso:list-section id="list_servicerecords" name="Service Records" target="servicerecord" alwaysShow="true" items="${serviceRecords}" config="{instrumentId: ${instrument.id}, retiredInstrument: ${instrument.dateDecommissioned != null && instrument.upgradedInstrument == null}, hasPositions: ${not empty instrument.instrumentModel.positions}, userIsAdmin: ${miso:isAdmin()}}"/>
+        <miso:list-section id="list_servicerecords" name="Service Records" target="servicerecord" alwaysShow="true" items="${serviceRecords}" config="{instrumentId: ${instrument.id}, retired: ${instrument.dateDecommissioned != null && instrument.upgradedInstrument == null}, hasPositions: ${not empty instrument.instrumentModel.positions}, userIsAdmin: ${miso:isAdmin()}}"/>
```
callunity, line 86: What prompted this change?

hyeonjeongkwon: There is a config called `retired` for freezer also. So on `list_servicerecord.js` I can use `config.retired` without the if case for checking whether it is a freezer or instrument

callunity: Thanks, I see that now!

callunity (approved):

PR #2760, opened 2023-02-27
Bug fix for file attachment on service records

hjkwon0:
Jira ticket or GitHub issue: GLT-3849

- [x] Includes a change file

### DIFF
--- a/miso-web/src/main/webapp/scripts/list_servicerecord.js
+++ b/miso-web/src/main/webapp/scripts/list_servicerecord.js
@@ -116,14 +116,13 @@ ListTarget.servicerecord = {
             });
             list += "</ul>";
             return list;
-          } else if (type === "filter") {
+          } else {
             return full.attachments
               .map(function (f) {
                 return f.filename;
               })
               .join(" ");
           }
-          return null;
         },
       },
       {


miso-web/src/main/webapp/scripts/list_servicerecord.js
```diff
@@ -114,7 +114,7 @@ ListTarget.servicerecord = {
             list += "</ul>";
             return list;
           }
-          return data;
+          return null;
```
callunity, line 117: Why `null` instead of something like a count of how many files there are?

hyeonjeongkwon: From my understanding, null will only be returned when there is no file attachment to the service records. If there is an attachment, the list of hyperlinks to each file attachment will be returned. My intention was to make columns with no file attachment to be empty.

hyeonjeongkwon: Will do!

miso-web/src/main/webapp/scripts/list_servicerecord.js
```diff
@@ -114,7 +115,17 @@ ListTarget.servicerecord = {
             list += "</ul>";
             return list;
           }
-          return data;
+          var filenames = "";
```
callunity, line 118: This could also be represented as 
```
full.attachments.map(f => f.filename).join(', ');
```
(@djcooke I think comma-separated might be slightly clearer in cases where the file names have spaces, but space-separated like you had recommended would be easily understood)

miso-web/src/main/webapp/scripts/list_servicerecord.js
```diff
@@ -114,7 +115,17 @@ ListTarget.servicerecord = {
...
+          var counts = 0;
+          full.attachments.forEach(function (file) {
+            if (full.attachments.length - 1 == counts) {
+              filenames += file.filename;
+            } else {
+              filenames += file.filename + " ";
+              ++counts;
+            }
+          });
+          return filenames;
```
djcooke, line 128: This will error if `file.attachments` is null/undefined. Could restructure function like:

```
if (!file.attachments || !file.attachments.length) {
  return null;
} else if (type === "display") {
  ...
} else {
  ...
}
```

miso-web/src/main/webapp/scripts/list_servicerecord.js
djcooke: This string is for filtering purposes, and is not displayed to the user. Commas should not be added, as they would alter what gets matched

callunity: Fair enough! Space-separated it is.

miso-web/src/main/webapp/scripts/list_servicerecord.js
hyeonjeongkwon: Could there be a case where file attachments is null? From my understanding, when there is no file attachments, `file.attachments` will be an empty array and `file.attachments.length` would be zero. Hence, it would simply return an empty string without iterating through each file attachment.

djcooke: This check makes it more resilient to back-end changes, so there is less chance of future bugs

hyeonjeongkwon: I see. Will make a change now!

miso-web/src/main/webapp/scripts/list_servicerecord.js
```diff
@@ -100,8 +100,11 @@ ListTarget.servicerecord = {
         mData: null,
         include: true,
         iSortPriority: 0,
+        bSortable: false,
         mRender: function (data, type, full) {
-          if (type === "display" && full.attachments && full.attachments.length) {
+          if (!full.attachments || !full.attachments.length) {
+            return null;
+          } else if (type === "display" && full.attachments && full.attachments.length) {
```
djcooke, line 107: This is duplicating the checks from above

miso-web/src/main/webapp/scripts/list_servicerecord.js
```diff
@@ -113,8 +116,19 @@ ListTarget.servicerecord = {
...
+            var counts = 0;
+            full.attachments.forEach(function (file) {
+              if (full.attachments.length - 1 == counts) {
+                filenames += file.filename;
+              } else {
+                filenames += file.filename + " ";
+                ++counts;
+              }
+            });
+            return filenames;
```
djcooke, line 130: @callunity made a suggestion to shorten this significantly using `map` and `join`

miso-web/src/main/webapp/scripts/list_servicerecord.js
```diff
@@ -113,8 +116,13 @@ ListTarget.servicerecord = {
             });
             list += "</ul>";
             return list;
+          } else {
+            return full.attachments
+              .map(function (f) {
+                f.filename;
```
djcooke, line 122: This needs to be a `return` statement

hyeonjeongkwon: @djcooke I am not sure what you mean. I thought this a `return` statement.

djcooke: A function written in this form does not return anything unless you explicitly use the `return` keyword

miso-web/src/main/webapp/scripts/list_servicerecord.js
```diff
@@ -113,8 +116,14 @@ ListTarget.servicerecord = {
             });
             list += "</ul>";
             return list;
+          } else if (type === "filter") {
```
hyeonjeongkwon, line 119: It turned out that mRender was called three times in some cases, 
one with type === "display", type === "filter", and type === "type"
we handled "display" one above, but did not for "filter" and "type"
so I added an extra if statement.

djcooke: Now you're returning a string for "filter" and `null` for "type." They should be the same. Was there a reason for this change?

hyeonjeongkwon: I could make it return the string for both cases. I thought it should only return string for filter and not for type.

djcooke (approved):

callunity (approved):

PR #2761, opened 2023-03-03
[GLT-3866] made assay test tissue type optional

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3866

- [x] Includes a change file

### DIFF
--- a/miso-web/src/main/webapp/scripts/list_assaytest.js
+++ b/miso-web/src/main/webapp/scripts/list_assaytest.js
@@ -53,12 +53,20 @@ ListTarget.assaytest = (function () {
           sTitle: "Tissue Type",
           mData: "tissueTypeId",
           mRender: function (data, type, full) {
-            var tissueType = Utils.array.findUniqueOrThrow(
-              Utils.array.idPredicate(data),
-              Constants.tissueTypes
-            );
-            var text = (full.negateTissueType ? "NOT " : "") + tissueType.alias;
-            return '<span title="' + tissueType.description + '">' + text + "</span>";
+            if (!data) {
+              return null;
+            } else {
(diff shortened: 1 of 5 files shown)


avarsava (approved):

miso-web/src/main/webapp/scripts/list_assaytest.js
```diff
@@ -53,12 +53,20 @@ ListTarget.assaytest = (function () {
...
+              return null;
+            } else {
+              var tissueType = Utils.array.findUniqueOrThrow(
+                Utils.array.idPredicate(data),
+                Constants.tissueTypes
+              );
+              var text = (full.negateTissueType ? "NOT " : "") + tissueType.alias;
+              if (type === "display") {
+                return '<span title="' + tissueType.description + '">' + text + "</span>";
+              } else {
```
hyeonjeongkwon, line 66: what is the else for? Can type be filter in this case?

hyeonjeongkwon (approved):

miso-web/src/main/webapp/scripts/list_assaytest.js
djcooke: `type` for `mRender` can always be "display", "sort", "filter", or "type". This is returning the same text for all, so sorting and filtering will be based on the same text that's being displayed, and the table will be able to figure out the data type (string). When `type === "display"`, there is HTML added.

We wouldn't want to include the HTML for the other `type`s, because it would cause strange behaviour with sorting and filtering

hyeonjeongkwon: Makes sense. Thanks!

PR #2764, opened 2023-03-07
[GLT-3863] added requisition stop reason field

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3863

- [x] Includes a change file

PR is ready for review. Marked as draft because a Pinery release and version bump is required before merging

### DIFF
--- a/pinery-miso/pom.xml
+++ b/pinery-miso/pom.xml
@@ -17,7 +17,7 @@
     <mysql.db>lims</mysql.db>
     <mysql.user>tgaclims</mysql.user>
     <mysql.pw>tgaclims</mysql.pw>
-    <pinery.version>2.25.0</pinery.version>
+    <pinery.version>2.25.1-SNAPSHOT</pinery.version>
     <spring-version>5.3.20</spring-version><!-- must match version used by Pinery -->
   </properties>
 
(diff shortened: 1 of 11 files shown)


callunity (approved):

pinery-miso/pom.xml
```diff
@@ -17,7 +17,7 @@
     <mysql.db>lims</mysql.db>
     <mysql.user>tgaclims</mysql.user>
     <mysql.pw>tgaclims</mysql.pw>
-    <pinery.version>2.25.0</pinery.version>
+    <pinery.version>2.25.1-SNAPSHOT</pinery.version>
```
hyeonjeongkwon, line 20: Could you explain what this would do?

hyeonjeongkwon (approved):

pinery-miso/pom.xml
djcooke: A snapshot is a version that's not yet released. This PR depends on updates to Pinery that haven't yet been released. I built the Pinery changes on my development machine so I could test this. Before merging this PR, there needs to be a release of Pinery, after which I'll update this to the proper release version

hyeonjeongkwon: Fair enough. Thanks!

PR #2777, opened 2023-05-23
#3922: Added troubleshooting section to the bottom of the MISO Bareme…

aryan2gandevia:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3922

 ✅ Includes a change file

### DIFF
--- a/docs/admin/baremetal-installation-guide.md
+++ b/docs/admin/baremetal-installation-guide.md
@@ -254,18 +254,18 @@ If you encounter other errors migrating the database, make sure that you are usi
 
 # Development Alternatives
 
- If using the correct MySQL version for your database server is causing trouble, consider downloading [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container:
+ If using the correct MySQL version for your database server is causing trouble, consider downloading [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container (note: things prefixed with `VAR_` are a variable, and their value will be specified below):
 
-    docker run --name CONTAINER_NAME -e MYSQL_ROOT_PASSWORD=ROOT_PASSWORD -e MYSQL_DATABASE=DB_NAME -e MYSQL_USER=DB_USERNAME -e MYSQL_PASSWORD=DB_PASSWORD -p 3306:3306 -d mysql:MYSQL_VERSION
+    docker run --name VAR_CONTAINER_NAME -e MYSQL_ROOT_PASSWORD=VAR_ROOT_PASSWORD -e MYSQL_DATABASE=VAR_DB_NAME -e MYSQL_USER=VAR_DB_USERNAME -e MYSQL_PASSWORD=VAR_DB_PASSWORD -p 3306:3306 -d mysql:VAR_MYSQL_VERSION
 
 
 Where:
-* `CONTAINER_NAME` is your desired Docker container name.
-* `ROOT_PASSWORD` is the root password to your MySQL.
-* `DB_NAME` is the name of the database (e.g. "lims").
-* `DB_USERNAME` is the username to access the database (e.g. "tgaclims").
-* `DB_PASSWORD` is the password to access the database (e.g. "tgaclims").
-* `MYSQL_VERSION` is the version of MySQL for the database (e.g. "5,7").
+* `VAR_CONTAINER_NAME` is your desired Docker container name.
+* `VAR_ROOT_PASSWORD` is the root password to your MySQL.
+* `VAR_DB_NAME` is the name of the database (e.g. "lims").
+* `VAR_DB_USERNAME` is the username to access the database (e.g. "tgaclims").
+* `VAR_DB_PASSWORD` is the password to access the database (e.g. "tgaclims").
+* `VAR_MYSQL_VERSION` is the version of MySQL for the database (e.g. "5,7").
 
 
 You do not need to do anything else from **Step 3** if you use this container method.
--- a/docs/admin/baremetal-installation-guide.md
+++ b/docs/admin/baremetal-installation-guide.md
@@ -254,7 +254,7 @@ If you encounter other errors migrating the database, make sure that you are usi
 
 # Development Alternatives
 
- If you can't or don't want to install the correct version of MySQL installed, this alternative allows the same result without you downloading it. To do this, download [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container:
+ If you can't or don't want to install the correct version of MySQL, this alternative allows the same result without you downloading it. To do this, download [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container:
 
     docker run --name $CONTAINER_NAME -e MYSQL_ROOT_PASSWORD=$ROOT_PASSWORD -e MYSQL_DATABASE=$DB_NAME -e MYSQL_USER=$DB_USERNAME -e MYSQL_PASSWORD=$DB_PASSWORD -p 3306:3306 -d mysql:8.0
 
--- a/docs/admin/baremetal-installation-guide.md
+++ b/docs/admin/baremetal-installation-guide.md
@@ -254,18 +254,18 @@ If you encounter other errors migrating the database, make sure that you are usi
 
 # Development Alternatives
 
- If using the correct MySQL version for your database server is causing trouble, consider downloading [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container (note: things prefixed with `VAR_` are a variable, and their value will be specified below):
+ If you are don't have the correct version of MySQL installed, this alternative allows the same result without you downloading it. To do this, download [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container:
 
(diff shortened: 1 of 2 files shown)


docs/admin/baremetal-installation-guide.md
```diff
@@ -240,3 +244,28 @@ If you run into an issue with migration `V0611`, ensure that the user running Fl
...
+* Whenever unexpected behaviour arises, be sure to check the build logs, located at `${CATALINA_BASE}/logs`. Read through the output of `catalina.out`, `catalina.<date>.log`, and `localhost.<date>.log`.
+* Ensure you are using the correct version of Java with Tomcat. This can be checked in `catalina.out`. In case you are not, append/modify the value of `$JAVA_HOME` in `setenv.sh` or `/etc/default/tomcat9`. 
+* If file access for `/storage/miso` is causing an error, [this](https://stackoverflow.com/questions/56827735/how-to-allow-tomcat-war-app-to-write-in-folder) may help.
+
+
+# Development Alternatives
+
+ If using the correct MySQL version for your database server is causing trouble, consider downloading [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container:
+
+    docker run --name CONTAINER_NAME -e MYSQL_ROOT_PASSWORD=ROOT_PASSWORD -e MYSQL_DATABASE=DB_NAME -e MYSQL_USER=DB_USERNAME -e MYSQL_PASSWORD=DB_PASSWORD -p 3306:3306 -d mysql:MYSQL_VERSION
```
callunity, line 259: It's not clear here which are variable names (e.g. `MYSQL_ROOT_PASSWORD`) and which should be variable values (e.g. `ROOT_PASSWORD`). Could you visually distinguish between them?

docs/admin/baremetal-installation-guide.md
```diff
@@ -240,3 +244,28 @@ If you run into an issue with migration `V0611`, ensure that the user running Fl
...
+
+
+Where:
+* `CONTAINER_NAME` is your desired Docker container name.
+* `ROOT_PASSWORD` is the root password to your MySQL.
+* `DB_NAME` is the name of the database (e.g. "lims").
+* `DB_USERNAME` is the username to access the database (e.g. "tgaclims").
+* `DB_PASSWORD` is the password to access the database (e.g. "tgaclims").
+* `MYSQL_VERSION` is the version of MySQL for the database (e.g. "5,7").
```
callunity, line 268: It might be worth moving this block above the docker command, for clarity.

callunity, line 268: For `MYSQL_VERSION` would the version be `5.7` (period instead of comma)?

docs/admin/baremetal-installation-guide.md
```diff
@@ -240,3 +244,28 @@ If you run into an issue with migration `V0611`, ensure that the user running Fl
...
+# Server Deployment Troubleshooting
+
+* Whenever unexpected behaviour arises, be sure to check the build logs, located at `${CATALINA_BASE}/logs`. Read through the output of `catalina.out`, `catalina.<date>.log`, and `localhost.<date>.log`.
+* Ensure you are using the correct version of Java with Tomcat. This can be checked in `catalina.out`. In case you are not, append/modify the value of `$JAVA_HOME` in `setenv.sh` or `/etc/default/tomcat9`. 
+* If file access for `/storage/miso` is causing an error, [this](https://stackoverflow.com/questions/56827735/how-to-allow-tomcat-war-app-to-write-in-folder) may help.
+
+
+# Development Alternatives
+
+ If using the correct MySQL version for your database server is causing trouble, consider downloading [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container (note: things prefixed with `VAR_` are a variable, and their value will be specified below):
```
djcooke, line 257: > If using the correct MySQL version for your database server is causing trouble

This wording sounds like you have the correct version installed and are having some problem with it. This is an alternative to installing the correct MySQL version, whether you're unable or just not wanting to install it.

> things prefixed with `VAR_` are a variable

The convention used in the rest of this document is to represent variables like `$CONTAINER_NAME`

docs/admin/baremetal-installation-guide.md
aryan2gandevia: I think below should be kept for consistency within the document, where variable values are given after they are referenced. Latest commit (commit 4) should give good clarity.

docs/admin/baremetal-installation-guide.md
```diff
@@ -240,3 +244,28 @@ If you run into an issue with migration `V0611`, ensure that the user running Fl
...
+Where:
+* `$CONTAINER_NAME` is your desired Docker container name.
+* `$ROOT_PASSWORD` is the root password to your MySQL.
+* `$DB_NAME` is the name of the database (e.g. "lims").
+* `$DB_USERNAME` is the username to access the database (e.g. "tgaclims").
+* `$DB_PASSWORD` is the password to access the database (e.g. "tgaclims").
+* `$MYSQL_VERSION` is the version of MySQL for the database (e.g. "5.7").
+
+
+You do not need to do anything else from **Step 3** if you use this container method.
```
callunity, line 271: "anything else" makes it sound like there's still something to be done in Step 3. Since it looks like this entirely replaces step 3, what do you think about rewording this to "Skip the database configuration (step 3) if you use this container method." ?

callunity (approved):

docs/admin/baremetal-installation-guide.md
```diff
@@ -240,3 +244,28 @@ If you run into an issue with migration `V0611`, ensure that the user running Fl
...
+# Server Deployment Troubleshooting
+
+* Whenever unexpected behaviour arises, be sure to check the build logs, located at `${CATALINA_BASE}/logs`. Read through the output of `catalina.out`, `catalina.<date>.log`, and `localhost.<date>.log`.
+* Ensure you are using the correct version of Java with Tomcat. This can be checked in `catalina.out`. In case you are not, append/modify the value of `$JAVA_HOME` in `setenv.sh` or `/etc/default/tomcat9`. 
+* If file access for `/storage/miso` is causing an error, [this](https://stackoverflow.com/questions/56827735/how-to-allow-tomcat-war-app-to-write-in-folder) may help.
+
+
+# Development Alternatives
+
+ If you are don't have the correct version of MySQL installed, this alternative allows the same result without you downloading it. To do this, download [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container:
```
djcooke, line 257:
```suggestion
 If you can't or don't want to install the correct version of MySQL, this alternative allows the same result. To do this, download [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container:
```

docs/admin/baremetal-installation-guide.md
```diff
@@ -240,3 +244,28 @@ If you run into an issue with migration `V0611`, ensure that the user running Fl
...
+    docker run --name $CONTAINER_NAME -e MYSQL_ROOT_PASSWORD=$ROOT_PASSWORD -e MYSQL_DATABASE=$DB_NAME -e MYSQL_USER=$DB_USERNAME -e MYSQL_PASSWORD=$DB_PASSWORD -p 3306:3306 -d mysql:$MYSQL_VERSION
+
+
+Where:
+* `$CONTAINER_NAME` is your desired Docker container name.
+* `$ROOT_PASSWORD` is the root password to your MySQL.
+* `$DB_NAME` is the name of the database (e.g. "lims").
+* `$DB_USERNAME` is the username to access the database (e.g. "tgaclims").
+* `$DB_PASSWORD` is the password to access the database (e.g. "tgaclims").
+* `$MYSQL_VERSION` is the version of MySQL for the database (e.g. "5.7").
```
djcooke, line 268: Put the version in the command instead of as a variable - you must use the version that MISO is compatible with. You can also specify 8.0 since this will be in the same release as that upgrade

docs/admin/baremetal-installation-guide.md
```diff
@@ -240,3 +244,28 @@ If you run into an issue with migration `V0611`, ensure that the user running Fl
...
+# Server Deployment Troubleshooting
+
+* Whenever unexpected behaviour arises, be sure to check the build logs, located at `${CATALINA_BASE}/logs`. Read through the output of `catalina.out`, `catalina.<date>.log`, and `localhost.<date>.log`.
+* Ensure you are using the correct version of Java with Tomcat. This can be checked in `catalina.out`. In case you are not, append/modify the value of `$JAVA_HOME` in `setenv.sh` or `/etc/default/tomcat9`. 
+* If file access for `/storage/miso` is causing an error, [this](https://stackoverflow.com/questions/56827735/how-to-allow-tomcat-war-app-to-write-in-folder) may help.
+
+
+# Development Alternatives
+
+ If you can't or don't want to install the correct version of MySQL installed, this alternative allows the same result without you downloading it. To do this, download [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container:
```
djcooke, line 257:
```suggestion
 If you can't or don't want to install the correct version of MySQL, this alternative allows the same result without you downloading it. To do this, download [Docker](https://docs.docker.com/get-docker/) to use a Docker container. Creating the container:
```

djcooke (approved):

PR #2779, opened 2023-05-25
#GLT-3917:  Fixed bug where all users were not given access to bulk a…

aryan2gandevia:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3917

✅ Includes a change file

### DIFF
--- a/miso-web/src/main/webapp/scripts/list_transferitem.js
+++ b/miso-web/src/main/webapp/scripts/list_transferitem.js
@@ -6,10 +6,6 @@ ListTarget.transferitem = (function () {
     },
     createBulkActions: function (config, projectId) {
       var actions = [];
-      // pushing a null into actions as people without admin access would otherside have no bulk actions to
-      // interact with table data. Other solutions require refactoring the way bulk actions are given for
-      // the following: samples, libraries, library aliquots and pools.
-      actions.push(null);
       if (config.editSend) {
         actions.push({
           name: "Remove",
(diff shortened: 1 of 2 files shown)


miso-web/src/main/webapp/scripts/list_transferitem.js
```diff
@@ -6,6 +6,10 @@ ListTarget.transferitem = (function () {
     },
     createBulkActions: function (config, projectId) {
       var actions = [];
+      // pushing a null into actions as people without admin access would otherside have no bulk actions to
+      // interact with table data. Other solutions require refactoring the way bulk actions are given for
+      // the following: samples, libraries, library aliquots and pools.
+      actions.push(null);
```
avarsava, line 12: What is the mechanism by which this fixes the issue?

aryan2gandevia: It makes the `actions` variable non-empty for all users, which then renders in the regular bulk actions. When `actions` was empty for users not part of a special group, the table would not load regular bulk actions that are needed to interact with table data.

djcooke: Please clarify the comment - sample/library/aliquot/pool actions are added later, after the table's main actions have already been processed. If the main bulk actions are empty, bulk controls (checkboxes, select all/none) are not added. This null causes the bulk actions to be non-empty and is rendered as a separator

The null should only be added if there are no other actions

djcooke (approved):

avarsava (approved):

PR #2784, opened 2023-05-30
[GLT-3911] Renamed "short name" and "alias" fields in Project.

aryan2gandevia:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3911

✅ Includes a change file

PR #2785, opened 2023-05-31
[GLT-3910] added project advanced search

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3910

- [x] Includes a change file

This is going to conflict hard with #2784. I'll rebase and fix up this PR after that gets merged


djcooke:
@callunity No plans to implement for other things yet, but I intentionally didn't make it specific to projects so that we could later if wanted. Run status is already filterable using "is:" but I didn't think was a great pattern to build on

aryan2gandevia (approved):
looks good!

Issue #2792, opened 2023-06-20
Miso irregular sorting order 


jayramr:
Hello, 

I would like to raise below two requests with you if you can include/integrate in the next release. I tested and verified the release notes till v1.56.0 and the below two feature is not yet added. Kindly let us know.

1) Enable description column in the Sample summary page. 
Currently to check the description for each sample, we selecting the specific sample and then the description field is visible. If it is available to see in the Sample summary page would be great. Kindly refer the screenshot.

<img width="1490" alt="Sample_description" src="https://github.com/miso-lims/miso-lims/assets/97015642/9d08da7b-36f3-4f34-9300-f658ded4cb12">

2) Incorrect sorting order for Sample names while creating libraries from samples 
When creating libraries from the samples, at this page mentioned by below screenshot, the sample names are initially comes with irregular order, currently we are manually click on "Sort -> Primary Sort(Sample Name) and then Click on Sort", this helps the Sample name comes in ascending order. But the problem is that multiple users are accessing our miso system, some people forget this sorting check. 

Is it possible could you enable the sorting enabled by-default for Sample Name column as ascending order? instead of showing the Sample name column as irregular order? 

<img width="1080" alt="sorting_v2" src="https://github.com/miso-lims/miso-lims/assets/97015642/411c6cca-5997-41ea-a51c-cb3fea1fb2af">

In addition, may I know when we can expect the Miso v2.0.0 official docker release? 

Kindly advise on this. 

Thanks
Jay

djcooke:
Hello,

1. We will not add the description column to the list page, as it is a long text field and can take up too much space.
2. I'm not sure how the sorting works on this page. I will investigate further before commenting.
3. MISO 2.0.0 has a critical bug, and will not have an official release. We are currently testing v2.1.0, and its offical release is currently planned for June 29

jayramr:
Thanks @djcooke for the update for point 1 and 3. This is well noted. 

Regarding the sorting order, this applicable for Projects, Samples, libraries as well. Please investigate and update us. 

Thanks
Jay

djcooke:
There will be a fix for your second point in the next release (after 2.1.0):

* When propagating libraries from the Edit Box page, rows will be sorted by sample box column
* When propagating samples, libraries (with the above exception), or library aliquots, rows will be sorted by parent ID
* When editing items, rows will be sorted by ID

Projects cannot be edited in bulk, so I'm not sure what you mean when you say they are affected?

[...]

PR #2793, opened 2023-06-20
[#2792] fixed default sorting on bulk pages

djcooke:
Jira ticket or GitHub issue: https://github.com/miso-lims/miso-lims/issues/2792

- [x] Includes a change file


djcooke:
Need a different approach - this broke later sorting

PR #2797, opened 2023-06-27
[GLT-3939] Added "details" field to Projects.

aryan2gandevia:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3939

✅ Includes a change file

### DIFF
--- a/changes/add_GLT-3939.md
+++ b/changes/add_GLT-3939.md
@@ -1,0 +1,1 @@
+Details field on Project form
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/ProjectImpl.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/ProjectImpl.java
@@ -64,6 +64,7 @@ public class ProjectImpl implements Project {
   private String title = "";
   private String code;
   private String rebNumber;
+  private String additionalDetails = "";
 
   private LocalDate rebExpiry;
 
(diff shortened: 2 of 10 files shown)


changes/add_GLT-3939.md
```diff
@@ -0,0 +1 @@
+Details field on Project form
```
djcooke, line 1:
```suggestion
Additional Details field on Edit Project page
```

avarsava:
Why is it called 'additional details' everywhere except the front end?

aryan2gandevia:
> Why is it called 'additional details' everywhere except the front end?

Dillan and I decided to call the field "details" on the frontend as opposed to the ticket calling the field "additional details". Should I change the method names to match this?

djcooke:
> > Why is it called 'additional details' everywhere except the front end?
> 
> Dillan and I decided to call the field "details" on the frontend as opposed to the ticket calling the field "additional details". Should I change the method names to match this?

It should be called the same thing everywhere either way. Looking at it now with its placement under Description, "Additional Details" is probably better

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/ProjectImpl.java
```diff
@@ -64,6 +64,7 @@ public class ProjectImpl implements Project {
   private String title = "";
   private String code;
   private String rebNumber;
+  private String additionalDetails = "";
```
djcooke, line 67: Don't initialize to empty String. I don't know why some of the other fields do this, but there might be some old code that depends on it for those

miso-web/src/main/webapp/scripts/form_project.js
```diff
@@ -111,6 +111,13 @@ FormTarget.project = (function ($) {
               type: "text",
               maxLength: 255,
             },
+            {
+              title: "Additional Details",
+              data: "additionalDetails",
+              type: "text",
```
djcooke, line 117: Use type "textarea" for a larger text field

djcooke (approved):
Approved after form field type is corrected

avarsava (approved):

PR #2801, opened 2023-07-14
[GLT-3941] modify projects contact field

aryan2gandevia:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-3941

✅ Includes a change file


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/ProjectImpl.java
```diff
@@ -77,9 +76,10 @@ public class ProjectImpl implements Project {
   private Deliverable deliverable;
 
 
-  @OneToMany(targetEntity = ProjectContactsAndRole.class, mappedBy = "project", orphanRemoval = true,
+  @OneToMany(targetEntity = ProjectContact.class, mappedBy = "project", orphanRemoval = true,
       cascade = CascadeType.ALL)
-  private Set<ProjectContactsAndRole> contacts = new HashSet<>();
+  private List<ProjectContact> contacts = new ArrayList<>();
+  // private Set<ProjectContact> contacts = new HashSet<>();
```
djcooke, line 82: Remove commented-out code

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/ContactDto.java
```diff
@@ -1,13 +1,11 @@
...
 
-import uk.ac.bbsrc.tgac.miso.core.data.impl.ContactRole;
-
 public class ContactDto {
 
   private Long id;
   private String name;
   private String email;
-  private ContactRole contactRole;
+  private ContactRoleDto contactRoleDto;
```
djcooke, line 8: Field name shouldn't contain "Dto"

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultContactRoleService.java
```diff
@@ -86,7 +86,7 @@ public ValidationResult validateDeletion(ContactRole object) throws IOException
     ValidationResult result = new ValidationResult();
     long usage = contactRoleDao.getUsage(object);
     if (usage > 0L) {
-      result.addError(ValidationError.forDeletionUsage(object, usage, Pluralizer.projects(usage)));
+      result.addError(ValidationError.forDeletionUsage(object, usage, Pluralizer.contacts(usage)));
```
djcooke, line 89: Change to `"project" + Pluralizer.contacts(usage)` - this is specifically a count of project-contacts. The same contact will be counted multiple times if associated with the same role in multiple projects

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultProjectService.java
```diff
@@ -182,6 +182,15 @@ private void validateChange(Project project, Project beforeChange) throws IOExce
       errors.add(new ValidationError("title", "There is already a project with this title"));
     }
 
+    List<Long> uniqueContacts = new ArrayList<>();
+    for (ProjectContact element : project.getContacts()) {
+      if (uniqueContacts.contains(element.getContact().getId())) {
+        errors.add(new ValidationError("contacts", "You cannot use the same contact twice"));
+        break;
```
djcooke, line 189: `continue` instead of `break` so that all duplicates will be reported instead of only the first

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultProjectService.java
```diff
@@ -229,14 +240,29 @@ private void applyChanges(Project original, Project project) {
...
+        if (to.get(i).getContact().getId() == from.get(j).getContact().getId()) {
+          to.get(i).setContactRole(from.get(j).getContactRole());
+          remover = false;
+          break;
+        }
+      }
+      if (remover) {
+        to.remove(i);
+      }
+    }
```
djcooke, line 256: This can all be replaced by checking the contact role ID in addition to the contact ID in the `removeIf...noneMatch...` below

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultProjectService.java
```diff
@@ -194,12 +203,14 @@ private void loadChildEntities(Project project) throws IOException {
...
+    Set<ProjectContact> contacts = new HashSet<>();
+    for (ProjectContact element : project.getContacts()) {
       Contact contact = contactService.get(element.getContact().getId());
       ContactRole contactRole = contactRoleService.get(element.getContactRole().getId());
-      contacts
-          .add(new ProjectContactsAndRole(project, contact, contactRole));
+      // check to ensure the values load properly, and if they do, add it to the set
+      if (contact != null && contactRole != null) {
+        contacts.add(element);
+      }
```
djcooke: * This will ignore invalid contact or role IDs. Use `loadChildEntity`, which will handle this properly
* You're removing all of the `ProjectContacts`s from the project, and then re-adding the exact same objects. There's no need to create a new collection or add/remove items from the existing one

miso-web/src/main/webapp/WEB-INF/pages/editProject.jsp
```diff
@@ -50,6 +50,24 @@
...
+</div>
+
+<div class="sectionDivider" onclick="Utils.ui.toggleLeftInfo(jQuery('#assays_section_arrowclick'), 'assays_section');">
+  Assays
+<div id="assays_section_arrowclick" class="toggleLeftDown"></div>
+</div>
+<div id="projectForm_assaysError"></div>
+<div id="assays_section" class="expandable_section">
+  <h1>Assays</h1>
+</div>
```
djcooke, line 69: Make these non-collapsible

miso-web/src/main/webapp/scripts/bulk_contactrole.js
```diff
@@ -14,8 +14,8 @@ BulkTarget.contactRole = (function () {
     getSaveProgressUrl: function (operationId) {
       return Urls.rest.contactRoles.bulkSaveProgress(operationId);
     },
-    getUserManualUrls: function () {
-      return Urls.external.userManual("type_data", "contactRoles");
+    getUserManualUrl: function () {
+      return Urls.external.userManual("type_data", "contactroles");
```
djcooke, line 18: "contact-roles" needs a hyphen to create a working link

miso-web/src/main/webapp/scripts/list_contactrole.js
```diff
@@ -1,17 +1,17 @@
-ListTarget.contactRole = (function ($) {
+ListTarget.contactrole = (function ($) {
   return {
     name: "Contact Roles",
     getUserManualUrl: function () {
-      return Urls.external.userManual("type_data", "contactRoles");
+      return Urls.external.userManual("type_data", "contactroles");
```
djcooke, line 5: Need a hyphen

callunity (approved):

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultProjectService.java
aryan2gandevia: Since there isn't a setter method for contact and contact roles for project (to put as a parameter for `loadChildEntities`), I decided to just follow a similar method to how assay tests and metrics were handled for validation.

djcooke: The setter doesn't have to be on the project:

```
for (ProjectContact projectContact : project.getContacts()) {
  loadChildEntity(projectContact::setContact, projectContact.getContact(), contactService, "contact");
  loadChildEntity(projectContact::setContactRole, projectContact.getContactRole(), contactRoleService, "contact");
}
```

Note "contact" is used as property name for both so that the error messages land in the same `projectForm_contactErrors` div

miso-web/src/main/webapp/WEB-INF/pages/editProject.jsp
```diff
@@ -50,19 +50,12 @@
...
+<div id="contacts_section">
   <h1>Contacts</h1>
 </div>
 
-<div class="sectionDivider" onclick="Utils.ui.toggleLeftInfo(jQuery('#assays_section_arrowclick'), 'assays_section');">
-  Assays
-<div id="assays_section_arrowclick" class="toggleLeftDown"></div>
-</div>
 <div id="projectForm_assaysError"></div>
 <div id="assays_section" class="expandable_section">
```
djcooke, line 60: Remove class

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultProjectService.java
```diff
@@ -182,13 +181,16 @@ private void validateChange(Project project, Project beforeChange) throws IOExce
...
-    for (ProjectContact element : project.getContacts()) {
-      if (uniqueContacts.contains(element.getContact().getId())) {
-        errors.add(new ValidationError("contacts",
-            "The contact " + element.getContact().getEmail() + " has been used more than once"));
+    List<ProjectContact> contacts = project.getContacts();
+    for (int i = 0; i < contacts.size(); i++) {
+      for (int j = i + 1; j < contacts.size(); j++) {
+        if (contacts.get(i).getContact().getId() == contacts.get(j).getContact().getId()
+            && contacts.get(i).getContactRole().getId() == contacts.get(j).getContactRole().getId()) {
+          errors.add(new ValidationError("contacts", "The contact " + contacts.get(i).getContact().getEmail()
```
djcooke, line 189: Field name should be "contact" to match error message div

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultProjectService.java
```diff
@@ -250,7 +244,8 @@ public void applySetChangesContacts(List<ProjectContact> to, List<ProjectContact
         toItem -> from.stream().noneMatch(fromItem -> fromItem.getContact().getId() == toItem.getContact().getId()
             && fromItem.getContactRole().getId() == toItem.getContactRole().getId()));
     from.forEach(fromItem -> {
-      if (to.stream().noneMatch(toItem -> toItem.getContact().getId() == fromItem.getContact().getId())) {
+      if (to.stream().noneMatch(toItem -> toItem.getContact().getId() == fromItem.getContact().getId()
+          && fromItem.getContactRole().getId() == toItem.getContactRole().getId())) {
```
djcooke, line 248: Can move the contactId+contactRoleId checking to a `private static boolean contactsMatch(ProjectContact a, ProjectContact b)` method to use here and the other 2 places you're checking the same (removeIf just above and in validation)

pinery-miso/src/main/resources/ca/on/oicr/pinery/lims/miso/queryAllSampleProjects.sql
```diff
@@ -42,4 +43,10 @@ LEFT JOIN (
   ) items
   GROUP BY project_projectId
 ) stats ON stats.project_projectId = proj.projectId
-LEFT JOIN Contact c ON c.contactId = proj.contactId
+LEFT JOIN Project_Contact pc ON pc.projectId = (
+  SELECT 
+    pc1.projectId
+  FROM Project_Contact pc1 WHERE proj.projectId = pc1.projectId AND pc1.contactRoleId = 1
+  ORDER BY pc1.projectId LIMIT 1
+)
```
djcooke, line 51: `LIMIT` applies to the entire query, which means this will only select 1 contact - not 1 for each project as you intended

miso-web/src/main/webapp/scripts/project_ajax.js
```diff
@@ -47,14 +53,28 @@ var Project = (function () {
 
     addContact: function (addContact) {
       var contacts = Project.getContacts();
-      contacts.push(addContact);
-      Project.setContacts(contacts);
+      addContact["projectId"] = projectId;
+      if (
+        contacts.find(function (contact) {
```
aryan2gandevia, line 58: Added this check because `setTableData`'s fake IDs would have weird behaviour in the following instance:

A duplicate contact+role is added, then the user attempts to save, resulting in the error message describing that there is a duplicate contact. When you try to select any item in the list after, even if you check the box, it does not select anything.

By adding this, it prevents duplication, and thus mitigates the potential of this error. The duplication check in `DefaultProjectService` should still be kept in the odd chance that somehow this check fails, as we don't want to cause an error in the database and corrupt data.

miso-web/src/main/webapp/scripts/list_contact.js
```diff
@@ -80,8 +95,15 @@ ListTarget.contact = (function ($) {
         return {
           name: contactRole.name,
           handler: function () {
-            contact.contactRole = contactRole;
-            Project.addContact(contact);
+            // setting contact variable in ProjectContactDto format
```
aryan2gandevia, line 98: Done this as we can use the existing add contact function made, and just transfer its information into a new variable with some extra fields (from adding the contact role).

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/ProjectContactDto.java
```diff
@@ -0,0 +1,60 @@
+package uk.ac.bbsrc.tgac.miso.dto;
+
+public class ProjectContactDto {
+  private long projectId;
+  private long contactId;
+  private long contactRoleId;
+  // fields required for the case where a new contact is made
```
aryan2gandevia, line 7: When a new contact is made, these fields are needed to be stored in order to make the new contact through `saveNewContact` function in `DefaultProjectService`. Before, when using a contactDto, we already had the fields, so this is just to mimic that.

miso-web/src/main/webapp/scripts/form.js
```diff
@@ -479,7 +486,7 @@ FormUtils = (function ($) {
           if (item.id < 0) {
             item.id = null;
           }
```
djcooke, line 488: This bit needs to be removed for this fix to work. The temporary ID should not be removed

miso-web/src/main/webapp/scripts/project_ajax.js
```diff
@@ -35,5 +42,47 @@ var Project = (function () {
...
+    addContact: function (addContact) {
+      var contacts = Project.getContacts();
+      addContact["projectId"] = projectId;
+      if (
+        contacts.find(function (contact) {
+          return (
+            contact.contactId === addContact.contactId &&
+            contact.contactRoleId === addContact.contactRoleId
+          );
+        })
```
djcooke, line 63: This doesn't handle new contacts - if two are added with the same role, it'll call them duplicates even though the name and email are different. Change to use the same checks as in `removeContacts` (and move that to a function so it's not duplicated)

djcooke (approved):

PR #2957, opened 2024-08-02
[GLT-4230] Make Target Cell Recovery an Integer

sinry-d:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4230

- [ ] Includes a change file

### DIFF
--- a/miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
+++ b/miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
@@ -870,7 +870,7 @@ public class Dtos {
     if (from instanceof SampleStockSingleCellRelative) {
       SampleStockSingleCellRelative scFrom = (SampleStockSingleCellRelative) from;
       SampleStockSingleCell sc = new SampleStockSingleCellImpl();
-      setBigDecimal(sc::setTargetCellRecovery, scFrom.getTargetCellRecovery());
+      setInteger(sc::setTargetCellRecovery, scFrom.getTargetCellRecovery());
       setBigDecimal(sc::setCellViability, scFrom.getCellViability());
       setBigDecimal(sc::setLoadingCellConcentration, scFrom.getLoadingCellConcentration());
       to = sc;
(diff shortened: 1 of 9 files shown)


miso-web/src/it/java/uk/ac/bbsrc/tgac/miso/webapp/integrationtest/BulkSampleCreateIT.java
```diff
@@ -804,7 +804,7 @@ public void testCreateOneSingleCellStock() throws Exception {
     stock.put(SamColumns.QC_STATUS, "Ready");
     stock.put(SamColumns.INITIAL_CELL_CONC, "12.34");
     stock.put(SamColumns.DIGESTION, "abcde");
-    stock.put(SamColumns.TARGET_CELL_RECOVERERY, "23.45");
+    stock.put(SamColumns.TARGET_CELL_RECOVERERY, "23");
```
avarsava, line 807: Recoverery?

avarsava (approved):

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -4600,6 +4600,10 @@ public static void setBoolean(@Nonnull Consumer<Boolean> setter, Boolean value,
     }
   }
 
+  public static void setInteger(@Nonnull Consumer<Integer> setter, String value) {
+    setter.accept(isStringEmptyOrNull(value) ? null : Integer.valueOf(value));
+  }
```
djcooke, line 4605: There's already a `setInteger` method that can be used; it has an extra parameter to specify whether null is allowed (should be `true` in this case)

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -4608,6 +4612,10 @@ public static void setInteger(@Nonnull Consumer<Integer> setter, Integer value,
     }
   }
 
+  public static void setString(@Nonnull Consumer<String> setter, Integer value) {
+    setter.accept(value == null ? null : value.toString());
+  }
```
djcooke, line 4617: This shouldn't be necessary. The `BigDecimal` was expressed as a `String` in the DTO in order to avoid floating point precision issues. Since that's not a concern for integers, the DTO field should be changed to `Integer` to match

miso-web/src/main/webapp/scripts/bulk_sample.js
```diff
@@ -1154,7 +1154,7 @@ BulkTarget.sample = (function ($) {
       columns.push(
         {
           title: "Target Cell Recovery",
-          type: "decimal",
+          type: "int",
```
djcooke, line 1157: `precision` and `scale` attributes should be removed since they are only for decimals. Can set `min` to 0 as well. Same in `form_sample.js`

djcooke (commented):
See comments

djcooke (approved):

PR #2968, opened 2024-08-23
[GLT-4258] fixed SOP URL validation

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4258

- [x] Includes a change file

### DIFF
--- a/miso-web/src/main/webapp/scripts/bulk.js
+++ b/miso-web/src/main/webapp/scripts/bulk.js
@@ -3362,23 +3362,24 @@ BulkUtils = (function ($) {
 
   function textValidator(column) {
     var sanitize = new RegExp(Utils.validation.sanitizeRegex);
-    var regex = column.regex ? new RegExp(column.regex) : null;
     return function (value, callback) {
       if (Utils.validation.isEmpty(value)) {
         return validateEmpty(this, callback);
       } else if (column.maxLength && value.length > column.maxLength) {
         callback(false);
         return "Max length is " + column.maxLength + " characters";
+      } else if (column.regex) {
+        var regex = new RegExp(column.regex);
+        if (!regex.test(value)) {
+          callback(false);
+          return "Does not match expected pattern";
+        }
       } else if (!sanitize.test(value)) {
         callback(false);
         return "Cannot contain tabs, line breaks, or the following characters: <>&";
-      } else if (regex && !regex.test(value)) {
-        callback(false);
-        return "Does not match expected pattern";
-      } else {
-        callback(true);
-        return null;
       }
+      callback(true);
+      return null;
     };
   }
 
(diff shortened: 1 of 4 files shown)


miso-web/src/main/webapp/scripts/lims.js
```diff
@@ -1280,7 +1280,8 @@ Utils.validation = {
   dateTimeRegex:
     "^(19|20)[0-9]{2}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01]) ([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]$",
   sanitizeRegex: "^[^<>&\\t\\n\\r]*$",
-  uriComponentRegex: "^[^<>&%;/\\\\]*$", // problem characters even when encoded in URI components: %;\/
+  uriComponentRegex: "^[^<>&%;/\\\\]*$", // problem characters even when encoded in URI components: <>&%;\/
+  uriRegex: "^[^<>\\\\]*$", // problem characters in full URIs: <>\
```
grapigeau, line 1284: Are new lines and tabs ok in a full uri?

djcooke: A lot more characters are actually bad, but any modern browser will encode them automatically, so it doesn't really matter

grapigeau (approved):

miso-web/src/main/webapp/scripts/lims.js
avarsava: It also looks like a separate sanitize method at bulk.js:3377 rejects those characters

miso-web/src/main/webapp/scripts/bulk.js
```diff
@@ -3362,23 +3362,24 @@ BulkUtils = (function ($) {
...
   function textValidator(column) {
     var sanitize = new RegExp(Utils.validation.sanitizeRegex);
-    var regex = column.regex ? new RegExp(column.regex) : null;
     return function (value, callback) {
       if (Utils.validation.isEmpty(value)) {
         return validateEmpty(this, callback);
       } else if (column.maxLength && value.length > column.maxLength) {
         callback(false);
         return "Max length is " + column.maxLength + " characters";
+      } else if (column.regex) {
```
avarsava, line 3371: If column.regex exists, then 'sanitize' is never tested?

djcooke: Sanitize is the default and prevents characters that we don't want to allow in most fields. Specifying a regex overrides that. Among other things, this can be used to allow characters that would normally be restricted by sanitize, as in this case. The form pages already work like this, but I guess we never had to allow characters that were prevented by sanitize on the bulk pages before.

miso-web/src/main/webapp/scripts/lims.js
avarsava: I don't think it's safe to make assumptions about what the client will be using. We should exclude at least \n, \t, \r, and \0

djcooke: We're assuming either way. Preventing characters that must be encoded assumes that the user isn't copying from somewhere that displays URLs unencoded, which is very common since it is widely supported.

I've disallowed tabs and newlines again since they don't really work either way, and it's not a change from the previous behaviour. Further improvements could be made, but the point of this PR is just to allow the ampersand character. If you see value in further improvements, please make a separate ticket.

PR #2970, opened 2024-09-05
[GLT-4262] Fixed tests

djcooke:
Jira ticket or GitHub issue:

- [ ] Includes a change file

Not ready for review. Making PR to run tests


djcooke:
It looks like the wind-size parameter value should have a comma rather than a "x" though it was working before... Changed to a more reliable method of setting window size anyway

avarsava (approved):

grapigeau (approved):

PR #3001, opened 2025-01-03
[GLT-4235] increase sop alias max length

wuall826:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4235

- [ ] Includes a change file


sqlstore/src/main/resources/db/migration/V8510__sop_alias_length.sql
grapigeau, line 1: This is an unreleased migration? (https://github.com/miso-lims/miso-lims/blob/92ac8a1c85f2be958517dc6d6fb8d77ab5242d55/sqlstore/src/main/resources/db/migration/readme.md)

grapigeau (approved):

PR #3002, opened 2025-01-15
[GLT-4234] show tissue processing volume and concentration

wuall826:
Jira ticket or GitHub issue:

- [x] Includes a change file

### DIFF
--- a/miso-web/src/main/webapp/scripts/bulk_sample.js
+++ b/miso-web/src/main/webapp/scripts/bulk_sample.js
@@ -1141,19 +1141,16 @@ BulkTarget.sample = (function ($) {
       );
 
       if (
-        (!Constants.isDetailedSample || show["Stock"] || show["Aliquot"]) &&
+        (!Constants.isDetailedSample ||
+          show["Stock"] ||
+          show["Aliquot"] ||
+          show["Tissue Processing"]) &&
         !config.isLibraryReceipt
       ) {
         columns = columns.concat(BulkUtils.columns.volume(true, config));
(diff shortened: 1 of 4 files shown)


grapigeau (approved):

miso-web/src/main/webapp/scripts/bulk_sample.js
```diff
@@ -1141,11 +1141,14 @@ BulkTarget.sample = (function ($) {
...
-        (!Constants.isDetailedSample || show["Stock"] || show["Aliquot"]) &&
+        (!Constants.isDetailedSample ||
+          show["Stock"] ||
+          show["Aliquot"] ||
+          show["Tissue Processing"]) &&
         !config.isLibraryReceipt
       ) {
         columns = columns.concat(BulkUtils.columns.volume(true, config));
-        if (Constants.isDetailedSample) {
+        if (Constants.isDetailedSample && !show["Tissue Processing"]) {
```
djcooke, line 1151: The `show[...]` values are true if the specified category is the target category (main sample being entered) OR if it's an upstream category that needs to be created as part of the hierarchy. e.g. `show["Tissue"]` is true when receiving stocks because we need to enter tissue information at the same time.

Here, you should check `targetCategory` instead because you only want to avoid showing `parentUsed` if the target category is "Tissue Processing."

djcooke (commented):
There are some failing tests. These are checking the setup of the bulk tables, so you have to add in the new columns that are expected

wuall826:
ignore the commit, still making changes

djcooke (approved):

PR #3004, opened 2025-02-03
[GLT-4231] added new fields for single cell

wuall826:
Jira ticket or GitHub issue:

- [ ] Includes a change file


wuall826:
fixing test

miso-web/src/main/webapp/scripts/bulk_sample.js
```diff
@@ -1095,6 +1095,27 @@ BulkTarget.sample = (function ($) {
...
           precision: 14,
           scale: 10,
+          description: "Initial concentration of cells in the sample at the time of receipt",
+        },
+        {
+          title: "Target Cell Recovery",
+          type: "int",
+          data: "targetCellRecovery",
+          min: 0,
+          include: show["Tissue Processing"] && !show["Stock"],
```
djcooke, line 1105: This will have the following effects:

- When creating tissue processing, it will show for the tissue processing level
- When creating stock, it will show for the stock level
- When creating aliquots, it will show for both the tissue processing and stock levels

We don't want it showing up twice, and if we're creating aliquots, it should probably be the lower-level stock field that we show. Can be fixed with:

```
include: targetCategory === "Tissue Processing"
```

djcooke (approved):

grapigeau (approved):

PR #3008, opened 2025-02-21
[GLT-4229] add prepared samples to edit requisition

wuall826:
Jira ticket or GitHub issue:

- [x] Includes a change file

### DIFF
--- a/changes/add_GLT-4229.md
+++ b/changes/add_GLT-4229.md
@@ -1,0 +1,1 @@
+Prepared Samples table on Edit Requisition Page to show all downstream samples
--- a/changes/add_GLT-4229.md
+++ b/changes/add_GLT-4229.md
@@ -1,1 +1,0 @@
-Prepared Samples table on Edit Requisition Page to show all downstream samples
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleRestController.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleRestController.java
@@ -210,33 +210,17 @@ public class SampleRestController extends RestController {
     List<Sample> requisitionedSamples = sampleService.list(0, 0, false, "id", PaginationFilter.requisitionId(id));
     List<Sample> supplementalSamples =
         sampleService.list(0, 0, false, "id", PaginationFilter.supplementalToRequisitionId(id));
-    Set<Long> rootSampleIds = Stream.concat(requisitionedSamples.stream(), supplementalSamples.stream())
+    Set<Long> sampleIds = Stream.concat(requisitionedSamples.stream(), supplementalSamples.stream())
         .map(Sample::getId)
         .collect(Collectors.toSet());
-    Set<Long> allDerivativeSampleIds = new HashSet<>();
-    List<Sample> tissueSamples = sampleService.getChildren(rootSampleIds, SampleTissue.CATEGORY_NAME, id);
-    Set<Long> tissueSampleIds = tissueSamples.stream()
-        .map(Sample::getId)
-        .collect(Collectors.toSet());
-    allDerivativeSampleIds.addAll(tissueSampleIds);
-    List<Sample> tissueProcessingSamples = sampleService.getChildren(
-        tissueSampleIds, SampleTissueProcessing.CATEGORY_NAME, id);
-    Set<Long> tissueProcessingIds = tissueProcessingSamples.stream()
-        .map(Sample::getId)
-        .collect(Collectors.toSet());
-    allDerivativeSampleIds.addAll(tissueProcessingIds);
-    List<Sample> stockSamples = sampleService.getChildren(rootSampleIds, SampleStock.CATEGORY_NAME, id);
-    Set<Long> stockSampleIds = stockSamples.stream()
-        .map(Sample::getId)
-        .collect(Collectors.toSet());
-    allDerivativeSampleIds.addAll(stockSampleIds);
-    List<Sample> aliquotSamples = sampleService.getChildren(stockSampleIds, SampleAliquot.CATEGORY_NAME, id);
-    Set<Long> aliquotSampleIds = aliquotSamples.stream()
-        .map(Sample::getId)
-        .collect(Collectors.toSet());
-    allDerivativeSampleIds.addAll(aliquotSampleIds);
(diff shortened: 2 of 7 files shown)


grapigeau (approved):

miso-web/src/main/webapp/scripts/list_samples.js
```diff
@@ -16,6 +16,8 @@ ListTarget.sample = (function () {
       } else if (config.requisitionId) {
         if (config.supplemental) {
           return Urls.rest.samples.requisitionSupplementalDatatable(config.requisitionId);
+        } else if (config.filter === "samples-prepared") {
+          return Urls.rest.samples.samplesPreparedDatatable(config.requisitionId);
         } else {
           return Urls.rest.samples.requisitionDatatable(config.requisitionId);
         }
```
djcooke, line 23: This is now using 2 parameters to control 1 thing. Switch to work the same as the library table - `config.relation` with value "requisitioned", "supplemental", or "indirect"

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleRestController.java
```diff
@@ -203,6 +203,32 @@ public DataTablesResponseDto<SampleDto> getDTArrayedSamplesByProject(@PathVariab
...
+    List<Sample> supplementalSamples =
+        sampleService.list(0, 0, false, "id", PaginationFilter.supplementalToRequisitionId(id));
+
+    Set<Long> sampleIds = Stream.concat(requisitionedSamples.stream(), supplementalSamples.stream())
+        .map(Sample::getId)
+        .collect(Collectors.toSet());
+
+    List<Sample> stockSamples = sampleService.getChildren(sampleIds, SampleStock.CATEGORY_NAME, id);
+    List<Sample> tissueSamples = sampleService.getChildren(sampleIds, SampleTissue.CATEGORY_NAME, id);
+    List<Sample> aliquotSamples = sampleService.getChildren(sampleIds, SampleAliquot.CATEGORY_NAME, id);
```
djcooke, line 222: Missing tissue processing. Can also improve performance by going top-down (tissue > processing > stock > aliquot), mapping to IDs immediately since that's all you need, and searching based on the previous level. e.g. to find aliquots, you can use the stock IDs so it doesn't have to search through all the upper levels of the hierarchy to get there.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleRestController.java
```diff
@@ -203,6 +203,32 @@ public DataTablesResponseDto<SampleDto> getDTArrayedSamplesByProject(@PathVariab
...
+
+    List<Sample> allSamples = Stream.of(stockSamples.stream(), tissueSamples.stream(), aliquotSamples.stream())
+        .flatMap(stream -> stream)
+        .collect(Collectors.toList());
+
+    return jQueryBackend.get(request, advancedSearchParser, PaginationFilter.ids(
+        allSamples.stream().map(Sample::getId).collect(Collectors.toList())));
```
djcooke, line 229: Map to IDs earlier so you don't need to stream and collect again here just for that

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleRestController.java
```diff
@@ -203,6 +203,42 @@ public DataTablesResponseDto<SampleDto> getDTArrayedSamplesByProject(@PathVariab
...
+    Set<Long> tissueProcessingIds = tissueProcessingSamples.stream()
+        .map(Sample::getId)
+        .collect(Collectors.toSet());
+    allDerivativeSampleIds.addAll(tissueProcessingIds);
+    List<Sample> stockSamples = sampleService.getChildren(rootSampleIds, SampleStock.CATEGORY_NAME, id);
+    Set<Long> stockSampleIds = stockSamples.stream()
+        .map(Sample::getId)
+        .collect(Collectors.toSet());
+    allDerivativeSampleIds.addAll(stockSampleIds);
+    List<Sample> aliquotSamples = sampleService.getChildren(stockSampleIds, SampleAliquot.CATEGORY_NAME, id);
```
djcooke, line 233: Sorry, but my suggestion you've implemented to do the lookup from the previous level isn't actually going to work. It's possible that there will be no stock children (e.g. if stocks were requisitioned), so then this wouldn't find the aliquots. Similar problem for the other levels too. All will need to use `rootSampleIds`.

It would be better to have a single service level method that gets all the samples - or even just their IDs - but I think this is reasonably fast, so we don't need to do that for now.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/SampleRestController.java
```diff
@@ -210,33 +210,17 @@ public DataTablesResponseDto<SampleDto> getPreparedSamplessByRequisition(@PathVa
...
-    Set<Long> aliquotSampleIds = aliquotSamples.stream()
-        .map(Sample::getId)
-        .collect(Collectors.toSet());
-    allDerivativeSampleIds.addAll(aliquotSampleIds);
+    List<Sample> stockSamples = sampleService.getChildren(sampleIds, SampleStock.CATEGORY_NAME, id);
+    List<Sample> tissueSamples = sampleService.getChildren(sampleIds, SampleTissue.CATEGORY_NAME, id);
+    List<Sample> aliquotSamples = sampleService.getChildren(sampleIds, SampleAliquot.CATEGORY_NAME, id);
+    List<Sample> allSamples = Stream.of(stockSamples.stream(), tissueSamples.stream(), aliquotSamples.stream())
+        .flatMap(stream -> stream)
+        .collect(Collectors.toList());
```
djcooke, line 221: Map to IDs before collecting so you don't need to stream again right below this

djcooke:
Missing a change file

changes/add_GLT-4229.md
```diff
@@ -0,0 +1 @@
+Prepared Samples table on Edit Requisition Page to show all downstream samples
```
djcooke, line 1: This doesn't mention what happened to the Extractions table. Should be a `change` type and note that the Extractions table was changed to Prepared Samples. Try to include "(detailed sample)" at the end of any notes that only affect detailed sample mode too

djcooke (approved):

PR #3013, opened 2025-03-07
Reapply "[GLT-4231] added new fields for single cell (#3004)"

wuall826:
Jira ticket or GitHub issue:

- [ ] Includes a change file


avarsava:
Why was this reverted and why is it ok to apply now?

djcooke:
@avarsava We were unsure of whether this change was the best solution and wanted to do a release, so we reverted it temporarily. After further discussion, we have decided to go ahead with it.

@wuall826 Can you please update actions/cache to v4 in all of the Github workflows (in `.github/workflows`) to fix the CI?

avarsava (approved):

wuall826:
> @avarsava We were unsure of whether this change was the best solution and wanted to do a release, so we reverted it temporarily. After further discussion, we have decided to go ahead with it.
> 
> @wuall826 Can you please update actions/cache to v4 in all of the Github workflows (in `.github/workflows`) to fix the CI?

yep np

PR #3015, opened 2025-03-14
GLT-4360 Changed PacBio run page table names

Christopher-Ho:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4360

- [x] Includes a change file

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/type/PlatformType.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/type/PlatformType.java
@@ -74,7 +74,7 @@ public enum PlatformType {
       return new IonTorrentRun();
     }
   }, //
-  PACBIO("PacBio", "SMRT Cells", "SMRT Cell Contents", "SMRT Cell Contents", null) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/type/PlatformType.java
```diff
@@ -74,7 +74,7 @@ public Run createRun() {
       return new IonTorrentRun();
     }
   }, //
-  PACBIO("PacBio", "8Pac", "SMRT Cell", "SMRT Cells", null) {
+  PACBIO("PacBio", "SMRT Cells", "SMRT Cell Contents", "SMRT Cell Contents", null) {
```
djcooke, line 77: `containerName` ("SMRT Cell") should be singular to match the others, as this is used in other places too. We do want the table name to be plural on the Edit Run page though. To do that, you can edit `editRun.jsp` and add an "s" to the name of `list_container`. Adding an "s" works for all platforms, so no need to add another property to the enum.

djcooke (approved):

avarsava (approved):

PR #3018, opened 2025-03-17
[GLT-4232] show parent notes on edit sample/library page

wuall826:
Jira ticket or GitHub issue:

- [ ] Includes a change file


wuall826:
branch clean up. don't review yet

miso-web/src/main/webapp/scripts/list_note.js
```diff
@@ -0,0 +1,208 @@
...
+
+      if (!id || !entityType) {
+        console.error("Error: Missing entity ID or type for notes");
+        return null;
+      }
+
+      return "/rest/notes/" + entityType + "/" + id + "/dt?includeRelated=true";
+    },
+    getQueryUrl: function () {
+      return "/rest/notes/search";
```
djcooke, line 16: This is for the "Select by Names" feature, which doesn't make sense for this table. You can omit `getQueryUrl` entirely

miso-web/src/main/webapp/scripts/list_note.js
```diff
@@ -0,0 +1,208 @@
...
+              Utils.showOkDialog("Error", ["Cannot delete notes from parent entities."]);
+              return;
+            }
+            var lines = ["Are you sure you want to delete these notes? This cannot be undone."];
+            var ids = [];
+            jQuery.each(currentEntityNotes, function (index, note) {
+              lines.push("* " + note.text);
+              ids.push(note.id);
+            });
+            var url = "/rest/notes/" + entityType + "/" + id + "/" + ids.join(",");
```
djcooke, line 48: This is a weird URL. IDs to delete should be provided in the request body like other bulk delete requests. See `createBulkDeleteAction` in `list.js` for example. That function won't work here, but this should follow the same pattern.

miso-web/src/main/webapp/scripts/list_note.js
```diff
@@ -0,0 +1,208 @@
...
+    ],
+    processData: function (data) {
+      return Array.isArray(data) ? data : data.data || [];
+    },
+    bServerSide: true,
+    bStateSave: false,
+  };
+})();
+
+Utils.notes = {
```
djcooke, line 160: This is defined in `lims.js`. Overwriting it here is confusing, and since you removed `addNote`, will break the other note tables in MISO

wuall826:
should I keep the source 
image
the same as it is now (for directly sample it shows sample then for parent sample it shows parent) or name it all parent sample?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/NoteRestController.java
```diff
@@ -70,41 +84,208 @@ public void setText(String text) {
...
+    @SuppressWarnings("unchecked")
+    NoteService<Identifiable> typedService = (NoteService<Identifiable>) service;
+
+    for (Long noteId : request.getIds()) {
+      typedService.deleteNote(entity, noteId);
+    }
+  }
+
+
+  private List<NoteDto> getNotesList(String entityType, long entityId, boolean includeRelated) throws IOException {
```
djcooke, line 226: Can make this much simpler and avoid all the bad casting and type checking by checking `entityType` at the start and selecting the appropriate service yourself. If it's a library, use `libraryService.get` directly to get the library, `library.getNotes` to get its notes, `library.getParent` to get its parent sample, etc. Throw an exception if `entityType` isn't sample/library since those are the only types this is expected to handle.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/NoteRestController.java
```diff
@@ -70,41 +84,208 @@ public void setText(String text) {
   @Autowired
   private WorksetService worksetService;
 
-  private <T extends Identifiable> void addNote(NoteService<T> service, long entityId, NoteRequest request)
```
djcooke, line 73: Why was this changed? It was cleaner before

PR #3020, opened 2025-03-18
[GLT-4335] print barcodes by column

wuall826:
Jira ticket or GitHub issue:

- [ ] Includes a change file

### DIFF
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/PrinterRestController.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/PrinterRestController.java
@@ -246,6 +246,32 @@ public class PrinterRestController extends RestController {
             .map(WhineyFunction.rethrow(this::loadBarcodable)));
   }
 
+  @PostMapping(value = "{printerId}/boxpositionsbycolumn", headers = {"Content-type=application/json"})
+  @ResponseBody
+  public long boxPositionsByColumn(@PathVariable("printerId") Long printerId,
+      @RequestBody BoxPositionPrintRequest request)
+      throws IOException {
+    User user = authorizationManager.getCurrentUser();
+    Printer printer = printerService.get(printerId);
+    if (printer == null) {
+      throw new RestException("No printer found with ID: " + printerId, Status.NOT_FOUND);
+    }
+    Box box = boxService.get(request.getBoxId());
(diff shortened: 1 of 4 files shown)


djcooke (commented):
This probably works, but now we have two buttons with long names on a table that sometimes already has a lot of buttons, and a bunch of duplicated code.

Please change it back to the single button and allow selecting "by row" or "by column" in the dialog. You could modify the `printSelectDialog` function to allow adding custom options, but the easier solution would be to just add a second dialog, and that's fine too. One controller method should handle both, look at the parameter, and sort based on that.

callunity (approved):

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -471,31 +471,63 @@ Box.ui = {
...
+          var dialog = jQuery("#dialogDialog");
+          jQuery("#dialogInfoBelow").html("");
+          jQuery("#dialogInfoAbove").html(
+            "<p>Please select how you would like to sort the barcodes for printing.</p>"
+          );
+          jQuery("#dialogVisual").html(
+            '<p><input type="radio" name="sortOrder" id="sortByRow" value="row" checked/><label for="sortByRow"> Sort by Row</label> ' +
+              '<input type="radio" name="sortOrder" id="sortByColumn" value="column"/><label for="sortByColumn"> Sort by Column</label></p>'
+          );
+          dialog.dialog({
```
djcooke, line 483: Use the dialog functions from `lims.js`, which take care of most of the setup you've done here, offer additional features, and keep things consistent

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/PrinterRestController.java
```diff
@@ -246,6 +246,32 @@ public long boxPositions(@PathVariable("printerId") Long printerId, @RequestBody
             .map(WhineyFunction.rethrow(this::loadBarcodable)));
   }
 
+  @PostMapping(value = "{printerId}/boxpositionsbycolumn", headers = {"Content-type=application/json"})
+  @ResponseBody
+  public long boxPositionsByColumn(@PathVariable("printerId") Long printerId,
+      @RequestBody BoxPositionPrintRequest request)
```
djcooke, line 252: We don't need a new endpoint for this. Make sort order a field in `BoxPositionPrintRequest` and handle that in the existing endpoint.

callunity (approved):

PR #3028, opened 2025-03-31
[GLT-4364] add new sequencing parameter

Christopher-Ho:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4364

- [x] Includes a change file

### DIFF
--- a/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSequencingParametersService.java
+++ b/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSequencingParametersService.java
@@ -160,8 +160,8 @@ public class DefaultSequencingParametersService extends AbstractSaveService<Sequ
             + "to 0"));
       }
     } else if (params.getMovieTime() != null) {
-      errors.add(new ValidationError("movieTime", "Movie time should be omitted for non-PacBio "
-          + "instruments"));
+      errors.add(new ValidationError("movieTime",
+          "Movie time is not valid for " + params.getInstrumentModel().getPlatformType().getKey()));
     }
   }
 


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
```diff
@@ -68,6 +68,9 @@ public class SequencingParameters
   @JoinColumn(name = "updatedBy", nullable = false)
   private User updatedBy;
 
+  @Column(nullable = false)
+  private int movieTime;
```
djcooke, line 72: Should be nullable (`Integer`)

djcooke:
Build is also failing currently

sqlstore/src/main/resources/db/migration/V8123__SequencerPartitionContainer.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE SequencerPartitionContainer ADD COLUMN sequencingParametersId bigint,
+ADD CONSTRAINT `FK_Spc_Sp` FOREIGN KEY
+(sequencingParametersId) REFERENCES SequencingParameters
+(parametersId);
+ALTER TABLE SequencingParameters ADD COLUMN movieTime bigint;
```
djcooke, line 5: `int` or even `mediumint` should be more than sufficient. Could even be unsigned to double the range since it can't be negative

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSequencingParametersService.java
```diff
@@ -148,13 +149,25 @@ private void validateRunType(SequencingParameters params, List<ValidationError>
     }
   }
 
+  private void validateMovieTime(SequencingParameters params, List<ValidationError> errors) {
+    if(params.getInstrumentModel().getPlatformType() == PlatformType.PACBIO) {
+      if(params.getMovieTime() < 0) {
+        errors.add(new ValidationError("movieTime", "Movie time must be greater than 0"));
+      }
+    } else if (params.getMovieTime() != null) {
```
djcooke, line 157: * Check for null first. If movie time is less than zero, this will currently throw a `NullPointerException`.
* You're checking whether the value is less than zero, but the message says it must be greater than zero. Is zero allowed or not? I don't think it should be.

miso-web/src/main/webapp/scripts/bulk_sequencingparameters.js
```diff
@@ -102,6 +102,13 @@ BulkTarget.sequencingparameters = (function () {
           data: "runType",
           maxLength: 255,
         },
+        {
+          title: "Movie Time",
```
djcooke, line 106: Update the Instrument Model field's `onChange` to enable/disable this as appropriate, similar to what it does for the ONT Run Type field. While you're at it, please add the missing space in the title "Instrument Model"

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
```diff
@@ -224,6 +227,7 @@ public int hashCode() {
     result = prime * result + readLength;
     result = prime * result + readLength2;
     result = prime * result + ((runType == null) ? 0 : runType.hashCode());
+    result = prime * result + movieTime;
```
djcooke, line 230: Need to check for null

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSequencingParametersService.java
```diff
@@ -148,13 +149,29 @@ private void validateRunType(SequencingParameters params, List<ValidationError>
     }
   }
 
+  private void validateMovieTime(SequencingParameters params, List<ValidationError> errors) {
+    if(params.getInstrumentModel().getPlatformType() == PlatformType.PACBIO) {
+      if(params.getMovieTime() == null) {
+        errors.add(new ValidationError("movieTime", "Movie time value must be specified for "
+            + "PacBio instruments"));
+      }
+      if(params.getMovieTime() < 0) {
```
djcooke, line 158: This needs to be `else if`; otherwise, it gets checked even if the movietime is null, leading to the NPE you sent me on Slack.

sqlstore/src/main/resources/db/migration/V8123__SequencerPartitionContainer.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE SequencerPartitionContainer ADD COLUMN sequencingParametersId bigint,
+ADD CONSTRAINT `FK_Spc_Sp` FOREIGN KEY
```
djcooke, line 2: Please name this similar to other recently added foreign keys in MISO: `fk_sequencerPartitionContainer_sequencingParameters`

djcooke (approved):

sqlstore/src/main/resources/db/migration/V8123__SequencerPartitionContainer.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE SequencerPartitionContainer ADD COLUMN sequencingParametersId bigint,
+ADD CONSTRAINT `fk_sequencerPartitionContainer_sequencingParameters` FOREIGN KEY
+(sequencingParametersId) REFERENCES SequencingParameters
+(parametersId);
+ALTER TABLE SequencingParameters ADD COLUMN movieTime int UNSIGNED;
```
callunity, line 5: Right now, if anyone attempts to save any of the existing "Custom (see notes)" sequencing parameters for PacBio, they'll get an error because no movie time is set. Is there a default value that would make sense to apply to existing sequencing parameters?

miso-web/src/main/webapp/scripts/bulk_sequencingparameters.js
```diff
@@ -102,6 +111,13 @@ BulkTarget.sequencingparameters = (function () {
...
```
callunity, line 115: What about "PACBIO Movie Time", to go with the "ONT Run Type" just beside it?

callunity (commented):
Right now, the sequencing parameters for Illumina show read lengths. It would be nice if the sequencing parameters for PacBio could show the movie time.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSequencingParametersService.java
```diff
@@ -148,13 +149,30 @@ private void validateRunType(SequencingParameters params, List<ValidationError>
...
+        errors.add(new ValidationError("movieTime", "Movie time value must be specified for "
+            + "PacBio instruments"));
+      }
+      else if(params.getMovieTime() < 0) {
+        errors.add(new ValidationError("movieTime", "Movie time must be greater than or equal "
+            + "to 0"));
+      }
+    } else if (params.getMovieTime() != null) {
+      errors.add(new ValidationError("movieTime", "Movie time should be omitted for non-PacBio "
+          + "instruments"));
```
callunity, line 164:
```suggestion
      errors.add(new ValidationError("movieTime", "Movie time is not valid for " + params.getInstrumentModel().getPlatformType().getKey()
          + " instruments"));
```

PR #3038, opened 2025-04-16
[GLT-4377] added Ultima sequencing parameters Flows field

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4377

- [x] Includes a change file

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
@@ -71,6 +71,8 @@ public class SequencingParameters
   @Column
   private Integer movieTime;
 
+  private Integer flows;
+
   public IlluminaChemistry getChemistry() {
     return chemistry;
   }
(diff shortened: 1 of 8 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
```diff
@@ -71,6 +71,8 @@ public class SequencingParameters
   @Column
   private Integer movieTime;
 
+  private Integer flows;
```
callunity, line 74: Does this need the `@Column` annotation?

callunity (approved):

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SequencingParameters.java
djcooke: `@Column` is only really useful if it has parameters. If you want Hibernate to ignore a field, you need to tell it that explicitly (probably via `@Transient`)

PR #3039, opened 2025-04-17
[GLT-4383] Update Sequencing Parameters to Run Partition Container

Christopher-Ho:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4383

- [ ] Includes a change file

### DIFF
--- a/miso-web/src/main/webapp/scripts/form_container.js
+++ b/miso-web/src/main/webapp/scripts/form_container.js
@@ -71,17 +71,6 @@ FormTarget.container = (function ($) {
               },
             },
             {
-              title: "Sequencing Parameters",
-              data: "sequencingParametersId",
-              type: "dropdown",
-              nullLabel: "SELECT",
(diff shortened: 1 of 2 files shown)


miso-web/src/main/webapp/scripts/form_container.js
```diff
@@ -70,6 +70,17 @@ FormTarget.container = (function ($) {
                 });
               },
             },
+            {
+              title: "Sequencing Parameters",
```
djcooke, line 74: This does not belong on the container form because it's not a property of the container itself.

djcooke (commented):
I think for this PR, don't include the field in the UI at all, and don't include a changelog. This will be just be a fix to the database migration and model so that we don't release those as they currently are.

Next, a separate PR should allow actually setting the parameters when adding a container to a run if appropriate. This should include determining when it is appropriate, depending on the platform.

avarsava (approved):

PR #3044, opened 2025-05-06
Gp 4509.samplesheetdragen

grapigeau:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GP-4509

Note: I will rebase after fixes are complete

Looking for feedback:
Download file name (this is standard from outside samplesheets atm, can it change to be more readable?)

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
@@ -4,6 +4,7 @@ import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -14,9 +15,9 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import uk.ac.bbsrc.tgac.miso.core.data.LibraryIndex;
 import uk.ac.bbsrc.tgac.miso.core.data.IndexedLibrary;
 import uk.ac.bbsrc.tgac.miso.core.data.InstrumentDataManglingPolicy;
+import uk.ac.bbsrc.tgac.miso.core.data.LibraryIndex;
 import uk.ac.bbsrc.tgac.miso.core.data.Pair;
 import uk.ac.bbsrc.tgac.miso.core.data.Pool;
 import uk.ac.bbsrc.tgac.miso.core.data.SequencingParameters;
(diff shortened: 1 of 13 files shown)


miso-web/src/main/resources/miso.properties
```diff
@@ -103,3 +103,6 @@ miso.pools.error.index.mismatches.message:DUPLICATE INDICES
 #   so default warning number of mismatches is set to 2.
 miso.pools.warning.index.mismatches:2
 miso.pools.warning.index.mismatches.message:Near-Duplicate Indices
+
+## The version of Illumina Dragen to be used when exporting sample sheets
```
avarsava, line 107: ditto about DRAGEN

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/PoolRestController.java
```diff
@@ -622,8 +618,13 @@ public HttpEntity<byte[]> samplesheet(@RequestBody SampleSheetRequest request,
       HttpServletResponse response, UriComponentsBuilder uriBuilder) throws IOException {
     IlluminaExperiment experiment = IlluminaExperiment.valueOf(request.getExperimentType());
     SequencingParameters parameters = sequencingParametersService.get(request.getSequencingParametersId());
-    List<Pool> pools =
-        request.getPoolIds().stream().map(WhineyFunction.rethrow(poolService::get)).collect(Collectors.toList());
+    List<Pool> pools = new ArrayList<>();
+    for (Long poolId : request.getPoolIds()) {
+      if (poolId != null) {
```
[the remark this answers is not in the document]
grapigeau: Basically to keep there from being changes to the old samplesheets (if a dropdown is left as empty).

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/PoolRestController.java
```diff
@@ -632,6 +633,38 @@ public HttpEntity<byte[]> samplesheet(@RequestBody SampleSheetRequest request,
...
+      HttpServletRequest httpRequest,
+      HttpServletResponse response, UriComponentsBuilder uriBuilder) throws IOException {
+    IlluminaDragenExperiment experiment = IlluminaDragenExperiment.valueOf(request.getExperimentType());
+    SequencingParameters parameters = sequencingParametersService.get(request.getSequencingParametersId());
+
+    List<Pool> pools = new ArrayList<>();
+    List<Integer> lanes = new ArrayList<>();
+    var lane = 1;
+    for (Long poolId : request.getPoolIds()) {
+      if (poolId != null) {
```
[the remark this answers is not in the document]
grapigeau: Each dropdown here is representing a lane, so if a middle lane was empty then there would be a null in the middle. Not sure how realistic this situation is but I think it makes sense to allow/account for it.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaDragenExperiment.java
```diff
@@ -0,0 +1,303 @@
...
+import uk.ac.bbsrc.tgac.miso.core.data.IndexedLibrary;
+import uk.ac.bbsrc.tgac.miso.core.data.InstrumentDataManglingPolicy;
+import uk.ac.bbsrc.tgac.miso.core.data.LibraryIndex;
+import uk.ac.bbsrc.tgac.miso.core.data.Pair;
+import uk.ac.bbsrc.tgac.miso.core.data.Pool;
+import uk.ac.bbsrc.tgac.miso.core.data.SequencingParameters;
+import uk.ac.bbsrc.tgac.miso.core.data.impl.view.ParentLibrary;
+import uk.ac.bbsrc.tgac.miso.core.data.impl.view.PoolElement;
+
+public enum IlluminaDragenExperiment {
```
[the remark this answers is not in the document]
grapigeau: It definitely can, the original decision with Mike was not to touch `IlluminaExperiment` (mostly so we don't risk changing existing samplesheet generation. But if it makes more sense to do the refactoring, I'll go ahead with those changes

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
```diff
@@ -14,91 +14,434 @@
...
+      applyDragenCloud(data, dataColumns, headers,
+          sampleIdsCol, laneCol, index1Col,
+          projectCol, libraryCol, libraryPrepKitCol);
     }
   };
 
   private static final DateTimeFormatter MDY = DateTimeFormatter.ofPattern("M/d/yyyy");
 
   private final String description;
+  private final Boolean isDragen;
```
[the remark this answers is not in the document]
grapigeau: Originally had it so could be "null" but with the way it's setup now everything is binary DRAGEN or not so I can make it boolean for efficiency sake

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/ConstantsController.java
```diff
@@ -416,6 +423,8 @@ private void rebuildConstants() {
       node.put("errorEditDistance", indexChecker.getErrorMismatches());
       node.put("warningEditDistance", indexChecker.getWarningMismatches());
       node.put("genomeFolder", genomeFolder);
+      node.put("dragenVersion", dragenVersion);
+      node.put("trimUMI", trimUMI);
```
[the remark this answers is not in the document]
grapigeau: If they're not in constants, is there an example you could point to where a value from the properties file is being passed to a js dialog box?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/PoolRestController.java
```diff
@@ -622,16 +644,31 @@ public HttpEntity<byte[]> samplesheet(@RequestBody SampleSheetRequest request,
...
     return new HttpEntity<>(experiment
         .makeSampleSheet(request.getGenomeFolder(), parameters, request.getCustomRead1Primer(),
             request.getCustomIndexPrimer(),
             request.getCustomRead2Primer(),
-            pools)
+            pools,
+            lanes,
+            request.getDragenVersion(),
+            request.getTrimUMI(),
+            request.getFastQCompressionFormat())
```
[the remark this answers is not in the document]
grapigeau: The `SampleSheetRequest` is defined in the `PoolRestController` so I don't think I can pull it in like `makeSampleSheet(SampleSheetRequest request...` can I?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/ConstantsController.java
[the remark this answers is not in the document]
djcooke: You can add the necessary stuff to the list config. For the Pools list page, `ListPoolsController` uses `TabbedListItemsPage` to set up the list, but in a way that doesn't let you override the `writeConfiguration` method, so you'll have to change that.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/PoolRestController.java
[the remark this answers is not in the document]
djcooke: Oh, right. I guess leave as is for now. Again, this is not really maintainable though. More sample sheet formats will mean more and more parameters.

It would really be ideal if the `IlluminaExperiment` defined its required parameters, sent that to the front end, and received a `Map` or something similar back to handle appropriately. I'm not saying to make this change now, but it would be a large improvement and should probably be done at some point if we keep expanding this.

miso-web/src/main/webapp/scripts/lims.js
```diff
@@ -411,6 +411,49 @@ var Utils = Utils || {
           input.appendChild(compareTypeControl);
           input.appendChild(valueControl);
           break;
+        case "orderDropdowns":
```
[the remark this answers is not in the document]
grapigeau: In my mind this strategy made sense to simply have one property for all the lanes - a lot of my original changes were based on not actually modifying too much of the structure.

miso-web/src/main/webapp/scripts/0-urls.js
```diff
@@ -719,6 +719,7 @@ Urls = (function () {
     spreadsheet: poolRestBase + "/spreadsheet",
     contentsSpreadsheet: poolRestBase + "/contents/spreadsheet",
     samplesheet: poolRestBase + "/samplesheet",
+    dragenSamplesheet: poolRestBase + "/dragensamplesheet",
```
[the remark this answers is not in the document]
djcooke: Remove

docs/admin/site-configuration.md
```diff
@@ -387,3 +387,21 @@ Maximum number of mismatches to consider a near-duplicate index warning. Default
...
+
+### `miso.pools.samplesheet.dragenversion`
+
+The version of Illumina DRAGEN to be used when exporting sample sheets
+
+### `miso.pools.samplesheet.NovaSeqXSeries`
+
+The value that maps to the NovaSeqXSeries instrument
+
+### `miso.pools.samplesheet.compressionformat`
```
djcooke: Please use `camelCase` in the property keys for consistency

docs/admin/site-configuration.md
```diff
@@ -387,3 +387,21 @@ Maximum number of mismatches to consider a near-duplicate index warning. Default
...
+
+The fastq compression format to be used when exporting Illumina DRAGEN sample sheets. Defaults to "gzip" if unspecified.
```
djcooke: List the options

miso-web/src/main/webapp/scripts/lims.js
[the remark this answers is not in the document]
djcooke: I'd prefer using selects since this doesn't really do anything unique. If keeping this though:

1. `value` for other field types is the default value of the field. Here, it's used for the number of dropdowns to create. That should be called something else, and an error should be thrown if it's missing.
2. `required` doesn't work.
3. `null` should be returned for lanes with no pool, rather than "empty". Automatically including the "empty" option isn't great for adding this as a reusable field type either, but we're unlikely to use it anywhere else, so it probably doesn't matter.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/view/ListPoolsController.java
```diff
@@ -52,14 +62,53 @@ public class ListPoolsController {
...
+
+    public <T> TabbedListPoolsPage(String targetType, String property, Stream<T> tabItems, Comparator<String> tabSorter,
+        Function<T, String> getName, Function<T, Object> getValue, ObjectMapper mapper) {
+      super(targetType, property, tabItems, tabSorter, getName, getValue, mapper);
+    }
+
+    public TabbedListPoolsPage(String targetType, String property, SortedMap<String, String> tabs,
+        ObjectMapper mapper) {
+      super(targetType, property, tabs, mapper);
+    }
```
djcooke: Only the first constructor is used/needed

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
```diff
@@ -153,56 +463,64 @@ public final String makeSampleSheet(String genomeFolder, SequencingParameters pa
...
+    List<String> sampleWellCol = new ArrayList<>();
+    List<String> index1IdCol = new ArrayList<>();
+    List<String> index1Col = new ArrayList<>();
+    List<String> index2IdCol = new ArrayList<>();
+    List<String> index2Col = new ArrayList<>();
+    List<String> projectCol = new ArrayList<>();
+    List<String> genomeFolderCol = new ArrayList<>();
+    List<String> descriptionCol = new ArrayList<>();
+    List<String> libraryCol = new ArrayList<>();
+    List<String> libraryPrepKitCol = new ArrayList<>();
```
djcooke: Why all the separate lists instead of a class/record with all of these as fields?

docs/admin/site-configuration.md
```diff
@@ -387,3 +387,21 @@ Maximum number of mismatches to consider a near-duplicate index warning. Default
...
+
+The genomeFolder to be used when exporting sample sheets
+
+### `miso.pools.samplesheet.dragenversion`
+
+The version of Illumina DRAGEN to be used when exporting sample sheets
+
+### `miso.pools.samplesheet.NovaSeqXSeries`
+
+The value that maps to the NovaSeqXSeries instrument
```
djcooke:
```suggestion
The instrument model alias that maps to the NovaSeqXSeries instrument
```

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
```diff
@@ -650,4 +754,23 @@ private void writeListByColumns(final List<List<String>> columns, final StringBu
...
+  // if (entry.contains(",")) {
+  // output.append("\"").append(entry).append("\"").append(",");
+  // } else {
+  // output.append(entry).append(",");
+  // }
+  // }
+  // output.append("\n");
+  // }
+  // }
+  // }
```
djcooke: Remove commented code

miso-web/src/main/webapp/scripts/bulk_pool.js
```diff
@@ -349,79 +351,153 @@ BulkTarget.pool = (function ($) {
...
+          type: "text",
+          label: "Custom Read 2 Primer Well",
+          required: false,
+        },
+      ];
+
+      for (var i = 0; i < partitionCount; i++) {
+        dialogFields.push({
+          property: "pool_" + (i + 1),
+          label: "Pool " + (i + 1),
```
djcooke: Label should indicate that it's for the partition, ideally using the platform-specific term (e.g. "Lane" for Illumina). You have the instrument model ID, so you can use that to find the instrument model in `Constants.instrumentModels`, use the instrument model's `platformType` to find the platform in `Constants.platformTypes`, and then the platform's `partitionName` is what you want.

miso-web/src/main/webapp/scripts/bulk_pool.js
```diff
@@ -349,79 +351,153 @@ BulkTarget.pool = (function ($) {
...
+          required: true,
```
djcooke: It should be possible to leave a lane empty, so not required

miso-web/src/main/webapp/scripts/lims.js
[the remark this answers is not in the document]
djcooke: Remove if not using

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
```diff
@@ -428,6 +370,180 @@ private int getMaxLength(List<Pool> pools, int position) {
...
+    public List<String> sampleWellCol = new ArrayList<>();
+    public List<String> index1IdCol = new ArrayList<>();
+    public List<String> index1Col = new ArrayList<>();
+    public List<String> index2IdCol = new ArrayList<>();
+    public List<String> index2Col = new ArrayList<>();
+    public List<String> projectCol = new ArrayList<>();
+    public List<String> genomeFolderCol = new ArrayList<>();
+    public List<String> descriptionCol = new ArrayList<>();
+    public List<String> libraryCol = new ArrayList<>();
+    public List<String> libraryPrepKitCol = new ArrayList<>();
```
djcooke: This is an improvement because it's all contained in one object, but it's still a bunch of lists inside. What I meant was to have something like this:

```
public record SamplesheetSample(String sampleId, String lane, String plate, String well ...) {}
```

(or a class if you want to set field values separately, but I think a record works well here)

And then create a list of those. Makes it much simpler than ensuring all of those lists remain parallel, and you can leave fields null when it makes sense.

grapigeau: I'm not sure I understand the value of having one record for each sample when they're eventually going to need to be put in a data structure like `final Map<String, String> data = new LinkedHashMap<>();` in a call like `applyData()` before being written to the `output`; other than the risk of missing a value and potentially un-syncing the lists. Is that your main reasoning?
Null values will still have to be handled separately for different types of sample sheets.

djcooke: The lists are already a temporary data structure. The records would be the same thing, but cleaner and more maintainable. The main potential for error is with getting the lists out of sync, yes.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
```diff
@@ -14,91 +15,358 @@
...
+      applyDragenHeader(header, experimentName, instrument, novaSeqXSeriesMapping);
+    }
+
+    protected void applySettings(Map<String, String> settings, String read1Primer, String indexPrimer,
+        String read2Primer, String overrideCycles, String dragenVersion, String trimUMI, String fastqCompression) {
+      settings.put("\n[BCLConvert_Settings]", "");
+      settings.put("FastqCompressionFormat", fastqCompression);
+      // TODO: Can only be shown if a UMI is present for at least 1 sample (according to OverrideCycles
+      // setting). Not sure how best to enable.
+      // settings.put("TrimUMI", trimUMI);
```
djcooke: Libraries have a boolean field indicating whether they have UMIs. Does that help?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
```diff
@@ -14,91 +15,358 @@
...
+        List<SamplesheetSample> samples) {
+      data.put("\n[Cloud_Settings]", "");
+      data.put("GeneratedVersion", "1.16.0.202410292136");
+      data.put("\n[Cloud_Data]", "");
+      Collections.addAll(headers, "Sample_ID", "ProjectName", "LibraryName");
+      for (SamplesheetSample sample : samples) {
+        List<String> currRow = new ArrayList<>();
+        Collections.addAll(currRow, sample.sampleId,
+            sample.project, sample.library);
+        dataColumns.add(currRow);
```
djcooke: It looks like "dataColumns" is a collection of rows, rather than columns?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
```diff
@@ -222,53 +525,63 @@ public final String makeSampleSheet(String genomeFolder, SequencingParameters pa
...
+              paddedIndices.getKey().getKey(), // Index1 ID
+              paddedIndices.getKey().getValue(), // Index 1
+              index2Id, // Index2 ID
+              index2, // Index2
+              element.getAliquot().getProjectCode(), // Project
+              genomeFolder, // Genome Folder
+              element.getAliquot().getAliquotBarcode() != null ? element.getAliquot().getAliquotBarcode()
+                  : null, // Description
+              library.getName(), // Library
+              null // Library Prep Kit
```
djcooke, line 567: Is this intentionally left null? I think what you'd want is the name of the library's kit descriptor

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
```diff
@@ -222,53 +525,63 @@ public final String makeSampleSheet(String genomeFolder, SequencingParameters pa
...
+              null, // Well
+              paddedIndices.getKey().getKey(), // Index1 ID
+              paddedIndices.getKey().getValue(), // Index 1
+              index2Id, // Index2 ID
+              index2, // Index2
+              element.getAliquot().getProjectCode(), // Project
+              genomeFolder, // Genome Folder
+              element.getAliquot().getAliquotBarcode() != null ? element.getAliquot().getAliquotBarcode()
+                  : null, // Description
+              library.getName(), // Library
```
djcooke: Note that MISO terms don't necessarily match samplesheet terms.

- "Sample ID" here is the MISO library aliquot alias, which is more of a "name" field
- "Library name" here is the MISO library name, which is more of an "ID" field
- "Sample" is usually higher up in the hierarchy than "Library", but they're reversed here

I'm not sure any of these are problems, as I'm not sure what is expected in the samplesheets. Things to consider though.

grapigeau: Yeah the comments are matching the Illumina expected terms which is kind of confusing but correct in the sample sheet :woman_shrugging:

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
grapigeau: Yes this is intentional. DRAGEN doesn't use the Library Prep Kit and the other Illumina sample sheets had it deliberately left null and I don't want to change those (maybe we will at some point but not now).

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/util/IlluminaExperiment.java
djcooke: Sorry, I wasn't questioning the terms - just whether library aliquot alias and library name were the correct MISO fields to use. I don't think the MISO library name/ID is actually useful to the instrument or for analysis since it's the library aliquot that is actually sequenced and needs to be identified

grapigeau: The sampleId being element.getAliquot().getAlias() is the correct MISO field (at least based on how it was being filled out originally which worked for CLONE_CHECKING ect). 
The library name I'm not sure about, you're likely right and it's not useful but that's only being used in Cloud settings and since we're not running on the cloud I don't actually know what the correct field would be (and ultimately I suppose it doesn't matter until / if we run on cloud).

djcooke: It doesn't matter for us, but others may use the feature too. If we're not sure what it should be, I think it's better to err on the side of more specific to avoid potential for unintentional duplication/merging, so maybe use library aliquot name?

djcooke (approved):

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/view/ListPoolsController.java
```diff
@@ -1,39 +1,47 @@
 /*
- * Copyright (c) 2012. The Genome Analysis Centre, Norwich, UK
- * MISO project contacts: Robert Davey @ TGAC
- * *********************************************************************
+ * Copyright (c) 2012. The Genome Analysis Centre, Norwich, UK MISO project contacts: Robert Davey @
```
avarsava: I think we remove these when we find them?
If not, should at least undo the formatting changes

PR #3052, opened 2025-06-09
GLT-4403 Controller Test Profile Setup

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4403

Added a test profile for the controller tests, created a new directory for the tests, and wrote one test to initialise the application.

### DIFF
--- a/miso-web/pom.xml
+++ b/miso-web/pom.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <artifactId>miso</artifactId>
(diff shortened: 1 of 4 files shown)


miso-web/src/spring-test/java/AbstractCT.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.PersistenceContext;
+
+
+@RunWith(SpringRunner.class)
+@ContextConfiguration("/it-context.xml")
+@PropertySource("classpath:/miso-web/src/main/resources/miso.properties")
```
djcooke: Use a properties file specifically for testing. `/miso-web/src/it/resources/tomcat-config/miso.it.properties` should be fine to share. You should be able to reference this as just `/tomcat-config/miso.it.properties`, similar to `/it-context.xml` above

miso-web/pom.xml
```diff
@@ -918,5 +916,146 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>controller-testing</id>
```
djcooke: Let's call this "spring-test" and "ST" instead of controller testing and CT. The tests we're wanting to create will call controller methods, but they're technically testing more than just the controllers. We may also decide to add tests on other components such as services in the future.

miso-web/pom.xml
```diff
@@ -918,5 +916,146 @@
...
+              </execution>
+            </executions>
+          </plugin>
+          <plugin>
+            <groupId>io.fabric8</groupId>
+            <artifactId>docker-maven-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>start</id>
+                <phase>pre-controller-test</phase>
```
djcooke: This isn't a valid phase and the plugin isn't running. Same for flyway-maven-plugin below. Since these are running like unit tests (Surefire plugin) `process-test-resources` is probably the phase to use to get the database set up before the tests run.

When you run the tests, you should see messages in the log about the Docker container starting and migrations running on the database. There are connection exceptions currently showing up because JDBC is failing to connect to the database since it doesn't exist.

miso-web/src/spring-test/java/AbstractST.java
```diff
@@ -0,0 +1,70 @@
...
+    Resource clearData = new FileSystemResource(getScript(CLEAR_DATA_SCRIPT));
+    Resource testData = new FileSystemResource(getScript(isDetailedSampleMode() ? DETAILED_SCRIPT : PLAIN_SCRIPT));
+    ResourceDatabasePopulator populator = new ResourceDatabasePopulator(clearData, testData);
+    populator.execute(dataSource);
+  }
+
+
+  protected boolean isDetailedSampleMode() {
+    return true;
+  }
```
djcooke: I think just assume detailed sample mode, since that's specified in the properties file. If the value here didn't match the properties file, there would be inconsistencies in the application that could interfere with testing.

We can just work within detailed sample mode for now, and if we want to add tests for plain sample mode in the future, we can add a different abstract class similar to this one that uses a different properties file and loads the plain sample data.

miso-web/src/spring-test/java/AbstractST.java
```diff
@@ -0,0 +1,65 @@
...
+@ContextConfiguration("/it-context.xml")
+@PropertySource("/tomcat-config/miso.it.properties")
+public abstract class AbstractST {
+  private static final Logger log = LoggerFactory.getLogger(AbstractST.class);
+
+  private static final String SCRIPT_DIR = System.getProperty("basedir") + "/src/it/resources/db/migration/";
+
+
+  private static final String CLEAR_DATA_SCRIPT = "clear_test_data.sql";
+  private static final String PLAIN_SCRIPT = "plainSample_integration_test_data.sql";
```
djcooke: Unused - remove

PR #3061, opened 2025-06-30
[GLT-4408] added projects page spring tests

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4408

Added Spring Tests for ProjectRestController.java


miso-web/pom.xml
```diff
@@ -953,7 +970,6 @@
                   <resources>
                     <resource>
                       <filtering>true</filtering>
-                      <!-- uses the same resources as the integration tests -->
```
djcooke, line 956: This is a useful comment to keep

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/util/MisoPropertyExporter.java
```diff
@@ -1,22 +1,19 @@
...
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
+ * MISO is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
  *
- * You should have received a copy of the GNU General Public License
- * along with MISO.  If not, see <http://www.gnu.org/licenses/>.
+ * You should have received a copy of the GNU General Public License along with MISO. If not, see
+ * <http://www.gnu.org/licenses/>.
```
djcooke: Please remove copyright notice/license

miso-web/src/spring-test/java/AbstractST.java
```diff
@@ -1,27 +1,51 @@
 package uk.ac.bbsrc.tgac.miso.webapp.springtest;
```
djcooke: This and other test classes should be moved into the proper source directory matching the package namespace

miso-web/src/spring-test/java/AbstractST.java
```diff
@@ -56,4 +97,18 @@ private File getScript(String filename) {
     return script;
   }
 
+  @Before
+  public void setup() {
+    this.mockMvc = webAppContextSetup(this.wac).build();
```
djcooke: Include this in the other `@Before` method so it's all together

miso-web/src/spring-test/java/ProjectRestST.java
```diff
@@ -0,0 +1,173 @@
...
+
+import java.util.List;
+import java.util.Arrays;
+import java.util.ArrayList;
+
+import org.springframework.test.web.servlet.MockMvc;
+
+public class ProjectRestST extends AbstractST {
+
+  private String controllerBase = "/rest/projects";
```
djcooke: Should be defined as a constant - static, final, and name in caps

miso-web/src/spring-test/java/ProjectRestST.java
```diff
@@ -0,0 +1,173 @@
...
+        .andExpect(jsonPath("$.creationDate").value("2017-06-26"));
+  }
+
+  @Test
+  public void testGetBySearch() throws Exception {
+    getMockMvc().perform(get(controllerBase + "/search").param("q", "PRO").accept(MediaType.APPLICATION_JSON))
+        .andExpect(status().isOk())
+        .andExpect(jsonPath("$").exists())
+        .andExpect(content().contentType(MediaType.APPLICATION_JSON))
+        .andExpect(jsonPath("$.*", hasSize(17))); // 17 = number of projects in test script
```
djcooke: Make the search more specific to ensure that it's filtering some of the projects out

miso-web/src/spring-test/java/ProjectRestST.java
```diff
@@ -0,0 +1,173 @@
...
+
+  }
+
+  @Test
+  public void testGetAttachments() throws Exception {
+    getMockMvc().perform(get(controllerBase + "/1/files").accept(MediaType.APPLICATION_JSON)).andExpect(status().isOk())
+        .andExpect(content().contentType(MediaType.APPLICATION_JSON))
+        .andExpect(jsonPath("$").exists());
```
djcooke: Assert that there are no attachments

miso-web/src/spring-test/java/ProjectRestST.java
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  @Test
+  @WithMockUser(username = "admin", password = "admin", roles = {"ADMIN", "INTERNAL"})
+  public void testCreate() throws Exception {
+
+    // project impl is needed here to set the reference genome
```
djcooke: ProjectImpl is not needed, and you should create the DTO directly to have better control over exactly what is being included in the request. ProjectDto has a referenceId field, and that is all that's needed.

miso-web/src/spring-test/java/ProjectRestST.java
```diff
@@ -0,0 +1,173 @@
...
+    project.setCreationTime(new Date());
+    project.setLastModified(new Date());
+    project.setName("testproj");
+    project.setCode("TESTCODE");
+    project.setId(0L); // unsaved project ID
+
+    ProjectDto dto = Dtos.asDto(project, false);
+
+    getMockMvc().perform(post(controllerBase).contentType(MediaType.APPLICATION_JSON).content(jsonMaker(dto)))
+        .andExpect(status().isOk());
```
djcooke: Check the ID and code of the returned object

miso-web/src/spring-test/java/ProjectRestST.java
```diff
@@ -0,0 +1,173 @@
...
+
+    assertNotNull(currentSession().get(ProjectImpl.class, 200002)); // test that project was successfully created
```
djcooke: * use the ID returned above instead of hard-coding so this doesn't break when we add another project to the test data
* assert a couple attributes of the project to ensure it's the one you just saved

miso-web/src/spring-test/java/ProjectRestST.java
```diff
@@ -0,0 +1,173 @@
...
+
+  }
+
+  private String jsonMaker(Object dto) throws Exception {
```
djcooke: Make static and move to `AbstractST` as this will be useful for all REST tests. Rename `makeJson`, to clarify that this method is making the JSON - not returning a thing that makes JSON

miso-web/src/spring-test/java/ProjectRestST.java
```diff
@@ -0,0 +1,173 @@
...
+
+  @Test
+  public void testGetAttachments() throws Exception {
+    getMockMvc().perform(get(controllerBase + "/1/files").accept(MediaType.APPLICATION_JSON)).andExpect(status().isOk())
+        .andExpect(content().contentType(MediaType.APPLICATION_JSON))
+        .andExpect(jsonPath("$").exists());
+  }
+
+  @Test
+  @WithMockUser(username = "admin", password = "admin", roles = {"ADMIN", "INTERNAL"})
```
djcooke: For each test that requires admin, add another with a non-admin user and ensure that it returns 401 status (unauthorized)

miso-web/src/spring-test/java/ProjectRestST.java
```diff
@@ -0,0 +1,173 @@
...
+    return requestJson;
+  }
+
+  @Test
+  @WithMockUser(username = "admin", password = "admin", roles = {"ADMIN", "INTERNAL"})
+  public void testUpdate() throws Exception {
+
+    Project proj = currentSession().get(ProjectImpl.class, 1);
+    proj.setTitle("changed testing project");
+    ProjectDto dto = Dtos.asDto(proj, true);
```
djcooke: Looking up the existing project and converting to DTO here is ok as it saves a lot of unnecessary code, but manipulate the DTO rather than the entity model to keep it a little cleaner.

miso-web/src/spring-test/java/ProjectRestST.java
```diff
@@ -0,0 +1,173 @@
...
+  public void testBulkDelete() throws Exception {
+    List<Long> ids = new ArrayList<Long>(Arrays.asList(7L));
+
+    getMockMvc()
+        .perform(post(controllerBase + "/bulk-delete").contentType(MediaType.APPLICATION_JSON).content(jsonMaker(ids)))
+        .andDo(print())
+        .andExpect(status().isNoContent());
+
+    // now check that the project was actually deleted
+    assertNull(currentSession().get(ProjectImpl.class, 7));
```
djcooke: Add check to ensure it existed before deleting

miso-web/src/spring-test/java/ProjectRestST.java
```diff
@@ -0,0 +1,173 @@
...
+    getMockMvc().perform(get(controllerBase + "/dt").accept(MediaType.APPLICATION_JSON)
+        .param("iDisplayStart", "0")
+        .param("iDisplayLength", "25")
+        .param("mDataProp_0", "id")
+        .param("sSortDir_0", "asc")
+        .param("iSortCol_0", "3")
+        .param("sEcho", "1"))
+        .andExpect(status().isOk())
+        .andExpect(content().contentType(MediaType.APPLICATION_JSON))
+        .andExpect(jsonPath("$").exists());
```
djcooke: * Create a helper function in AbstractST to do all of this since it'll be needed for testing a lot of similar endpoints
* Check a few attributes of the returned object, including some project attributes

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/ProjectRestST.java
```diff
@@ -57,11 +51,11 @@
 
 public class ProjectRestST extends AbstractST {
 
-  private String controllerBase = "/rest/projects";
+  static final String CONTROLLER_BASE = "/rest/projects";
```
djcooke: Should still be private

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/ProjectRestST.java
```diff
@@ -74,100 +68,140 @@ public void testGetById() throws Exception {
...
-    project.setCreationTime(new Date());
-    project.setLastModified(new Date());
-    project.setName("testproj");
+    project.setPipelineId(1L);
+    project.setStatus("Active");
+    project.setReferenceGenomeId(1L);
+    project.setCreationDate("2025-07-02");
     project.setCode("TESTCODE");
-    project.setId(0L); // unsaved project ID
+    project.setId(0L);
```
djcooke: ID should not be set - will normally be null in the DTO. That will result in `0L`/`UNSAVED_ID` in `ProjectImpl` after conversion, but that's a detail that doesn't need to be known/handled here.

Don't need to set creationDate either

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/ProjectRestST.java
```diff
@@ -74,100 +68,140 @@ public void testGetById() throws Exception {
...
 
-    ProjectDto dto = Dtos.asDto(project, false);
 
-    getMockMvc().perform(post(controllerBase).contentType(MediaType.APPLICATION_JSON).content(jsonMaker(dto)))
-        .andExpect(status().isOk());
+    MvcResult result = getMockMvc()
+        .perform(post(CONTROLLER_BASE).contentType(MediaType.APPLICATION_JSON).content(AbstractST.makeJson(project)))
+        .andExpect(status().isOk())
+        .andExpect(jsonPath("$.id").value(200002))
```
djcooke: Check that the value is present, but don't check the specific value since that can change if we modify the test data.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/ProjectRestST.java
```diff
@@ -74,100 +68,140 @@ public void testGetById() throws Exception {
...
+  public void testCreateFail() throws Exception {
+    ProjectDto project = new ProjectDto();
+    project.setTitle("test title");
+    project.setPipelineId(1L);
+    project.setStatus("Active");
+    project.setReferenceGenomeId(1L);
+    project.setCreationDate("2025-07-02");
+    project.setName("testproj");
+    project.setCode("TESTCODE");
+    project.setId(0L);
```
djcooke: Don't set creationDate, name, or ID

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/ProjectRestST.java
```diff
@@ -74,100 +68,140 @@ public void testGetById() throws Exception {
...
+
+
+    getMockMvc()
+        .perform(post(CONTROLLER_BASE).contentType(MediaType.APPLICATION_JSON).content(AbstractST.makeJson(project)))
+        .andExpect(status().isInternalServerError());
```
djcooke: Hmm... Should be returning 401 (unauthorized) - not 500 (internal server error). It may be because the user you're providing in `@WithMockUser` doesn't exist. Try using a non-admin user that exists in the test data.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/ProjectRestST.java
```diff
@@ -74,100 +68,140 @@ public void testGetById() throws Exception {
...
+  public void testDeleteFail() throws Exception {
+    List<Long> ids = new ArrayList<Long>(Arrays.asList(7L));
+
+    // check that the project we want to delete exists
+    assertNotNull(currentSession().get(ProjectImpl.class, 7));
+
+    getMockMvc()
+        .perform(post(CONTROLLER_BASE + "/bulk-delete").contentType(MediaType.APPLICATION_JSON)
+            .content(AbstractST.makeJson(ids)))
+        .andExpect(status().isBadRequest());
```
djcooke: I'm not sure why you're getting 400 (bad request) here, especially since the request setup looks the same as the positive test. Should be returning 401 (unauthorized) though. I tested in MISO and got 401 as expected.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/AbstractST.java
```diff
@@ -111,4 +125,19 @@ protected MockMvc getMockMvc() {
     return mockMvc;
   }
 
+  protected ResultActions dtResponse(String controllerBase) throws Exception {
```
djcooke: Rename `performDtRequest` to clarify, and for consistency with other methods.

Might want to parameterize sort column and possibly other values, but that can happen later if useful.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/ProjectRestST.java
```diff
@@ -74,100 +68,140 @@ public void testGetById() throws Exception {
...
-        .param("mDataProp_0", "id")
-        .param("sSortDir_0", "asc")
-        .param("iSortCol_0", "3")
-        .param("sEcho", "1"))
-        .andExpect(status().isOk())
-        .andExpect(content().contentType(MediaType.APPLICATION_JSON))
-        .andExpect(jsonPath("$").exists());
+    ResultActions result = dtResponse(CONTROLLER_BASE);
+    result
+        .andExpect(jsonPath("$.iTotalRecords").value(17)); // 17 = number of projects in test data script
```
djcooke: Test a couple attributes of a project within the data. Sorting is specified, so you should be able to identify the first project that'll be included

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/ProjectRestST.java
PrabhjotChopra: I tried doing that, and it turns out that updating and creating do not require admin permissions (just needed internal permissions). I have removed those failure tests accordingly.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/ProjectRestST.java
PrabhjotChopra: I changed the user to a non-admin one that exists in the integration test data, and it passed. Turns out delete only requires "internal" permissions. I have deleted this test accordingly.

Should I add another user without internal or admin permissions to test the unauthorized failure?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/ProjectRestST.java
djcooke: Ah, ok! Please change the tests to use a regular user where admin is not needed then

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/ProjectRestST.java
djcooke: You have to be admin or the project's creator to delete. For the delete failure test, you can use a regular user who is internal, but not admin, and not the project creator.

Don't add a user without internal role, as that is required to even log into MISO

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ProjectRestST.java
```diff
@@ -51,7 +51,7 @@
 
 public class ProjectRestST extends AbstractST {
```
djcooke: Rename `ProjectRestControllerST` so it matches the class being tested

PR #3064, opened 2025-07-07
[GLT-4424] Increased SOP URL field size

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4424


sqlstore/src/main/resources/db/migration/V1090__auto_main.sql
```diff
@@ -28,7 +28,7 @@ CREATE TABLE Sop (
   alias varchar(100) NOT NULL,
   version varchar(50) NOT NULL,
   category varchar(20) NOT NULL,
-  url varchar(255) NOT NULL,
+  url varchar(500) NOT NULL,
```
callunity: This will cause a checksum mismatch when Flyway attempts to validate that this file is the same as what's already been applied to the database. Please instead create a new migration file that updates the `Sop` table to change the `url` field size.

sqlstore/src/main/resources/db/migration/V1690__auto_main.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE Sop
```
callunity: Ah, I don't think we would have mentioned this to you yet, but: the `compact-migrations.sh` script which runs during the release procedure compacts all migrations named like V8###_.*.sql and does the correct numbering for the given release (most useful for when we had multiple people working simultaneously on features requiring migrations). Could you please rename your file?

PrabhjotChopra: For clarification, what should I rename the file to? Would V8000_.sql be good?

callunity: I'd suggest something informative, like `V8000_update_url_field.sql`

callunity (approved):

sqlstore/src/main/resources/db/migration/V8000_update_url_field.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE Sop
```
callunity: Ah - actually, failing tests indicate the migration name needs a double-underscore after V8000

callunity (approved):

PR #3067, opened 2025-07-08
[GLT-4429] added SOP rest tests

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4429

Added spring tests to SopRestController.java

### DIFF
--- a/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
+++ b/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
@@ -126,13 +126,13 @@ public abstract class AbstractST {
     return mockMvc;
   }
 
-  protected String pollingResponse(String id, String controllerBase, String url) throws Exception {
+  protected String pollingResponse(String url) throws Exception {
     String response =
-        getMockMvc().perform(get(controllerBase + url + id)).andReturn().getResponse().getContentAsString();
+        getMockMvc().perform(get(url)).andReturn().getResponse().getContentAsString();
     String status = JsonPath.read(response, "$.status");
     while (status.equals("running")) {
       response =
-          getMockMvc().perform(get(controllerBase + url + id)).andReturn().getResponse().getContentAsString();
+          getMockMvc().perform(get(url)).andReturn().getResponse().getContentAsString();
       status = JsonPath.read(response, "$.status");
       Thread.sleep(1000);
     }
--- a/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/LabRestControllerST.java
+++ b/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/LabRestControllerST.java
@@ -68,7 +68,7 @@ public class LabRestControllerST extends AbstractST {
         .andReturn();
 
     String id = JsonPath.read(mvcResult.getResponse().getContentAsString(), "$.operationId");
-    String response = pollingResponse(id, CONTROLLER_BASE, "/bulk/");
+    String response = pollingResponse(CONTROLLER_BASE + "/bulk/" + id);
 
     Integer id1 = JsonPath.read(response, "$.data[0].id");
     Integer id2 = JsonPath.read(response, "$.data[1].id");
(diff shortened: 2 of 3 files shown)


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -125,8 +126,22 @@ protected MockMvc getMockMvc() {
     return mockMvc;
   }
 
-  protected ResultActions performDtRequest(String controllerBase, int displayLength, String dataProp, int sortCol) throws Exception {
-    return getMockMvc().perform(get(controllerBase + "/dt").accept(MediaType.APPLICATION_JSON)
+  protected String pollingResponse(String id, String CONTROLLER_BASE, String url) throws Exception {
```
avarsava: Formal parameters are not constants and shouldn't be all uppercase like this

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -125,8 +126,22 @@ protected MockMvc getMockMvc() {
...
+          getMockMvc().perform(get(controllerBase + url + id)).andReturn().getResponse().getContentAsString();
+      status = JsonPath.read(response, "$.status");
+      Thread.sleep(1000);
+    }
+    return response;
+  }
+
+  protected ResultActions performDtRequest(String url, int displayLength, String dataProp, int sortCol)
+      throws Exception {
+    return getMockMvc().perform(get(url).accept(MediaType.APPLICATION_JSON)
```
callunity, line 144: If this will always be to a `/dt` endpoint, the arg could be `baseUrl` and you could append `/dt` inside this function.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -125,8 +126,22 @@ protected MockMvc getMockMvc() {
     return mockMvc;
   }
 
-  protected ResultActions performDtRequest(String controllerBase, int displayLength, String dataProp, int sortCol) throws Exception {
-    return getMockMvc().perform(get(controllerBase + "/dt").accept(MediaType.APPLICATION_JSON)
+  protected String pollingResponse(String id, String controllerBase, String url) throws Exception {
```
avarsava: Sorry for missing this on first read, something like ""/bulk/"" is not a URL, only a path segment component within a URL. Having both 'controllerBase' and 'url' might be confusing for future test implementers. It's ok to have all 3 arguments, but the name should change

callunity: Additionally, this function only uses the 3 arguments to construct a URL. If you construct the URL at the top of this function, you can use it twice below and then you don't have to put URL path separators in the correct places in every single file when passing arguments to this function.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -125,8 +126,22 @@ protected MockMvc getMockMvc() {
...
+    while (status.equals("running")) {
+      response =
+          getMockMvc().perform(get(controllerBase + url + id)).andReturn().getResponse().getContentAsString();
+      status = JsonPath.read(response, "$.status");
+      Thread.sleep(1000);
+    }
+    return response;
+  }
+
+  protected ResultActions performDtRequest(String url, int displayLength, String dataProp, int sortCol)
```
avarsava, line 142: The calls in the tests to this function all have to form their own URLs by appending "/dt" and sometimes more path to CONTROLLER_BASE, which is inconsistent with pollingResponse

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/LabRestControllerST.java
```diff
@@ -51,7 +51,6 @@ public class LabRestControllerST extends AbstractST {
   private static final String CONTROLLER_BASE = "/rest/labs";
 
   @Test
-  @WithMockUser(username = "user", password = "user", roles = {"INTERNAL"})
```
callunity, line 54: Why remove the user info?

PrabhjotChopra: AbstractST is annotated with @WithMockUser with this user, so the only @WithMockUser annotations needed are overriding ones (e.g. if a test needed admin permissions)

callunity: Thanks for clarifying!

PR #3069, opened 2025-07-08
[GLT-4430] added admin-related rest tests

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4430

Spring tests for AdminRestController.java

### DIFF
--- a/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AdminRestControllerST.java
+++ b/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AdminRestControllerST.java
@@ -68,26 +68,6 @@ public class AdminRestControllerST extends AbstractST {
   }
 
   @Test
-  @WithMockUser(username = "admin", password = "admin", roles = {"INTERNAL", "ADMIN"})
-  public void testRegenBarcodes() throws Exception {
-
-    MvcResult mvcResult = getMockMvc().perform(post(CONTROLLER_BASE + "/barcode/regen"))
-        .andDo(print())
-        .andExpect(status().isOk()).andReturn();
-
-    assertNotNull(mvcResult.getResponse());
-
-  }
-
-  @Test
-  public void failRegenBarcodes() throws Exception {
-    getMockMvc().perform(post(CONTROLLER_BASE + "/barcode/regen"))
-        .andDo(print())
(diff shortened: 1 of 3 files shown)


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AdminRestControllerST.java
```diff
@@ -0,0 +1,82 @@
...
+
+
+  @Test
+  @WithMockUser(username = "admin", password = "admin", roles = {"INTERNAL", "ADMIN"})
+  public void testClearCache() throws Exception {
+
+
+    MvcResult mvcResult =
+        getMockMvc().perform(post(CONTROLLER_BASE + "/cache/clear")).andExpect(status().isOk()).andReturn();
+    // check if cache is cleared somehow
```
callunity: What do you think about updating this comment to `"Check that the response is consistent with what should be returned if the cache is cleared"? Since the test is not actually checking if the cache is cleared. (Nor should it - I believe this is more vestigial code from its initial TGAC days that was never actually used by OICR. We don't look to have any level-two caches implemented.)

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AdminRestControllerST.java
```diff
@@ -0,0 +1,82 @@
...
+    // I guess response body
+
+    assertEquals("true", mvcResult.getResponse().getContentAsString());
+  }
+
+  @Test
+  public void failClearCache() throws Exception {
+    getMockMvc().perform(post(CONTROLLER_BASE + "/cache/clear"))
+        .andDo(print())
+        .andExpect(status().isInternalServerError());
```
callunity: It would be preferable to have the method return 401 Unauthorized if the user is not admin, as that's a user issue and not a server issue. Could you make this change?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AdminRestControllerST.java
PrabhjotChopra: Oh sure will do, I think I accidentally left that comment in

PR #3070, opened 2025-07-09
[GLT-4432] array model rest tests added

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4432

ArrayModelRestController.java spring tests added, and template tests added for async endpoints.

### DIFF
--- a/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayModelRestControllerST.java
+++ b/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayModelRestControllerST.java
@@ -99,7 +99,7 @@ public class ArrayModelRestControllerST extends AbstractST {
     // array models can only be created by an admin user, so this test expected failure due to
     // insufficient permissions
 
-    abstractTestBulkCreateAsyncFail(CONTROLLER_BASE, controllerClass, makeCreateDtos());
+    abstractBulkCreateAsyncFail(CONTROLLER_BASE, controllerClass, makeCreateDtos());
 
   }
 


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -109,7 +114,15 @@ private File getScript(String filename) {
     return script;
   }
 
-  public static String makeJson(Object obj) throws Exception {
+  public String makeJson(Object obj) throws Exception {
```
avarsava, line 128: Aside from the signature, this method is identical to makeJson(List<?> obj). Object is the parent class of all objects, including Lists, so this duplication should be unneeded, could you change the signature of the existing method instead?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -139,6 +152,93 @@ protected String pollingResponse(String url) throws Exception {
     return response;
   }
 
+  protected <T> void abstractTestBulkCreateAsync(String controllerBase, Class<T> createType, List<?> dtos)
```
callunity, line 155: This test shares a lot of code (generating mvcResult through to getting the `response`) with the abstrctBulkCreateAsyncFail test. Can you extract the common code so the test really focuses on the places where they differ?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -109,7 +114,15 @@ private File getScript(String filename) {
     return script;
   }
 
-  public static String makeJson(Object obj) throws Exception {
+  public String makeJson(Object obj) throws Exception {
+    ObjectMapper mapper = new ObjectMapper();
```
callunity: Rather than have the method create a new ObjectMapper each time it is called (especially since it's configured with the same values in your two functions), you can set the mapper to be a class-level private static field.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -139,6 +152,93 @@ protected String pollingResponse(String url) throws Exception {
...
+    
+    MvcResult mvcResult = getMockMvc()
+          .perform(post(controllerBase + "/bulk").contentType(MediaType.APPLICATION_JSON).content(makeJson(dtos)))
+          .andExpect(status().isAccepted())
+          .andDo(print())
+          .andReturn();
+    String id = JsonPath.read(mvcResult.getResponse().getContentAsString(), "$.operationId");
+    String response = pollingResponse(controllerBase + "/bulk/" + id);
+    String status = JsonPath.read(response, "$.status");
+    assertEquals("failed", status); // request should fail without admin permissions
```
callunity, line 170: For tests that are expected to fail because admin permissions are required, then it's more accurate to test for `status().isUnauthorized()`, with a different test for the scenarios where they're expected to fail for a different reason.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -139,6 +152,93 @@ protected String pollingResponse(String url) throws Exception {
...
+        .andReturn();
+
+    String id = JsonPath.read(mvcResult.getResponse().getContentAsString(), "$.operationId");
+    String response = pollingResponse(controllerBase + "/bulk/" + id);
+
+    Integer id1 = JsonPath.read(response, "$.data[0].id");
+    Integer id2 = JsonPath.read(response, "$.data[1].id");
+
+    assertNotNull(currentSession().get(createType, id1));
+    assertNotNull(currentSession().get(createType, id2));
```
callunity: This implementation requires you to supply two (and only two) dtos to the function. Can you make this support any number of DTOs? (Same with abstractTestBulkUpdateAsync below)

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayModelRestControllerST.java
```diff
@@ -0,0 +1,122 @@
...
+  public void testBulkCreateAsync() throws Exception {
+    ArrayModelDto arrone = new ArrayModelDto();
+    arrone.setAlias("array model one");
+
+    ArrayModelDto arrtwo = new ArrayModelDto();
+    arrtwo.setAlias("array model two");
+
+    List<ArrayModelDto> dtos = new ArrayList<ArrayModelDto>();
+    dtos.add(arrone);
+    dtos.add(arrtwo);
```
callunity, line 64: If you extract this into its own function, then you can reuse that in the testBulkCreateFail method.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
PrabhjotChopra: I needed to add this to get the class template methods like AbstractBulkCreateAsync working, wasn't working before I added this overloaded method.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayModelRestControllerST.java
PrabhjotChopra: I would like to do that, but my concern is with the specific setter methods available to different objects, as I am not sure if all objects with a rest controller also have a setAlias method (I have not looked through the codebase enough to determine either way, and would appreciate some guidance here).

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
PrabhjotChopra: I spent a while trying to get it to throw the unauthorized exception and will take another stab at it, however my reasoning is down below for some background information.


The only difference between the create and create failure tests is the user assigned, so I think that it may be sufficient for admin permission testing (from what I have seen so far, the main testable creation failure that doesn't involve not setting non-nullable fields is admin permissions. If other creation fail (or update), I will go back and rework this.
I did spend a while trying to get it to throw unauthorized, and I repeatedly kept getting RuntimeExceptions, even after I started manually throwing authorization errors in the place where the exception is thrown, and ran into a bunch of issues in general with that approach.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
avarsava: Right, I understand that you need a method with this signature. However, with this method present, `makeJson(List<?> obj)` is now redundant because Object is a parent class of List and the logic in the method is identical. If you remove `makeJson(List<?> obj)` entirely the tests all still compile and pass, and redundant code should be removed.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayModelRestControllerST.java
callunity: You can keep the function in the individual controller ST files - so, `makeNewArrayModelDtos`. (Most of the classes have a setAlias method, but not all)

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/LabRestControllerST.java
```diff
@@ -49,6 +48,7 @@
 public class LabRestControllerST extends AbstractST {
 
   private static final String CONTROLLER_BASE = "/rest/labs";
+  private static final Class<LabImpl> controllerClass = LabImpl.class;
 
   @Test
   public void testBulkCreateAsync() throws Exception {
```
callunity, line 54: It's probably valuable to add a comment here that there are no restrictions on who can create a lab. (And makes it clearer why this class doesn't have a `testBulkCreateAsyncFail` test, unlike say the SopRestControllerST).

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
callunity: Ah, I see - looking at the AsyncOperationManager and BulkSaveOperation classes, they do convert all types of exceptions into the consistent "status: failed" that you're checking for here. This looks good.

callunity: Could you please update the comments though? If the test is expecting failure due to lack of admin permissions, that comment should be in the individual controller's ST file. It's possible that bulk create async may fail for other reasons as well.

PR #3072, opened 2025-07-10
[GLT-4635] increased library index field size

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4635

Also added fix from SOP url field size that didn't get released (was using V1690 naming scheme for the migration instead of V8___)


callunity:
The V1690 script got applied to the stage and production databases with the latest deploy, so please revert the rename as it's not needed.

avarsava (approved):

callunity (approved):

PR #3074, opened 2025-07-11
[GLT-4434] added array run rest tests

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4434

note: can only be reviewed after https://github.com/miso-lims/miso-lims/pull/3071 is merged


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayRunRestControllerST.java
```diff
@@ -0,0 +1,135 @@
...
+
+
+  @Test
+  public void testDtResponse() throws Exception {
+    performDtRequest(CONTROLLER_BASE + "/dt")
+        .andExpect(jsonPath("$.iTotalRecords").value(3))
+        .andExpect(jsonPath("$.aaData[0].alias").value("ArrayRun_1"))
+        .andExpect(jsonPath("$.aaData[2].instrumentName").value("iScan1"))
+        .andExpect(jsonPath("$.aaData[1].alias").value("to_delete"))
+        .andExpect(jsonPath("$.aaData[0].status").value("Running"));
```
callunity: Checking the attributes out of order here (0, 2, 1, 0) makes it harder to cross-reference against the test data. Can you please check them one row/item at a time here and in other methods?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayRunRestControllerST.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  @Test
+  public void testDtByProj() throws Exception {
+    performDtRequest(CONTROLLER_BASE + "/dt/project/3")
+        // project 3 has the sample that is shared by the test array runs
+        // all test array runs are a part of requisition 1
```
callunity: It doesn't look like the requisition is relevant to this particular test, so can the comment be removed?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayRunRestControllerST.java
```diff
@@ -0,0 +1,135 @@
...
+        .andExpect(jsonPath("$.aaData[0].alias").value("ArrayRun_1"))
+        .andExpect(jsonPath("$.aaData[2].instrumentName").value("iScan1"))
+        .andExpect(jsonPath("$.aaData[1].alias").value("to_delete"))
+        .andExpect(jsonPath("$.aaData[0].status").value("Running"));
+  }
+
+  @Test
+  public void testSave() throws Exception {
+    // "health" is a non-nullable field that does not have a setter in the dto, so I need to use the
+    // actual ArrayRun type
```
callunity: "health" [is determined based on the](https://github.com/miso-lims/miso-lims/blob/develop/miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java#L3493) DTO `status` attribute (so, status="Running"). You can use this to build the DTO directly.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -242,7 +244,37 @@ protected <T> void abstractTestDeleteFail(Class<T> deleteType, int id, String co
...
+  protected <T, D> T abstractUpdate(String controllerBase, D dto, int id, Class<T> updateType) throws Exception {
+    getMockMvc()
+        .perform(put(controllerBase + "/" + id).contentType(MediaType.APPLICATION_JSON).content(makeJson(dto)))
+        .andExpect(status().isOk());
+
+    assertNotNull(currentSession().get(updateType, id));
+    return currentSession().get(updateType, id);
+  }
+
+  protected void abstractSearch(String url, String searchTerm, int expectedSize) throws Exception {
```
callunity: I think this method name is a bit too abstract! (sorry.) But when I'm looking at it being called in a test class, it's not obvious what it's actually doing, and it's not obvious that it contains assertions. Perhaps something like `searchByTermAndAssertExpectedResultsCount`?

Same with `abstractSave` and `abstractUpdate` - please make the method names clearer about what they do, and consider appending `ReturningItem` so it's clear that further assertions can be called up it if desired.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -242,7 +244,37 @@ protected <T> void abstractTestDeleteFail(Class<T> deleteType, int id, String co
...
+        .andExpect(jsonPath("$").exists())
+        .andReturn();
 
+    Integer id = JsonPath.read(result.getResponse().getContentAsString(), "$.id");
+
+    assertNotNull(currentSession().get(controllerClass, id));
+    return currentSession().get(controllerClass, id);
+  }
+
+  protected <T, D> T abstractUpdate(String controllerBase, D dto, int id, Class<T> updateType) throws Exception {
```
callunity: Is there a difference between `updateType` here and the `controllerClass` argument in `abstractSave`? If not, please harmonize the names.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -258,6 +290,26 @@ protected ResultActions performDtRequest(String url, int displayLength, String d
         .andExpect(jsonPath("$").exists());
   }
 
+  // version of the dt response method with default values, because java doesn't support default
+  // method param values
```
callunity: To get around this and to make class this more flexible, you could create a second `performDtRequest` function that takes the args `(String url, int displayLength, String dataProp, int sortCol)` ,; this method could call that overloaded method, and then you'd have the more flexible method in case it is useful later.

PrabhjotChopra: Sorry, I'm not sure that I understand what you mean. Would this second `performDtRequest` method call the overloaded "default" dt response method?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -242,7 +244,40 @@ protected <T> void abstractTestDeleteFail(Class<T> deleteType, int id, String co
     // this user doesn't have permissions to delete things
   }
 
+  protected <T, D> T abstractCreateAndReturnEntity(String controllerBase, D dto, Class<T> controllerClass, int status)
```
avarsava: Sorry for not noticing this on your first PR, but "abstract" has a particular meaning in Java, which doesn't match how you're using it here. What's the pattern you're using for deciding whether to call a method 'abstract' or not? AbstractIT, AbstractDAOTest, etc don't name their methods this way

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
avarsava: Also not sure what Heather means because a performDtRequest with that signature is at line 282: but you should call it here. Currently you're implementing the logic of performDtRequest in two different places and you should want to minimize repeat code as much as possible. Your comment says "java doesn't support default method param values" - the way one does this in Java is immediately call the more specific method. So this method would look like:
`protected ResultActions performDtRequest(String url) throws Exception {
   return performDtRequest(url, 25, "id", 3);
 }`

PrabhjotChopra:
ignore the previous few commits while I try and debug some issues with my ability to run the db migrations for MISO

avarsava (approved):

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayRunRestControllerST.java
callunity: I see you built the DTO! Can you remove this comment too?

PR #3075, opened 2025-07-11
[GLT-4435] added assay rest tests

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4435

note: can only be reviewed after https://github.com/miso-lims/miso-lims/pull/3074 is merged


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -140,19 +139,17 @@ protected MockMvc getMockMvc() {
...
-          getMockMvc().perform(get(url)).andReturn().getResponse().getContentAsString();
+      response = getMockMvc().perform(get(url)).andReturn().getResponse().getContentAsString();
       status = JsonPath.read(response, "$.status");
       Thread.sleep(1000);
     }
     return response;
   }
 
-  protected <T> void abstractTestBulkCreateAsync(String controllerBase, Class<T> createType, List<?> dtos)
+  protected <T> void baseTestBulkCreateAsync(String controllerBase, Class<T> createType, List<?> dtos)
```
djcooke: Make this return the created objects, similar to `baseTestBulkUpdateAsync`. You should always check one or more attributes of the created objects to ensure they match what was submitted; otherwise, you're assuming that the IDs being returned from the endpoint you're testing are correct.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -231,7 +227,7 @@ protected <T> void abstractTestDelete(Class<T> deleteType, int id, String contro
     assertNull(currentSession().get(deleteType, id));
   }
 
-  protected <T> void abstractTestDeleteFail(Class<T> deleteType, int id, String controllerBase) throws Exception {
+  protected <T> void baseTestDeleteFail(Class<T> deleteType, int id, String controllerBase) throws Exception {
```
djcooke: Rename `testDeleteUnauthorized`. There are other reasons a deletion could fail - usually that the item is linked to something else - which this does not cover (and shouldn't)

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -163,14 +160,14 @@ protected <T> void abstractTestBulkCreateAsync(String controllerBase, Class<T> c
     }
   }
 
-  protected <T> void abstractBulkCreateAsyncFail(String controllerBase, Class<T> createType, List<?> dtos)
+  protected <T> void baseBulkCreateAsyncFail(String controllerBase, Class<T> createType, List<?> dtos)
```
djcooke: * In testing the operation failure, you should ensure that it's failing for the reason you expect - at least the status. The failure could be due to authorization (HTTP 401) or bad data submitted (HTTP 400). The comments below suggest that this method is checking for failures due to authorization (admin required), but that's not actually being checked. Should also check the status of each item returned - not just the second.
* Rename `testBulkCreateAsyncUnauthorized`

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -217,8 +214,7 @@ private String[] pollingResponserHelper(String requestType, List<?> dtos, String
     return new String[] {response, status}; // lets the caller choose what they want to use
   }
 
-
-  protected <T> void abstractTestDelete(Class<T> deleteType, int id, String controllerBase) throws Exception {
+  protected <T> void baseTestDelete(Class<T> deleteType, int id, String controllerBase) throws Exception {
```
djcooke: This completes the operation, performs all necessary assertions, and returns nothing, so it's the complete test - not just a "base." It also specifically tests "bulk-delete" endpoints. Rename `testBulkDelete`

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -242,7 +238,53 @@ protected <T> void abstractTestDeleteFail(Class<T> deleteType, int id, String co
     // this user doesn't have permissions to delete things
   }
 
+  protected <T, D> T baseCreateAndReturnEntity(String controllerBase, D dto, Class<T> controllerClass, int status)
```
djcooke: The method has a return type; don't need `AndReturnEntity` in the method name. Rename `baseTestCreate`. Same for `baseTestUpdate` below

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -242,7 +238,53 @@ protected <T> void abstractTestDeleteFail(Class<T> deleteType, int id, String co
...
+        .andExpect(jsonPath("$").exists())
+        .andReturn();
 
+    Integer id = JsonPath.read(result.getResponse().getContentAsString(), "$.id");
+
+    assertNotNull(currentSession().get(controllerClass, id));
+    return currentSession().get(controllerClass, id);
+  }
+
+  protected <T, D> void baseCreateFail(String controllerBase, D dto, Class<T> controllerClass)
```
djcooke: Rename `testCreateUnauthorized`

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -242,7 +238,53 @@ protected <T> void abstractTestDeleteFail(Class<T> deleteType, int id, String co
...
+    MvcResult result = getMockMvc()
+        .perform(post(controllerBase).contentType(MediaType.APPLICATION_JSON).content(makeJson(dto)))
+        .andExpect(status().is(status)) // either created (201) or OK (200) dependening on the controller
+        .andExpect(jsonPath("$").exists())
+        .andReturn();
 
+    Integer id = JsonPath.read(result.getResponse().getContentAsString(), "$.id");
+
+    assertNotNull(currentSession().get(controllerClass, id));
+    return currentSession().get(controllerClass, id);
```
djcooke: Retrieve the object only once

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -258,6 +300,11 @@ protected ResultActions performDtRequest(String url, int displayLength, String d
         .andExpect(jsonPath("$").exists());
   }
 
-
-
+  // version of the dt response method with default values, because java doesn't
+  // support default
+  // method param values
```
djcooke: This is called method overloading. Don't need comments to explain standard Java convention

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -242,7 +238,53 @@ protected <T> void abstractTestDeleteFail(Class<T> deleteType, int id, String co
...
+    return currentSession().get(controllerClass, id);
+  }
+
+  protected <T, D> void baseUpdateFail(String controllerBase, D dto, int id, Class<T> updateType) throws Exception {
+    getMockMvc()
+        .perform(put(controllerBase + "/" + id).contentType(MediaType.APPLICATION_JSON).content(makeJson(dto)))
+        .andExpect(status().isUnauthorized());
+  }
+
+  protected void baseSearchByTermWithExpectedNumResults(String url, String searchTerm, int expectedSize)
```
djcooke: * Should be checking that the correct objects are returned - not just the correct number. The best way to do that is probably by taking a `Collection<Long> ids` parameter here and checking that there is an object with each ID
* Don't need `WithExpectedNumResults` in the method name - that should be clear from the parameters

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -242,7 +238,53 @@ protected <T> void abstractTestDeleteFail(Class<T> deleteType, int id, String co
...
+      throws Exception {
+    getMockMvc()
+        .perform(put(controllerBase + "/" + id).contentType(MediaType.APPLICATION_JSON).content(makeJson(dto)))
+        .andExpect(status().isOk());
+
+    assertNotNull(currentSession().get(controllerClass, id));
+    return currentSession().get(controllerClass, id);
+  }
+
+  protected <T, D> void baseUpdateFail(String controllerBase, D dto, int id, Class<T> updateType) throws Exception {
```
djcooke: `testUpdateUnauthorized`

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayRunRestControllerST.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.List;
+import java.util.Arrays;
+import java.util.ArrayList;
+
+import org.springframework.test.web.servlet.MockMvc;
+import java.util.Date;
+import java.time.LocalDate;
+
+
+public class ArrayRunRestControllerST extends AbstractST {
```
djcooke: Isn't this PR for testing `AssayRestController`?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayRunRestControllerST.java
```diff
@@ -0,0 +1,132 @@
...
+
+
+  @Test
+  public void testDtResponse() throws Exception {
+    performDtRequest(CONTROLLER_BASE + "/dt")
+        .andExpect(jsonPath("$.iTotalRecords").value(3))
+        .andExpect(jsonPath("$.aaData[0].alias").value("ArrayRun_1"))
+        .andExpect(jsonPath("$.aaData[0].status").value("Running"))
+        .andExpect(jsonPath("$.aaData[1].alias").value("to_delete"))
+        .andExpect(jsonPath("$.aaData[2].instrumentName").value("iScan1"));
```
djcooke: Check the same attributes for each item. Simplest would be to just check the IDs. Then you could create a method in `AbstractST` to simplify all of these similar tests

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayRunRestControllerST.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Test
+  public void testSave() throws Exception {
+    // "health" is a non-nullable field that does not have a setter in the dto, so I need to use the
+    // actual ArrayRun type
```
djcooke: Looks like you figured out to use the status field on the DTO. Remove this comment

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayRunRestControllerST.java
PrabhjotChopra: yes, but it is built off the code from the Array Run PR

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
PrabhjotChopra: When using this method in my controller tests, I make a helper method to create the DTOs, and then use those same DTOs for both the create and the create fail tests. Since the only difference between the tests is the user that does the action (same data and everything), it's safe to assume that the authorization was the issue.

Also I think I kept getting Runtime exceptions when I was trying to get unauthorized to be thrown, and ended up using this solution.

djcooke: The response status should be 401 and that should also be included in the JSON error response. If either of those aren't happening, that's a bug. Please assert both of these things.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -271,19 +284,31 @@ protected <T, D> T baseUpdateAndReturnEntity(String controllerBase, D dto, int i
...
   }
 
-  protected void baseSearchByTermWithExpectedNumResults(String url, String searchTerm, int expectedSize)
+  protected MultiValueMap searchTerm(String term) {
+    MultiValueMap<String, String> map = new LinkedMultiValueMap();
+    map.add("q", term);
+    return map;
+  }
+
+  protected void baseSearchByTerm(String url, MultiValueMap params, int expectedSize, List<Integer> ids)
```
djcooke: Remove `expectedSize` parameter and use `ids.size()` instead

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -271,19 +284,31 @@ protected <T, D> T baseUpdateAndReturnEntity(String controllerBase, D dto, int i
...
       throws Exception {
-    getMockMvc().perform(get(url).param("q", searchTerm).accept(MediaType.APPLICATION_JSON))
+    String response = getMockMvc().perform(get(url).params(params).accept(MediaType.APPLICATION_JSON))
```
djcooke: What was the reason for changing the parameter to a `MultiValueMap` instead of `String searchTerm`? Combined with the `searchTerm` method above, it's doing the same thing, but was simpler before.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -271,19 +284,31 @@ protected <T, D> T baseUpdateAndReturnEntity(String controllerBase, D dto, int i
...
         .andExpect(status().isOk())
         .andExpect(jsonPath("$").exists())
         .andExpect(content().contentType(MediaType.APPLICATION_JSON))
-        .andExpect(jsonPath("$.*", hasSize(expectedSize)));
+        .andExpect(jsonPath("$.*", hasSize(expectedSize)))
+        .andReturn().getResponse().getContentAsString();
+
+    for (int i = 0; i < ids.size(); i++) { // checks that the ids found are the right ones
+      assertTrue(ids.contains(JsonPath.read(response, "$[" + i + "].id")));
```
djcooke: You're checking that `ids` (the expected IDs) contains each returned ID. It should be the other way around

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -298,13 +323,37 @@ protected ResultActions performDtRequest(String url, int displayLength, String d
...
-  // version of the dt response method with default values, because java doesn't
-  // support default
-  // method param values
   protected ResultActions performDtRequest(String url)
       throws Exception {
     return performDtRequest(url, 25, "id", 3);
   }
+
+  // checks ids for datatable endpoints and "list all" endpoints
+  protected ResultActions checkIds(ResultActions r, List<Integer> ids, boolean dt) throws Exception {
```
djcooke: Is there a reason not to call this explicitly in `performDtRequest` (and probably rename to `testDtRequest`)? I think you'd always want to verify the results when calling `performDtRequest`

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
djcooke: I looked into this some more. Authorization exceptions aren't handled well in the async interface. 401 is not returned, which actually makes makes because the request is for the operation status. Even when reporting failure, the status is returned successfully. When an async save fails due to validation, the response JSON includes something like this:

```
detail: "Validation failed"
data: [{
  row: #,
  fields: [{
    field: "name",
	errors: ["message"]
  }]
}]
```

When the save fails due to authorization exception, there is no `data` field, and the `detail` message is "An unexpected error has occurred". I don't think there's a need to improve this though. Generally the front-end doesn't offer you options that you're not authorized to do, so this probably won't come up unless you've done something weird.

For now, please just assert the above - no `data` field, and `detail` is "An unexpected error has occurred".

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
PrabhjotChopra: some search endpoints require more params than just a search term. By changing the parameter to a `MultiValueMap` I can test multiple parameters, and with the `searchTerm` helper method it's not too cumbersome for regular search endpoint testing.

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.html#params(org.springframework.util.MultiValueMap)

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
PrabhjotChopra: My reasoning is, if the sizes are the same ("iTotalRecords"), and every returned ID is present in the expected IDs list, then the two lists (returned IDs and expected IDs) should be the same.

djcooke: What if there are duplicates in the returned IDs?

PrabhjotChopra: good point, I will fix that

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
djcooke: If the `searchTerm` method is only meant to be used when there is a single parameter, you could simplify usage by overloading the `baseSearchByTerm` method:

```
protected void baseSearchByTerm(String url, String searchTerm, List<Integer> ids) {
  baseSearchByTerm(String url, searchTerm(searchTerm), List<Integer> ids);
}
```

The `searchTerm` method can then be private, and should also be static.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -263,18 +287,38 @@ protected <T, D> T baseUpdateAndReturnEntity(String controllerBase, D dto, int i
...
+
+    for (int i = 0; i < ids.size(); i++) { // checks that the ids found are the right ones
+      assertTrue(ids.contains(JsonPath.read(response, "$[" + i + "].id")));
+    }
+
   }
 
-  protected ResultActions performDtRequest(String url, int displayLength, String dataProp, int sortCol)
+  protected ResultActions testDtRequest(String url, int displayLength, String dataProp, int sortCol, List<Integer> ids,
+      boolean dt)
```
djcooke: What is `dt`? Isn't this method always for testing dt (DataTables) data?

PrabhjotChopra: There are some "list-all" kinds of endpoints where I can reuse almost all the code from this method, and just need to modify the path to access the ids a bit.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -327,19 +324,34 @@ protected ResultActions testDtRequest(String url, int displayLength, String data
...
-    for (int i = 0; i < ids.size(); i++) { // checks that the ids found are the right ones
-      assertTrue(ids.contains(JsonPath.read(response, "$" + dtPath + "[" + i + "].id")));
+    for (int i = 0; i < expectedIds.size(); i++) { // checks that the ids found are the right ones
+      Integer returned = JsonPath.read(response, "$" + dtPath + "[" + i + "].id");
+      assertTrue(expectedIds.contains(returned));
+
+      expectedIds.set(expectedIds.indexOf(returned), -1);
+      // Arrays.asList is primarily used to make the expected ids list, but it returns a fixed size array
+      // to work around this, this line sets that id to an invalid id, so if a duplicate is found, the
+      // method will throw on the assertion in the previous line
```
djcooke: Now what if something is returned with an invalid ID of -1? 🙃

PrabhjotChopra: Can we assume that only saved objects can be returned? If so, would it work to set that to 0 instead of -1? (since 0 is only ever used as an unsaved id)

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
djcooke: A `testDtRequest` method should only test DT requests, and this method is adding parameters that only make sense for DT requests. It sounds like "list-all" methods should be tested separately.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
djcooke: The point of testing is to make sure something is behaving correctly, so you can't assume it's not doing something weird. You should go through the expected IDs and ensure that each one appears in the returned objects.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -283,14 +328,57 @@ protected ResultActions performDtRequest(String url, int displayLength, String d
...
+    try {
+      // this should throw
+      JsonPath.read(response, "$" + dtPath + "[" + expectedIds.size() + "].id");
+      // this accounts for the edge case of a non-datatable id-checking having more returned ids than
+      // expected ids
+      // also accounts for duplicates that are higher in index then the length of the expected ids list
+
+      assertTrue(false); // if the line above does not throw, this will trigger
+    } catch (Exception e) {
+    }
```
djcooke: This can be much simpler and straight-forward. Just collect the returned IDs into a list and then check the size and all expected IDs. I haven't tested this, but looking at the [JsonPath docs](https://github.com/json-path/JsonPath/blob/master/README.md), something like this should work:

```
List<Integer> resultIds = JsonPath.read(response, "$" + dtPath + "[*].id");
assertEquals(expectedIds.size(), resultIds.size());
for (Integer expectedId : expectedIds) {
  assertTrue(resultIds.contains(expectedId);
}
```

avarsava (approved):

PR #3077, opened 2025-07-11
[GLT-4439] added barcodable search test

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4439

Note: https://github.com/miso-lims/miso-lims/pull/3074 must be merged before this can be reviewed


avarsava (approved):

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/BarcodableSearchRestControllerST.java
```diff
@@ -0,0 +1,14 @@
...
+import org.junit.Test;
+import java.util.Arrays;
+
+public class BarcodableSearchRestControllerST extends AbstractST {
+
+  private static final String CONTROLLER_BASE = "/rest/barcodables";
+
+  @Test
+  public void testSearch() throws Exception {
+    baseSearchByTerm(CONTROLLER_BASE + "/search", "11111", Arrays.asList(1));
```
djcooke: This returns `BarcodableDtos` which can be created from several different entity types, so checking the ID is not enough to verify the results. You should check both the ID and the `entityType` field.

djcooke (commented):
Please add a few tests to cover the different entity types that this should be able to find - sample, box, pool, library, library aliquot, and container.

PR #3083, opened 2025-07-17
[GLT-4445] added contact role rest tests

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4445

Added contact role rest tests

### DIFF
--- a/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/InstrumentStatusRestControllerST.java
+++ b/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/InstrumentStatusRestControllerST.java
@@ -55,7 +55,7 @@ public class InstrumentStatusRestControllerST extends AbstractST {
     getMockMvc().perform(get(CONTROLLER_BASE))
         .andExpect(status().isOk())
         .andExpect(jsonPath("$.*", hasSize(8)))
-        .andExpect(jsonPath("$[0].instrument.id").value(2))
+        .andExpect(jsonPath("$[0].instrument.id").value(5002))
         .andExpect(jsonPath("$[0].positions[0].run.type").value("Base"))
         .andExpect(jsonPath("$[1].instrument.id").value(2))
         .andExpect(jsonPath("$[2].instrument.id").value(3))


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ContactRoleRestControllerST.java
```diff
@@ -0,0 +1,129 @@
...
+    List<ContactRoleDto> dtos = new ArrayList<ContactRoleDto>();
+    dtos.add(Dtos.asDto(con2));
+    dtos.add(Dtos.asDto(con3));
+
+    List<ContactRole> contactRoles =
+        (List<ContactRole>) baseTestBulkUpdateAsync(CONTROLLER_BASE, controllerClass, dtos,
+            Arrays.asList(2, 3));
+
+    assertEquals("con2", contactRoles.get(0).getName());
+    assertEquals("con3", contactRoles.get(1).getName());
```
djcooke, line 101: assert IDs for order

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ContactRoleRestControllerST.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.ArrayList;
+
+import org.springframework.test.web.servlet.MockMvc;
+import java.util.Date;
+
+
+public class ContactRoleRestControllerST extends AbstractST {
+
+  private static final String CONTROLLER_BASE = "/rest/contactroles";
+  private static final Class<ContactRole> controllerClass = ContactRole.class;
```
djcooke: `entityClass`

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ContactRoleRestControllerST.java
```diff
@@ -74,15 +74,15 @@ public void testBulkCreateFail() throws Exception {
...
   @Test
   @WithMockUser(username = "admin", password = "admin", roles = {"INTERNAL", "ADMIN"})
   public void testBulkUpdateAsync() throws Exception {
     // contact roles can only be updated by an admin user
-    ContactRole con2 = currentSession().get(controllerClass, 2);
-    ContactRole con3 = currentSession().get(controllerClass, 3);
+    ContactRole con2 = currentSession().get(entityClass, 2);
+    ContactRole con3 = currentSession().get(entityClass, 3);
     con2.setName("con2");
     con3.setName("con3");
```
djcooke: Didn't notice this in other PRs, but you should be modifying the fields on the DTOs - not on the entity classes. This is very important. As is, you're updating the entities within the session, so these changes could be persisted before (and without) any interaction with the controller.

Please check other merged tests and correct if necessary.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -76,6 +79,13 @@ public abstract class AbstractST {
   private static final String CLEAR_DATA_SCRIPT = "clear_test_data.sql";
   private static final String DETAILED_SCRIPT = "integration_test_data.sql";
 
+  private static final boolean DEBUG_MODE;
+
+  static {
+    DEBUG_MODE = Boolean.parseBoolean(System.getProperty("debug", "false"));
```
djcooke: change property name to "st.debug" or something similar so it's clear that this is only for the tests

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -35,10 +35,13 @@
 import org.springframework.test.web.servlet.MvcResult;
 import org.springframework.util.MultiValueMap;
 import org.hibernate.Session;
+import org.hibernate.engine.jdbc.env.spi.IdentifierCaseStrategy;
```
djcooke: remove unused imports (this and others)

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -185,6 +195,9 @@ protected <T> List<T> baseTestBulkUpdateAsync(String controllerBase, Class<T> up
     for (int id : ids) {
       T obj = currentSession().get(updateType, id);
       assertNotNull(obj);
+      // asserts the id to ensure the order of object ids is the same order as the list of ids given
+      // this removes the need for id checking in update
+      assertEquals((long) id, ((Identifiable) obj).getId());
```
djcooke: Hmm... As is, it's actually pointless to check the ID order because you're looking them up in the order you've specified yourself. You're also completely ignoring the results returned by the controller.

* Remove `ids` parameter
* Change `dtos` type to `<D>` and add parameter to get IDs: `Function<D, Long> getId`
* Construct `ids` here: `List<Long> ids = dtos.stream().map(getId).toList()`
* Confirm expected order
* Ideally this method would return the DTOs as they were returned by the controller and you would check the changed values on the DTOs. Checking the values saved in the database is also useful though, so I think you can just leave that part as is. With the above updates, you're at least ensuring that the controller is returning the correct items, and in the correct order.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -201,26 +214,32 @@ protected <T> void testBulkUpdateAsyncUnauthorized(String controllerBase, Class<
...
-            .perform(post(controllerBase + "/bulk").contentType(MediaType.APPLICATION_JSON).content(makeJson(dtos)))
-            .andExpect(status().isAccepted())
-            .andReturn();
+        ac = getMockMvc()
+            .perform(post(controllerBase + "/bulk").contentType(MediaType.APPLICATION_JSON).content(makeJson(dtos)));
+        if (DEBUG_MODE)
+          ac.andDo(print());
+
+        mvcResult = ac.andExpect(status().isAccepted()).andReturn();
         break;
```
djcooke: Throw an exception instead of defaulting to post if something invalid is specified. Alternately, change the parameter to a `putMethod` or similar boolean where true == PUT and false == POST

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -201,26 +214,32 @@ protected <T> void testBulkUpdateAsyncUnauthorized(String controllerBase, Class<
...
-        mvcResult = getMockMvc()
-            .perform(post(controllerBase + "/bulk").contentType(MediaType.APPLICATION_JSON).content(makeJson(dtos)))
-            .andExpect(status().isAccepted())
-            .andReturn();
+        ac = getMockMvc()
+            .perform(post(controllerBase + "/bulk").contentType(MediaType.APPLICATION_JSON).content(makeJson(dtos)));
+        if (DEBUG_MODE)
+          ac.andDo(print());
+
+        mvcResult = ac.andExpect(status().isAccepted()).andReturn();
```
djcooke: Move everything after `post(...)` outside of the `switch` so you don't need to repeat it

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -272,27 +302,39 @@ protected <T, D> T baseTestCreate(String controllerBase, D dto, Class<T> control
...
+
+    ac.andExpect(status().isOk());
+
+    T obj = currentSession().get(controllerClass, id);
+
+    assertNotNull(obj);
+    assertEquals(id, (int) ((Identifiable) obj).getId());
 
-    assertNotNull(currentSession().get(controllerClass, id));
     return currentSession().get(controllerClass, id);
```
djcooke: Retrieving the same object twice

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
PrabhjotChopra: I'm not sure that I understand how this gets the results from the controller. The async endpoints don't really return any useful information back, e.g. `Body = {"operationId":"47b60580-d188-4f13-9d8d-2d7aeb5c3057","status":"running","completedUnits":0,"totalUnits":2}`, so I'm not sure how I could get anything from the controller itself. Since I can't get any actual dtos back from the controller here, I just test it via saved values in the db.

djcooke: Your `pollingResponse` method polls the controller until the operation is no longer "running," but then does nothing with the response. When complete, the saved DTOs are included in the response. it should also be confirmed that the operation succeeded (or failed, as appropriate).

PrabhjotChopra: Oh ok, I see what you mean, did not realize that the last response contains the data. 
If the intention is to verify that the returned/response DTOs are in the proper order, verified by checking their IDs, would the following snippet be acceptable?
```
String response = pollingResponserHelper("put", dtos, controllerBase);

    for (int i = 0; i < ids.size(); i++) {
      assertEquals(ids.get(i), JsonPath.read(response, "$.data[" + i + "].id"));
    }
```
With the assumption being that the ids are passed as parameters and in the correct order (which I believe that I have done in all tests so far).

djcooke: Almost. Again, there's no need for the `ids` parameter because you can pull the IDs out of the DTOs.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -175,19 +176,23 @@ protected <T> void testBulkCreateAsyncUnauthorized(String controllerBase, Class<
...
+      assertTrue("expected: " + ids.get(i) + "\nactual: " + JsonPath.read(response, "$.data[" + i + "].id"),
+          (int) (long) ids.get(i) == (int) JsonPath.read(response, "$.data[" + i + "].id"));
+    }
 
-    // now check if the updates went through
     List<T> objects = new ArrayList<T>();
-    for (int id : ids) {
-      T obj = currentSession().get(updateType, id);
+    for (long id : ids) {
+      T obj = currentSession().get(updateType, (int) id);
```
djcooke: Don't cast to int. `Long` is the correct type here (and the line above).

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -175,19 +176,23 @@ protected <T> void testBulkCreateAsyncUnauthorized(String controllerBase, Class<
...
   }
 
-  protected <T> List<T> baseTestBulkUpdateAsync(String controllerBase, Class<T> updateType, List<?> dtos,
-      List<Integer> ids)
+  protected <T, D> List<T> baseTestBulkUpdateAsync(String controllerBase, Class<T> updateType, List<D> dtos,
+      Function<D, Long> getId)
       throws Exception {
-    pollingResponserHelper("put", dtos, controllerBase);
+    String response = pollingResponserHelper("put", dtos, controllerBase);
+    System.out.println(response);
```
djcooke: remove

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -175,19 +176,23 @@ protected <T> void testBulkCreateAsyncUnauthorized(String controllerBase, Class<
...
+
+    // check order of returned IDs
+    List<Long> ids = dtos.stream().map(getId).toList();
+    for (int i = 0; i < ids.size(); i++) {
+      assertTrue("expected: " + ids.get(i) + "\nactual: " + JsonPath.read(response, "$.data[" + i + "].id"),
```
djcooke: Use `assertEquals`. It will generate the expected vs. actual message for you.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -175,19 +176,23 @@ protected <T> void testBulkCreateAsyncUnauthorized(String controllerBase, Class<
...
+          (int) (long) ids.get(i) == (int) JsonPath.read(response, "$.data[" + i + "].id"));
```
djcooke: Avoid all this casting. These should both be `Long`. If JsonPath.read returns an int or Integer, use `Long.valueOf` to convert it.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -175,19 +176,23 @@ protected <T> void testBulkCreateAsyncUnauthorized(String controllerBase, Class<
     // request should fail without admin permissions
   }
 
-  protected <T> List<T> baseTestBulkUpdateAsync(String controllerBase, Class<T> updateType, List<?> dtos,
-      List<Integer> ids)
+  protected <T, D> List<T> baseTestBulkUpdateAsync(String controllerBase, Class<T> updateType, List<D> dtos,
```
djcooke, line 183: Add JavaDoc that explains dtos must be provided in ID order

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -176,22 +176,24 @@ protected <T> void testBulkCreateAsyncUnauthorized(String controllerBase, Class<
...
       throws Exception {
     String response = pollingResponserHelper("put", dtos, controllerBase);
-    System.out.println(response);
 
     // check order of returned IDs
     List<Long> ids = dtos.stream().map(getId).toList();
     for (int i = 0; i < ids.size(); i++) {
-      assertTrue("expected: " + ids.get(i) + "\nactual: " + JsonPath.read(response, "$.data[" + i + "].id"),
-          (int) (long) ids.get(i) == (int) JsonPath.read(response, "$.data[" + i + "].id"));
+      assertEquals(ids.get(i).longValue(), Integer.toUnsignedLong(JsonPath.read(response, "$.data[" + i + "].id")));
```
djcooke: `toUnsignedLong` will change negative values. Negative IDs shouldn't exist, but we are testing, so should not assume. `JsonPath.read` returns an `Integer` so you can use `.longValue()` to get a `long` from it.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -176,22 +176,24 @@ protected <T> void testBulkCreateAsyncUnauthorized(String controllerBase, Class<
...
     }
 
     List<T> objects = new ArrayList<T>();
     for (long id : ids) {
```
djcooke: `Long`

djcooke: Bump. It makes little difference here, but autoboxing/unboxing should be avoided wherever possible. It's coming from a `List<Long>` and `Long` will work where it's used below, so keep it `Long`

PrabhjotChopra: Oh ok, will change that

PrabhjotChopra: My reasoning for unboxing here is that, when I don't unbox the left side, I get a `reference to assertEquals is ambiguous
[ERROR]   both method assertEquals(long,long) in org.junit.Assert and method assertEquals(java.lang.Object,java.lang.Object) in org.junit.Assert match`

Fixing this would then require a cast to `Long` on the right side, which doesn't seem ideal (although I can change it to that if that works better).

djcooke: I think you're thinking of above? This `id` is not used in any `assertEquals` call

PrabhjotChopra: oh the for loop one at line 195, yes I will fix that now

djcooke (approved):

PR #3092, opened 2025-07-21
[GLT-4455] added instrument model rest tests

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4455
 
#3074 must be merged first


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/InstrumentModelRestControllerST.java
```diff
@@ -0,0 +1,120 @@
...
+
+import org.junit.Test;
+
+import org.springframework.web.servlet.*;
+
+import static org.springframework.test.web.servlet.setup.MockMvcBuilders.*;
+
+import javax.ws.rs.core.MediaType;
+
+import org.checkerframework.checker.units.qual.Temperature;
```
djcooke: Remove unused imports

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/InstrumentModelRestControllerST.java
```diff
@@ -0,0 +1,120 @@
...
+
+import org.springframework.test.web.servlet.MockMvc;
+import java.util.Date;
+import java.time.LocalDate;
+
+
+public class InstrumentModelRestControllerST extends AbstractST {
+
+  private static final String CONTROLLER_BASE = "/rest/instrumentmodels";
+  private static final Class<InstrumentModel> controllerClass = InstrumentModel.class;
```
djcooke: Entity class

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/InstrumentModelRestControllerST.java
```diff
@@ -0,0 +1,120 @@
...
+    dto.setDataManglingPolicy("NONE");
+    return dto;
+  }
+
+  @Test
+  @WithMockUser(username = "admin", password = "admin", roles = {"INTERNAL", "ADMIN"})
+  public void testCreate() throws Exception {
+    // must be admin to create instrument model
+    InstrumentModel newModel = baseTestCreate(CONTROLLER_BASE, makeCreateDto(), controllerClass, 200);
+    assertEquals("testmodel", newModel.getAlias());
```
djcooke, line 59: check all fields

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/InstrumentModelRestControllerST.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+
+  @Test
+  @WithMockUser(username = "admin", password = "admin", roles = {"INTERNAL", "ADMIN"})
+  public void testUpdate() throws Exception {
+    // must be admin to update instrument model
+
+    InstrumentModel model = currentSession().get(controllerClass, 1);
+    model.setAlias("updated");
```
djcooke, line 80: Modify DTO - not entity class

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/InstrumentModelRestControllerST.java
```diff
@@ -0,0 +1,120 @@
...
+    InstrumentModel updated = baseTestUpdate(CONTROLLER_BASE, Dtos.asDto(model), 1, controllerClass);
+    assertEquals("updated", updated.getAlias());
+  }
+
+  @Test
+  public void testUpdateFail() throws Exception {
+    // must be admin to update instrument model
+
+    InstrumentModel model = currentSession().get(controllerClass, 1);
+    model.setAlias("updated");
```
djcooke, line 90: Modify DTO

djcooke (approved):

avarsava (approved):

miso-web/src/it/resources/db/migration/clear_test_data.sql
```diff
@@ -140,6 +140,7 @@ DELETE FROM WorksetCategory;
 DELETE FROM WorksetStage;
 DELETE FROM Deliverable;
 DELETE FROM ContactRole;
+DELETE FROM SequencingContainerModel;
```
djcooke: This table is already cleared on L39

miso-web/src/it/resources/db/migration/integration_test_data.sql
```diff
@@ -1038,7 +1038,7 @@ INSERT INTO Run (runId, name, alias, instrumentId, startDate, completionDate, he
 (2, 'RUN2', 'PacBio_Run_1', 3, '2017-08-01', NULL, 'Running', 3, '2017-08-01 10:03:02', 3, '2017-08-01 10:03:02');
 
 INSERT INTO Run (runId, name, alias, instrumentId, sequencingParameters_parametersId, description, filePath, startDate, completionDate, health, creator, created, lastModifier, lastModified) VALUES
-(5001, 'RUN5001', 'Change_Values_Run', 5002, 2, 'description', '/filePath', '2017-09-05', NULL, 'Running', 1, '2017-09-05 11:00:00', 1, '2017-09-05 11:00:00'),
+(5001, 'RUN5001', 'Change_Values_Run', 5002, 2, 'description', '/filePath', '2017-05-05', NULL, 'Running', 1, '2017-09-05 11:00:00', 1, '2017-09-05 11:00:00'),
```
djcooke: Did you change this so that the sorting for a particular test was consistent? Instead, either

1. make the sorting consistent by adding a secondary sort by ID, or
2. don't depend on sort order in the test - find the object you want by ID and then test its attributes

PrabhjotChopra: The instrument status sorting is done by 1. whether the instrument has "ALL", "SOME", or "NONE" of its positions running, then 2. by Run completion date, or by start date if completion date is not available.

In this case, both instruments had all running positions I think, and their start dates were exactly the same, with neither having completion dates. Given this, I don't think that there was any code to handle the explicit sorting in this case, so I don't really know how it was sorting those two entries before (or just randomizing it). Changing the start date here made the sorting order explicit, and the rest of the values have consistent sorting order from what I can see.

djcooke: This change to the test data is avoiding the problem instead of solving it. If the test data is changed in the future, the next person will experience the same confusion that you did. Making the sorting consistent is the best solution, and can be done by adding another level of sorting using the instrument ID or name.

PrabhjotChopra: Sure, will do

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/InstrumentStatusRestController.java
```diff
@@ -43,15 +41,16 @@ public List<InstrumentStatusDto> listAll() throws IOException {
...
+          long idA = a.getId();
+          long idB = b.getId();
+          if (idA < idB) {
             return 1;
+          } else if (idA == idB) {
+            return 0;
+          } else {
+            return -1;
           }
-          return bDate.compareTo(aDate);
```
djcooke: * ID should be another level of sorting _after_ sorting by date - it should not replace sorting by date
* Can use `Long.compare(long, long)`

miso-web/src/it/resources/db/migration/integration_test_data.sql
djcooke: Please change this back

PR #3096, opened 2025-07-22
[GLT-4640] fixed not null sql migrations

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4640


sqlstore/src/main/resources/db/migration/V1700__auto_main.sql
```diff
@@ -0,0 +1,3 @@
+-- update_url_field
```
djcooke: New migrations need to be added in the V8000-8999 range and should have a descriptive name. The "auto_main" migrations are generated from these during release

PR #3126, opened 2025-08-05
[GLT-4484] added requisition rest tests

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4484

### DIFF
--- a/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
+++ b/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
@@ -302,11 +302,11 @@ public abstract class AbstractST {
    * 
    * @param requestType either "put" for async update or "post" for async creation
    * @param dtos dtos to create
-   * @param url base URL for the async request
+   * @param baseUrl base URL for the async request
    * @return request response
    */
-  private String pollingResponserHelper(String requestType, List<?> dtos, String url) throws Exception {
-    return pollingResponserHelper(requestType, dtos, url + "/bulk", url, 202);
+  private String pollingResponserHelper(String requestType, List<?> dtos, String baseUrl) throws Exception {
+    return pollingResponserHelper(requestType, dtos, baseUrl + "/bulk", baseUrl + "/bulk", 202);
   }
 
   protected <T> void testBulkDelete(Class<T> deleteType, int id, String controllerBase) throws Exception {


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/RequisitionRestController.java
```diff
@@ -304,9 +304,25 @@ public record MoveItemsRequest(Long requisitionId, String requisitionAlias, Long
...
     public List<Long> requisitionIds;
     public String resumeDate;
+
+    public List<Long> getRequisitionIds() {
+      return requisitionIds;
+    }
+
+    public String resumeDate() {
+      return resumeDate;
+    }
```
djcooke: Both getters should start with `get` for consistency. Alternately (and better), you could change this class to a record.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -88,7 +89,7 @@ public abstract class AbstractST {
   private MockMvc mockMvc;
 
   @PersistenceContext
-  private EntityManager entityManager;
+  protected EntityManager entityManager;
```
djcooke: This should remain private. If it's needed in subclasses, you can add a getter for it, but I don't think it is.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -142,6 +143,7 @@ protected MockMvc getMockMvc() {
 
   protected String pollingResponse(String url) throws Exception {
     String response = getMockMvc().perform(get(url)).andReturn().getResponse().getContentAsString();
+    System.out.println(response);
```
djcooke: remove

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -231,13 +234,84 @@ private String pollingResponserHelper(String requestType, List<?> dtos, String c
     if (DEBUG_MODE)
       ac.andDo(print());
 
-    mvcResult = ac.andExpect(status().isAccepted()).andReturn();
+    mvcResult = ac.andExpect(status().is(isOneOf(200, 204))).andReturn();
```
djcooke: You should be testing the specific expected status of every request. If a request returns 200 when you're expecting 204, that's bad.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -205,24 +207,25 @@ protected <T, D> List<T> baseTestBulkUpdateAsync(String controllerBase, Class<T>
...
     // tests failure for async update endpoints where admin permissions are required
-    String response = pollingResponserHelper("put", dtos, controllerBase);
+    String response = pollingResponserHelper("put", dtos, controllerBase + "/bulk", controllerBase + "/bulk");
     assertEquals("An unexpected error has occurred", JsonPath.read(response, "$.detail"));
-    // request should fail without admin permissions }
+    // request should fail without admin permissions
   }
 
-  private String pollingResponserHelper(String requestType, List<?> dtos, String controllerBase) throws Exception {
+  private String pollingResponserHelper(String requestType, List<?> objs, String url, String pollResponse)
```
djcooke: * `dtos` was a clearer name than `objs` since it is a DTO that should be used here. Some request-specific objects may not have "Dto" in the class name, but they are classes created specifically for data transfer purposes.
* `url` _is_ a URL, and `pollResponse` is a partial URL - not a response at all. Give these better names so it's clear what they are and what they're used for.
* You usually call this with the same `url` and `pollResponse`. Overload this method with another that only takes one URL parameter and then calls this one, using it for both.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -231,13 +234,84 @@ private String pollingResponserHelper(String requestType, List<?> dtos, String c
...
-    mvcResult = ac.andExpect(status().isAccepted()).andReturn();
+    mvcResult = ac.andExpect(status().is(isOneOf(200, 204))).andReturn();
 
     String id = JsonPath.read(mvcResult.getResponse().getContentAsString(), "$.operationId");
-    String response = pollingResponse(controllerBase + "/bulk/" + id);
+    String response = pollingResponse(pollResponse + "/" + id);
     return response;
   }
 
+  protected <D> void testAddAsync(long referenceId, String url, List<Long> entityIds, Class<D> entityType,
```
djcooke: You've tried to make these testAdd, testRemove, and isContained methods generic, but they're really specific to requisitions and would be cleaner kept in the implementation rather than the abstract class

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/RequisitionRestControllerST.java
```diff
@@ -0,0 +1,279 @@
...
+  private static final HibernateRequisitionDao dao = new HibernateRequisitionDao();
+
+
+  @Test
+  public void testDatatable() throws Exception {
+    testDtRequest(CONTROLLER_BASE + "/dt", Arrays.asList(1, 2));
+  }
+
+  private <T> void testMove(int id, String type, MoveItemsRequest req, Class<T> targetType,
+      Function<T, Identifiable> getReq) throws Exception {
```
djcooke: Please call requisitions requisitions and requests requests. "req" is confusing when both are involved

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/RequisitionRestControllerST.java
```diff
@@ -0,0 +1,279 @@
...
+
+import org.springframework.test.web.servlet.MockMvc;
+
+
+
+public class RequisitionRestControllerST extends AbstractST {
+
+  private static final String CONTROLLER_BASE = "/rest/requisitions";
+  private static final Class<Requisition> entityClass = Requisition.class;
+  private static final HibernateRequisitionDao dao = new HibernateRequisitionDao();
```
djcooke: DAOs are part of what's being tested, so don't rely on DAOs to validate the results. Look up supplemental samples/libraries by querying `RequisitionSupplementalSample`/`Library`. These have composite IDs including the sample/library and the requisition ID. If the `RequisitionSupplemental...` entity exists, then the sample/library is included in the requisition's supplemental items; otherwise it is not.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -234,85 +234,20 @@ private String pollingResponserHelper(String requestType, List<?> objs, String u
...
-
-    String response = pollingResponserHelper("post", entityIds, url, pollUrl);
-    for (Long entityId : entityIds) {
-      assertTrue(isContained(referenceId, entityId, entityType, getParent));
-      // asserts the objects were indeed added
-    }
+  // overloaded method for cases where the polling response URL is the same as the url
+  // this also assume that the expected response code is 202, or HTTP ACCEPTED, as is standard for
+  // nearly all of the async endpoints
+  private String pollingResponserHelper(String requestType, List<?> dtos, String url) throws Exception {
```
djcooke: Use JavaDoc format for method-level comments

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/RequisitionRestControllerST.java
```diff
@@ -9,55 +9,107 @@
...
+
+  // template test for requsition sample and library testing
+  private <T> boolean isContained(long requisitionId, long entityId, Class<T> entityType,
+      Function<T, Identifiable> getRequisition) {
+    List<Identifiable> parent =
+        Stream.of(entityType.cast(currentSession().get(entityType, entityId))).map(getRequisition)
+            .filter(s -> s != null)
+            .toList();
+    boolean isNotContained = true; // true by default
+    if (!parent.isEmpty()) { // if the entity has a parent
```
djcooke: * Don't make a stream/list of 1 thing
* If the entity isn't found right now, it'll result in returning false (is not contained), which is technically not wrong, but hides the actual issue. Maybe throw an exception or just let it NPE since you should only be referencing entities you expect to exist
* `parent` is confusing because samples/libraries do have parent samples, but a requisition is not usually referred to as a sample/library's parent
* `isNotContained` is not necessary and returning its opposite creates a confusing double/triple negative - just return true if `entity.requisition.id == requisitionId`

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/RequisitionRestControllerST.java
```diff
@@ -9,55 +9,107 @@
...
+      if (parent.get(0).getId() == requisitionId) { // if the parent ids match
+        isNotContained = false;
+      } else {
+        // for debugging
+        System.out.println(
```
djcooke: remove

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/RequisitionRestControllerST.java
```diff
@@ -9,55 +9,107 @@
...
+      assertFalse(isContained(requisitionId, entityId, entityType, getRequisition));
+      // asserts the objects were indeed removed from that requisition
+      // as a result that entity should not have any requisition so a list of the
+      // requisitions should be empty
+    }
+  }
+
+  // template test for requsition sample and library testing
+  private <T> boolean isContained(long requisitionId, long entityId, Class<T> entityType,
+      Function<T, Identifiable> getRequisition) {
```
djcooke: `getRequisition` should be `Function<T, Requisition>` because you specifically need to get a requisition

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/RequisitionRestControllerST.java
```diff
@@ -9,55 +9,107 @@
...
     getMockMvc()
-        .perform(post(CONTROLLER_BASE + "/" + id + "/" + type + "/move").content(makeJson(req))
+        .perform(post(CONTROLLER_BASE + "/" + id + "/" + type + "/move").content(makeJson(request))
             .contentType(MediaType.APPLICATION_JSON))
         .andDo(print())
         .andExpect(status().isOk());
 
 
-    assertTrue(isContained(req.requisitionId(), req.itemIds().get(0), targetType, getReq));
+    assertTrue(isContained(request.requisitionId(), request.itemIds().get(0), targetType, getRequisition));
```
djcooke: What if there are more items in the request?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/RequisitionRestControllerST.java
```diff
@@ -122,46 +174,56 @@ public void testMoveSamples() throws Exception {
...
+    Long reqId = 1L;
+    Long targetId = 4L;
+    RequisitionSupplementalSampleId id = new RequisitionSupplementalSampleId();
+    id.setRequisitionId(reqId);
+    id.setSample(currentSession().get(SampleImpl.class, targetId));
+    Class<RequisitionSupplementalSample> targetClass = RequisitionSupplementalSample.class;
+
+
+    // the sample should not be a part of any requisition as a supplemental sample at this point
+    assertNull(currentSession().get(targetClass, id));
```
djcooke: Comment suggests that this is checking all requisitions, but it's only checking one specific requisition

PR #3127, opened 2025-08-06
[GLT-4524] added array model view tests

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4524

Added ArrayModelController.java tests, as well as template tests for model bulk edit, bulk create, and list.

### DIFF
--- a/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
+++ b/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
@@ -27,6 +27,8 @@ import org.springframework.test.context.web.WebAppConfiguration;
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.web.context.WebApplicationContext;
 import org.springframework.util.LinkedMultiValueMap;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.springframework.test.context.TestExecutionListeners.MergeMode;
 import org.springframework.test.context.TestExecutionListeners;
@@ -409,5 +411,81 @@ public abstract class AbstractST {
         .andExpect(jsonPath("$.id").value(id));
 
     return result;
+  }
+
+
+  // simple utility method to make maps out of property lists
+  // form (key, value, key, value, ...)
+  private HashMap<String, String> propMapMaker(List<String> props) throws Exception {
+    HashMap<String, String> properties = new HashMap<String, String>();
+    for (int i = 0; i < props.size() / 2; i += 2) {
+      properties.put(props.get(i), props.get(i + 1));
+    }
+    return properties;
+  }
+
+  // makes a list of maps
+  // lets you make maps for specific properties for several objects
+  private List<HashMap<String, String>> multiPropMapMaker(List<List<String>> props) throws Exception {
+    List<HashMap<String, String>> properties = new ArrayList<HashMap<String, String>>();
+    for (List<String> lis : props) {
+      properties.add(propMapMaker(lis));
+    }
+    return properties;
+  }
+
+  // tests the properties of objectes in a JSON array
+  private void testJsonArrayProperties(String response, List<HashMap<String, String>> properties) throws Exception {
+    assertEquals(properties.size(), ((List<Object>) JsonPath.read(response, "$.*")).size());
+    for (int i = 0; i < properties.size(); i++) {
+      HashMap<String, String> currMap = properties.get(i);
+      for (Map.Entry<String, String> entry : currMap.entrySet()) {
+        String key = entry.getKey();
+        String expectedValue = entry.getValue();
+        assertEquals(expectedValue, JsonPath.read(response, "$[" + i + "]." + key).toString());
+      }
+    }
+  }
+
+  // tests model static list endpoints
+  protected void testModelList(String url, String listAttribute, List<List<String>> properties)
+      throws Exception {
+    ResultActions ac = getMockMvc().perform(get(url).accept(MediaType.APPLICATION_JSON));
+    if (DEBUG_MODE)
+      ac = ac.andDo(print());
+
+    String response = ac.andExpect(status().isOk())
+        .andExpect(model().attributeExists(listAttribute))
+        .andReturn().getModelAndView().getModel().get(listAttribute).toString();
+
+    testJsonArrayProperties(response, multiPropMapMaker(properties));
+  }
+
+  // tests model bulk edit endpoints
+  protected void testModelBulkEdit(String url, String ids, String listAttribute, List<List<String>> properties)
+      throws Exception {
+    ResultActions ac = getMockMvc().perform(post(url).param("ids", ids).accept(MediaType.APPLICATION_JSON));
+    if (DEBUG_MODE)
+      ac = ac.andDo(print());
+
+    String response = ac.andExpect(status().isOk())
+        .andExpect(model().attributeExists(listAttribute))
+        .andReturn().getModelAndView().getModel().get(listAttribute).toString();
+
+    testJsonArrayProperties(response, multiPropMapMaker(properties));
+  }
+
+
+  // test model bulk create endpoints
+  protected void testModelBulkCreate(String url, int numCreated, String listAttribute) throws Exception {
+    ResultActions ac = getMockMvc().perform(get(url)
+        .accept(MediaType.APPLICATION_JSON).param("quantity", Integer.toString(numCreated)));
+    if (DEBUG_MODE)
+      ac = ac.andDo(print());
+
+    ac = ac.andExpect(status().isOk());
+
+    String response = ac.andReturn().getModelAndView().getModel().get(listAttribute).toString();
+    assertEquals(numCreated, ((List<Object>) JsonPath.read(response, "$.*")).size());
   }
 }
--- a/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayModelControllerST.java
+++ b/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayModelControllerST.java
@@ -1,0 +1,26 @@
+package uk.ac.bbsrc.tgac.miso.webapp.springtest;
+
+import org.junit.Test;
+import java.util.Arrays;
+
+public class ArrayModelControllerST extends AbstractST {
+
+  private static final String CONTROLLER_BASE = "/arraymodel";
+
+  @Test
+  public void testList() throws Exception {
+    testModelList(CONTROLLER_BASE + "/list", "data", Arrays
+        .asList(Arrays.asList("id", "1", "alias", "Test BeadChip"), Arrays.asList("id", "2", "alias", "Unused")));
+  }
+
+  @Test
+  public void testBulkCreate() throws Exception {
+    testModelBulkCreate(CONTROLLER_BASE + "/bulk/new", 3, "input");
+  }
+
+  @Test
+  public void testBulkEdit() throws Exception {
+    testModelBulkEdit(CONTROLLER_BASE + "/bulk/edit", "1,2", "input",
+        Arrays.asList(Arrays.asList("id", "1", "alias", "Test BeadChip"), Arrays.asList("id", "2", "alias", "Unused")));
+  }
+}


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -410,4 +412,80 @@ protected ResultActions baseTestGetById(String controllerBase, int id) throws Ex
 
     return result;
   }
+
+
+  // simple utility method to make maps out of property lists
+  // form (key, value, key, value, ...)
+  private HashMap<String, String> propMapMaker(List<String> props) throws Exception {
```
djcooke: MISO has a `MapBuilder` utility class that is a cleaner way to accomplish this

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -410,4 +412,80 @@ protected ResultActions baseTestGetById(String controllerBase, int id) throws Ex
...
+    List<HashMap<String, String>> properties = new ArrayList<HashMap<String, String>>();
+    for (List<String> lis : props) {
+      properties.add(propMapMaker(lis));
+    }
+    return properties;
+  }
+
+  // tests the properties of objectes in a JSON array
+  private void testJsonArrayProperties(String response, List<HashMap<String, String>> properties) throws Exception {
+    assertEquals(properties.size(), ((List<Object>) JsonPath.read(response, "$.*")).size());
```
djcooke: Will probably still require a cast, but you could use JsonPath's `.length()` function to simplify it to an `Integer` instead of casting the list just to call `.size()` afterwards

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -410,4 +412,80 @@ protected ResultActions baseTestGetById(String controllerBase, int id) throws Ex
...
+  // lets you make maps for specific properties for several objects
+  private List<HashMap<String, String>> multiPropMapMaker(List<List<String>> props) throws Exception {
+    List<HashMap<String, String>> properties = new ArrayList<HashMap<String, String>>();
+    for (List<String> lis : props) {
+      properties.add(propMapMaker(lis));
+    }
+    return properties;
+  }
+
+  // tests the properties of objectes in a JSON array
```
djcooke: Use JavaDoc format when when making class/method-level comments like this. This is especially useful on public/protected methods so the documentation becomes visible outside of the file within IDEs. Better to use the correct format on private methods too though.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -410,4 +412,82 @@ protected ResultActions baseTestGetById(String controllerBase, int id) throws Ex
...
+  private void testJsonArrayProperties(String response, List<HashMap<String, String>> properties) throws Exception {
+    assertEquals(properties.size(), ((List<Object>) JsonPath.read(response, "$.*")).size());
+    for (int i = 0; i < properties.size(); i++) {
+      HashMap<String, String> currMap = properties.get(i);
+      for (Map.Entry<String, String> entry : currMap.entrySet()) {
+        String key = entry.getKey();
+        String expectedValue = entry.getValue();
+        String actual = JsonPath.read(response, "$[" + i + "]." + key).toString();
```
djcooke: Not all fields should be evaluated as strings

PrabhjotChopra: In order to compare the expected properties to the actual retrieved properties, I went with using a <String, String> map to pass the expected properties, as I was unable to find anything online about a map-style data structure in java with multiple possible value types (e.g. Long value type for ids, String value typefor aliases, ...). I am not sure as to how I might pass the expected properties as params if they aren't just all strings.

PrabhjotChopra: In order to compare the expected properties to the actual retrieved properties, I went with using a <String, String> map to pass the expected properties, as I was unable to find anything online about a map-style data structure in java with multiple possible value types (e.g. Long value type for ids, String value typefor aliases, ...). I am not sure as to how I might pass the expected properties as params if they aren't just all strings.

djcooke: Maybe leave this testing to the implementations rather than trying to abstract it

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -410,4 +413,85 @@ protected ResultActions baseTestGetById(String controllerBase, int id) throws Ex
...
+        .andExpect(model().attributeExists(listAttribute))
+        .andReturn().getModelAndView().getModel().get(listAttribute).toString();
+
+    testJsonArrayProperties(response, properties);
+  }
+
+  /**
+   * Tests model bulk edit endpoints.
+   */
+  protected void testModelBulkEdit(String url, String ids, String listAttribute,
```
djcooke: Change `ids` to `List<Long>` and do the conversion here so you don't need to do it for every implementation

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -410,4 +413,85 @@ protected ResultActions baseTestGetById(String controllerBase, int id) throws Ex
 
     return result;
   }
+
+  /**
+   * Simple utility method to turn a list of integer ids into a readable string of ids (as read by the
+   * miso.core.util package)
+   */
+  protected String idListString(List<Integer> ids) {
```
djcooke: `String.join` does what you need here

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -410,4 +413,85 @@ protected ResultActions baseTestGetById(String controllerBase, int id) throws Ex
...
+    for (int i = 0; i < properties.size(); i++) {
+      HashMap<String, Object> currMap = properties.get(i);
+      for (Map.Entry<String, Object> entry : currMap.entrySet()) {
+        String key = entry.getKey();
+        Object expectedValue = entry.getValue();
+        Object actualValue = JsonPath.read(response, "$[" + i + "]." + key);
+
+        assertEquals(expectedValue, actualValue);
+        if (DEBUG_MODE)
+          System.out.println("\n\nEXPECTED IS: " + expectedValue + ", WHILE ACTUAL IS: " + actualValue + "\n\n");
```
djcooke: `assertEquals` outputs a similar message already

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayModelControllerST.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.List;
+import java.util.Map;
+
+public class ArrayModelControllerST extends AbstractST {
+
+  private static final String CONTROLLER_BASE = "/arraymodel";
+  private static final Class<ArrayModel> entityClass = ArrayModel.class;
+
+  private List<HashMap<String, Object>> makeObjectMaps(List<Integer> ids) {
+    List<HashMap<String, Object>> mappedProperties = new ArrayList<HashMap<String, Object>>();
```
djcooke: Declare and return as `Map` instead of `HashMap`. Then you don't need to cast below. Always best practice to declare the interface unless you really need something implementation-specific. In this case, this all breaks if `MapBuilder` changes at some point to use a different type of `Map`.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -410,4 +413,85 @@ protected ResultActions baseTestGetById(String controllerBase, int id) throws Ex
...
+      }
+    }
+  }
+
+  /**
+   * Tests model static list endpoints.
+   */
+  protected void testModelList(String url, String listAttribute, List<HashMap<String, Object>> properties)
```
djcooke: Unclear what a model list is. Rename `testStaticListPage`. Similar for other methods - `testBulkEditPage`, etc.

Also unclear what `listAttribute` is. Please expand JavaDoc to include descriptions of the parameters. Same for below methods.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayModelControllerST.java
```diff
@@ -0,0 +1,50 @@
...
+
+    for (int i = 0; i < ids.size(); i++) {
+      MapBuilder<String, Object> currMap = new MapBuilder<String, Object>();
+      ArrayModel arrayModel = currentSession().get(entityClass, ids.get(i));
+      currMap.put("id", ids.get(i)).put("alias", arrayModel.getAlias());
+      mappedProperties.add((HashMap) currMap.build()); // safe cast since MapBuilder using HashMap
+    }
```
djcooke: This is why I'm not seeing the purpose of this abstraction. Building the properties map seems like at least as much effort as writing your assertions here for each field would be.

PrabhjotChopra: Given that quite a few other view controllers only have bulk create, bulk edit, and static list endpoints, this abstraction makes writing those tests significantly faster. With this, I can essentially paste the tests here to any other simple view controller (which there are a lot of), and all I need to change is the IDs in the `ids` list, the object type I initialize (i.e. change ArrayModel to whatever the other endpoint is testing), and maybe the keys in the map (e.g. if the object doesn't have an alias, then something else should be used). 
I found that the time to write and finish a full REST controller test file went down significantly after I added the template bulk async create/update and delete methods, so I wanted to do a similar thing here.

I would write a template method for this snippet too, but that would require passing a list of Function types with different return types, which seems like more trouble than its worth.

djcooke: I'm not saying don't abstract those methods at all. I'm saying that checking the properties of the objects returned isn't a helpful part of that. Without this, you would have the same thing, but simpler. Below would be instead of building the properties map AND asserting the properties separately in the abstract class, and it's simpler than either one of those. You just need to update the type and fields you're checking if necessary, same as you said you'd need to already.

```
private void assertDbVsModelObjects(String resultJson) {
   for (int i = 0; i < ids.size(); i++) {
      ArrayModel dbObject = currentSession().get(entityClass, ids.get(i));
      assertEquals(dbObject.getId(), JsonPath.read(response, "$[" + i + "].id"));
      assertEquals(dbObject.getAlias(), JsonPath.read(response, "$[" + i + "].alias"));
    }
}
```

It would also be good to return the model, or maybe even the model and view, from your abstract methods because there are often additional model attributes that should be tested beyond the main collection of objects, and some of those are specific to the implementation.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -414,84 +534,85 @@ protected ResultActions baseTestGetById(String controllerBase, int id) throws Ex
...
+   * @return Model and View response from the endpoint
    */
-  protected void testModelBulkEdit(String url, String ids, String listAttribute,
-      List<HashMap<String, Object>> properties)
+  protected ModelAndView testBulkEditPage(String url, List<Long> ids)
       throws Exception {
-    ResultActions ac = getMockMvc().perform(post(url).param("ids", ids).accept(MediaType.APPLICATION_JSON));
+
+    List<String> stringList = ids.stream().map(String::valueOf).collect(Collectors.toList());
+    String stringIds = String.join(",", stringList);
```
djcooke: Instead of collecting to a list and then using `String.join`, you can use `Collectors.joining`

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayModelControllerST.java
```diff
@@ -3,48 +3,45 @@
...
+  private void assertDbVsModelObjects(String resultJson, List<Long> ids) {
+    assertEquals(Integer.valueOf(ids.size()), JsonPath.read(resultJson, "$.length()"));
 
     for (int i = 0; i < ids.size(); i++) {
-      MapBuilder<String, Object> currMap = new MapBuilder<String, Object>();
-      ArrayModel arrayModel = currentSession().get(entityClass, ids.get(i));
-      currMap.put("id", ids.get(i)).put("alias", arrayModel.getAlias());
-      mappedProperties.add((HashMap) currMap.build()); // safe cast since MapBuilder using HashMap
+      ArrayModel dbObject = currentSession().get(entityClass, ids.get(i));
+      assertEquals(dbObject.getId(), ((Integer) JsonPath.read(resultJson, "$[" + i + "].id")).longValue());
```
djcooke: Could add utility methods to the abstract class to eliminate casting and simplify reuse. Something like this:

```
private static Integer readInteger(String resultJson, String path) {
  return JsonPath.read(resultJson, path);
}

private static long readLong(String resultJson, String path) {
  return readInteger(resultJson, path).longValue();
}
```

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayModelControllerST.java
```diff
@@ -0,0 +1,47 @@
...
+      assertEquals(dbObject.getAlias(), JsonPath.read(resultJson, "$[" + i + "].alias"));
+    }
+  }
+
+  @Test
+  public void testList() throws Exception {
+    List<Long> ids = Arrays.asList(1L, 2L);
+
+    assertDbVsModelObjects(testStaticListPage(CONTROLLER_BASE + "/list", "data"), ids);
```
djcooke: Splitting into two lines rather than nesting would improve readability. Same below

avarsava (approved):

PR #3128, opened 2025-08-06
[GLT-4537] added edit array view tests

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4537

Added edit array view tests. Also added template test for edit new entity, and the setup form endpoints.

Depends on #3127


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/EditArrayControllerST.java
```diff
@@ -0,0 +1,35 @@
...
+    assertEquals(dbObject.getAlias(), JsonPath.read(resultJson, "$.alias"));
+  }
+
+  @Test
+  public void testNew() throws Exception {
+    baseTestNewModel(CONTROLLER_BASE + "/new", "New Array");
+  }
+
+  @Test
+  public void testSetupForm() throws Exception {
```
djcooke: I realize that's the method name, but please rename `testEditPage` for clarity

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -647,4 +663,67 @@ protected static Integer readInteger(String resultJson, String path) {
...
+  }
+
+  /**
+   * Tests model "edit entity" form setup
+   * 
+   * @param url url to query
+   * @param listModelAttribute Model attribute, if relevant
+   * @return Resulting JSON response
+   */
+  protected String testModelFormSetup(String url, String listModelAttribute) throws Exception {
```
djcooke: Rename `listModelObject` -> `modelJson`

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -647,4 +663,67 @@ protected static Integer readInteger(String resultJson, String path) {
...
+    return ac;
+  }
+
+  /**
+   * Tests single-entity edit model endpoints (usually a get mapping to /{entityId})
+   * 
+   * @param url URL to query
+   * @return ResultActions from request
+   */
+  protected ResultActions baseTestEditModel(String url) throws Exception {
```
djcooke: This is unused in this PR and is identical to `testModelFormSetup` except for returning ResultActions instead of the ModelAndView. `baseTestEditModel` is a better name though

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
PrabhjotChopra: are you referring to `listModelAttribute`? That's just the model attribute that has the dto data.

djcooke: Yes. It's not a list, though it is actually the attribute name, so I suppose `jsonModelAttribute` might be a better name

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java
```diff
@@ -647,4 +663,58 @@ protected static Integer readInteger(String resultJson, String path) {
...
+  /**
+   * Tests model "edit entity" form setup
+   * 
+   * @param url url to query
+   * @param listModelAttribute Model attribute, if relevant
+   * @return Resulting JSON response
+   */
+  protected String testModelFormSetup(String url, String listModelAttribute) throws Exception {
+    return testModelFormSetup(url).getModel().get(listModelAttribute).toString();
+  }
```
djcooke: 1. Why are these named differently from `baseTestEditModel` when they serve the same purpose?
2. Why do we need 3 different versions of this?

PrabhjotChopra: 1. I made these base tests based on the endpoints I saw as I went along, not realising that they all did basically the same thing.
2. for some endpoints, it's easier to test with String or Model and View than a ResultActions I think. Would it make sense to just rename these based on their return types? (so like "baseTestEditModelString" or something) Or should I just remove them and make the changes when they're called in the test files?

djcooke: Do you ever actually check anything on the View? It might be simpler to just have one base test that returns the Model. The caller can then call `.get(attributeName)` on that if they want, rather than providing the attribute name in a parameter. It's no less convenient.

djcooke: I don't think the ResultActions ever needs to be returned either as the base test should check what is needed there

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/EditArrayControllerST.java
```diff
@@ -0,0 +1,36 @@
...
+  @Test
+  public void testNew() throws Exception {
+    baseTestNewModel(CONTROLLER_BASE + "/new", "New Array");
+  }
+
+  @Test
+  public void testEditPage() throws Exception {
+    int id = 1;
+    String objectAttribute = "arrayJson";
+    String resultJson = baseTestEditModel(CONTROLLER_BASE + "/1").get(objectAttribute).toString();
```
callunity: Please use the ID in the URL

djcooke (approved):

PR #3140, opened 2025-08-12
[GLT-4645] added Spring Test controller endpoint coverage metrics

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4645

Now, when you run `mvn clean verify -DskipSTs=false ...`, an `index.html` file will be created in `/miso-web/target/site/`. Opening this file in a browser will take you to a report which gives you information regarding the endpoint coverage.

### DIFF
--- a/miso-web/pom.xml
+++ b/miso-web/pom.xml
@@ -288,11 +288,6 @@
       <version>3.0</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.jacoco</groupId>
-      <artifactId>jacoco-maven-plugin</artifactId>
-      <version>0.8.12</version>
-    </dependency>
   </dependencies>
   <build>
     <finalName>ROOT</finalName>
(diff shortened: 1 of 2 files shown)


miso-web/pom.xml
```diff
@@ -287,6 +288,11 @@
       <version>3.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jacoco</groupId>
+      <artifactId>jacoco-maven-plugin</artifactId>
+      <version>0.8.12</version>
+    </dependency>
```
djcooke: Don't include as a dependency

miso-web/pom.xml
```diff
@@ -1066,6 +1072,40 @@
               </placeholders>
             </configuration>
           </plugin>
+          <plugin>
+            <groupId>org.jacoco</groupId>
+            <artifactId>jacoco-maven-plugin</artifactId>
+            <version>0.8.12</version>
```
djcooke: Add to `pluginManagement` in the root MISO `pom.xml` instead of setting the version here

miso-web/pom.xml
```diff
@@ -13,6 +13,7 @@
   <properties>
     <skipITs>true</skipITs>
     <skipSTs>true</skipSTs>
+    <jacoco.skip>true</jacoco.skip>
```
djcooke, line 16: Please add documentation to the developer's guide for how to enable, and what to look for in the results

PR #3147, opened 2025-08-19
[GLT-4516] added user rest tests

PrabhjotChopra:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4516

### DIFF
--- a/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/UserRestControllerST.java
+++ b/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/UserRestControllerST.java
@@ -36,7 +36,7 @@ public class UserRestControllerST extends AbstractST {
     dto.setInternal(true);
     dto.setLoginName("new");
     dto.setFullName("new guy");
-    dto.setPassword("OntarioInstituteOfCancerResearch2025");
+    dto.setPassword("OntarioInstituteForCancerResearch2025");
     return dto;
   }
 
@@ -105,7 +105,7 @@ public class UserRestControllerST extends AbstractST {
     // non-admin user cannot change password of another user
     PasswordChangeDto dto = new PasswordChangeDto();
     dto.setOldPassword("user");
-    dto.setNewPassword("OntarioInstituteOfCancerResearch2025");
+    dto.setNewPassword("OntarioInstituteForCancerResearch2025");
 
     getMockMvc()
         .perform(post(CONTROLLER_BASE + "/3/password").content(makeJson(dto)).contentType(MediaType.APPLICATION_JSON))
@@ -121,7 +121,7 @@ public class UserRestControllerST extends AbstractST {
     // changing of one's own password
     PasswordChangeDto dto = new PasswordChangeDto();
     dto.setOldPassword("user");
-    dto.setNewPassword("OntarioInstituteOfCancerResearch2025");
+    dto.setNewPassword("OntarioInstituteForCancerResearch2025");
 
     getMockMvc()
         .perform(post(CONTROLLER_BASE + "/3/password").content(makeJson(dto)).contentType(MediaType.APPLICATION_JSON))
@@ -137,7 +137,7 @@ public class UserRestControllerST extends AbstractST {
 
     PasswordChangeDto dto = new PasswordChangeDto();
     dto.setOldPassword("user");
-    dto.setNewPassword("OntarioInstituteOfCancerResearch2025");
+    dto.setNewPassword("OntarioInstituteForCancerResearch2025");
 
     getMockMvc()
         .perform(post(CONTROLLER_BASE + "/3/password").content(makeJson(dto)).contentType(MediaType.APPLICATION_JSON))


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/UserRestControllerST.java
```diff
@@ -0,0 +1,129 @@
...
+  @Test
+  @WithMockUser(username = "admin", password = "admin", roles = {"INTERNAL", "ADMIN"})
+  public void testCreate() throws Exception {
+    // must be admin to create user
+    UserImpl user = baseTestCreate(CONTROLLER_BASE, makeCreateDto(), entityClass, 200);
+    assertEquals(false, user.isAdmin());
+    assertEquals(true, user.isActive());
+    assertEquals(true, user.isInternal());
+    assertEquals("new", user.getLoginName());
+    assertEquals("new guy", user.getFullName());
```
callunity: If you save the results of makeCreateDto earlier in this function, you could replace these hard-coded expected values with the actual values of the result of makeCreateDto (e.g. `assertEquals(dto.isAdmin(), user.isAdmin())` )

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/UserRestControllerST.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @Test
+  public void testDeleteFail() throws Exception {
+    testDeleteUnauthorized(entityClass, 4, CONTROLLER_BASE);
+  }
+
+
+  @Test
+  public void testSearch() throws Exception {
+    baseSearchByTerm(CONTROLLER_BASE, "hhenderson", Arrays.asList(4));
```
callunity, line 99: I had to open up the AbstractST class to figure out what `Arrays.asList(4))` represented. Could you make this more readily understandable by extracting that argument to a variable named something like `expectedIds` ? 

(Though it looks like that's not truly what the test is asserting; it's just confirming that the number of IDs passed in is the number of IDs returned in the search. I think it would be clearer if the argument was an integer of `expectedResultsCount` rather than a list of integers, but I'm ambivalent on whether it's worth refactoring all the tests that use this function.)

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/UserRestControllerST.java
```diff
@@ -0,0 +1,129 @@
...
+        .perform(post(CONTROLLER_BASE + "/3/password").content(makeJson(dto)).contentType(MediaType.APPLICATION_JSON))
+        .andDo(print())
+        .andExpect(status().isOk());
+
+  }
+
+  @Test
+  @WithMockUser(username = "hhenderson", roles = {"INTERNAL"})
+  public void testChangePasswordFail() throws Exception {
+    // must be admin or matching owner to change user password
```
callunity: `// non-admin user cannot change password of another user` ?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/UserRestControllerST.java
```diff
@@ -0,0 +1,129 @@
...
+
+    PasswordChangeDto dto = new PasswordChangeDto();
+    dto.setOldPassword("user");
+    dto.setNewPassword("OntarioInstituteOfCancerResearch2025");
+
+    getMockMvc()
+        .perform(post(CONTROLLER_BASE + "/3/password").content(makeJson(dto)).contentType(MediaType.APPLICATION_JSON))
+        .andDo(print())
+        .andExpect(status().isUnauthorized());
+  }
```
callunity, line 147: Please also add a test that a user can successfully change their own password.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/UserRestControllerST.java
PrabhjotChopra: Sure, I'll change the parameter name there. I just use `Arrays.asList....` a lot since it's slightly faster to write.

As for the ID checking, the function `baseSearchByTerm` calls private function `AbstractST#checkIds` at the end, and the `checkIds` function asserts that both that the number of results returned and the IDs both match the expected values.

callunity: Oh! If that's the case, it seems like this line might be redundant? https://github.com/miso-lims/miso-lims/blob/16116a00a0907261eaa24621d4f095d7594197a1/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/AbstractST.java#L442C3-L442C57 (since it's also checked in the function)

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/UserRestControllerST.java
```diff
@@ -0,0 +1,148 @@
...
+  private PasswordEncoder passwordEncoder;
+
+  private UserDto makeCreateDto() throws Exception {
+    UserDto dto = new UserDto();
+    dto.setAdmin(false);
+    dto.setActive(true);
+    dto.setInternal(true);
+    dto.setLoginName("new");
+    dto.setFullName("new guy");
+    dto.setPassword("OntarioInstituteOfCancerResearch2025");
```
djcooke: Ontario Institute _for_ Cancer Research

callunity (approved):

PR #3159, opened 2025-10-06
Update URL for miso.properties in web 

genericdata:
Original URL for miso.properties is wrong. Fixed by going up a directory.


djcooke (approved):
Thanks for the correction!

PR #3170, opened 2025-11-26
GLT-4593-Test-ListUsersController

Nandu-26:
Jira ticket or GitHub issue: GLT-4593

- [ ] Includes a change file

### DIFF
--- a/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ListUsersControllerST.java
+++ b/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ListUsersControllerST.java
@@ -17,39 +17,16 @@ public class ListUsersControllerST extends AbstractST {
   private static final String USERS_ENDPOINT = "/admin/users";
 
   @Test
-  public void testList() throws Exception {
-    String json = testStaticListPage(USERS_ENDPOINT, "data");
-
-    // verify response contains users
-    int userCount = JsonPath.<Integer>read(json, "$.length()");
-    assertTrue("Expected at least one user in response", userCount > 0);
-
-    // spot check first user has required fields
-    assertNotNull(JsonPath.read(json, "$[0].id"));
-    assertNotNull(JsonPath.read(json, "$[0].loginName"));
-    assertNotNull(JsonPath.read(json, "$[0].fullName"));
-    assertNotNull(JsonPath.read(json, "$[0].email"));
-
-    // boolean flags should be present
-    assertNotNull(JsonPath.read(json, "$[0].admin"));
-    assertNotNull(JsonPath.read(json, "$[0].internal"));
-    assertNotNull(JsonPath.read(json, "$[0].active"));
-    assertNotNull(JsonPath.read(json, "$[0].loggedIn"));
-  }
-
-  @Test
   public void testUserDataIntegrity() throws Exception {
     String json = testStaticListPage(USERS_ENDPOINT, "data");
 
-    // check a sample of users from the JSON response for data consistency
     int jsonUserCount = JsonPath.<Integer>read(json, "$.length()");
-    int samplesToCheck = Math.min(3, jsonUserCount);
+    assertTrue("Expected at least 3 users in test data", jsonUserCount >= 3);
 
-    for (int i = 0; i < samplesToCheck; i++) {
+    for (int i = 0; i < 3; i++) {
       String jsonPath = "$[" + i + "]";
       String loginName = JsonPath.read(json, jsonPath + ".loginName");
 
-      // query database for this specific user by login name
       UserImpl expected = currentSession()
           .createQuery("from UserImpl where loginName = :loginName", UserImpl.class)
           .setParameter("loginName", loginName)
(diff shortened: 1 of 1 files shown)


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ListUsersControllerST.java
```diff
@@ -0,0 +1,76 @@
...
+    assertNotNull(JsonPath.read(json, "$[0].email"));
+
+    // boolean flags should be present
+    assertNotNull(JsonPath.read(json, "$[0].admin"));
+    assertNotNull(JsonPath.read(json, "$[0].internal"));
+    assertNotNull(JsonPath.read(json, "$[0].active"));
+
+    // loggedIn flag is populated from session registry
+    Object loggedInFlag = JsonPath.read(json, "$[0].loggedIn");
+    assertTrue("loggedIn should be boolean", loggedInFlag instanceof Boolean);
```
avarsava: why test the data type of loggedIn but not anything else?

Nandu-26: You're right - that was inconsistent. I've fixed it to use assertNotNull() for loggedIn like all the other fields. Thanks for catching that!

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ListUsersControllerST.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @Test
+  public void testUserDataIntegrity() throws Exception {
+    String json = testStaticListPage(USERS_ENDPOINT, "data");
+    List<UserImpl> fromDb = currentSession()
+        .createQuery("from UserImpl order by id", UserImpl.class)
+        .list();
+
+    // check a sample of users for data consistency
+    int samplesToCheck = Math.min(3, fromDb.size());
```
callunity: The data in `json` is paginated and may not be the full data set, so there's a theoretical possibility that the users in `fromDb` aren't in the `json`. You could instead get a subset of samplesToCheck from the `json`, and then query them in the database by login name or similar

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ListUsersControllerST.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static final String USERS_ENDPOINT = "/admin/users";
+
+  @Test
+  public void testList() throws Exception {
+    String json = testStaticListPage(USERS_ENDPOINT, "data");
+    List<UserImpl> persisted = currentSession().createQuery("from UserImpl", UserImpl.class).list();
+
+    // verify count matches what we have in db
+    assertEquals(persisted.size(), JsonPath.<Integer>read(json, "$.length()").intValue());
```
callunity: It's possible that the database count could exceed the json count, since the data is paginated - unlikely in test data, but possible. Is this check needed, or is there another way you could assess data integrity?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ListUsersControllerST.java
Nandu-26: Thanks for pointing this out, I'll get users from JSON first, query database for those specified users by LoginName and that should work will pagination.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ListUsersControllerST.java
Nandu-26: Yes, you're correct that count check isn't reliable with pagination. I'll remove it and just verify if the response has users (> 0 check), All required fields are present and data accuracy is checked in testUserDataIntegrity by querying the DB for specific users in the JSON.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ListUsersControllerST.java
```diff
@@ -0,0 +1,72 @@
...
+    assertNotNull(JsonPath.read(json, "$[0].loggedIn"));
+  }
+
+  @Test
+  public void testUserDataIntegrity() throws Exception {
+    String json = testStaticListPage(USERS_ENDPOINT, "data");
+
+    // check a sample of users from the JSON response for data consistency
+    int jsonUserCount = JsonPath.<Integer>read(json, "$.length()");
+    int samplesToCheck = Math.min(3, jsonUserCount);
```
djcooke: 1. This could potentially be zero, causing this test to pass without checking anything when there is obviously a problem.
2. We control the test data, so there is no need to make it flexible like this. You can expect the number of samples that are present in the test data.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ListUsersControllerST.java
```diff
@@ -0,0 +1,72 @@
...
+    assertNotNull(JsonPath.read(json, "$[0].id"));
+    assertNotNull(JsonPath.read(json, "$[0].loginName"));
+    assertNotNull(JsonPath.read(json, "$[0].fullName"));
+    assertNotNull(JsonPath.read(json, "$[0].email"));
+
+    // boolean flags should be present
+    assertNotNull(JsonPath.read(json, "$[0].admin"));
+    assertNotNull(JsonPath.read(json, "$[0].internal"));
+    assertNotNull(JsonPath.read(json, "$[0].active"));
+    assertNotNull(JsonPath.read(json, "$[0].loggedIn"));
```
djcooke: This test is redundant. It's a worse version of the one below because it's only checking for non-null instead of checking the specific values.

PR #3171, opened 2025-11-27
GLT-4489 SampleClassRestController

Nandu-26:
Jira ticket or GitHub issue:

- [ ] Includes a change file
- [ ] GLT-4489 SampleClassRestController

### DIFF
--- a/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
+++ b/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
@@ -84,11 +84,6 @@ public class SampleClassRestControllerST extends AbstractST {
     SampleClassImpl existing = (SampleClassImpl) currentSession().get(ENTITY_CLASS, (long) UPDATE_CLASS_ID);
     assertNotNull(existing);
 
-    assertNotNull("parent relationships is null", existing.getParentRelationships());
-    assertFalse("parent relationships is empty", existing.getParentRelationships().isEmpty());
-    assertTrue("existing has parent relationships with null parent",
-        existing.getParentRelationships().stream().allMatch(rel -> rel.getParent() != null));
-
     int parentCountBefore = existing.getParentRelationships().size();
     Set<Long> parentIdsBefore = existing.getParentRelationships().stream()
         .map(rel -> rel.getParent().getId())
@@ -111,9 +106,6 @@ public class SampleClassRestControllerST extends AbstractST {
     assertNotNull("parent relationships is null", returned.getParentRelationships());
     assertFalse("parent relationships is empty", returned.getParentRelationships().isEmpty());
     assertEquals("parent relationships count changed", parentCountBefore, returned.getParentRelationships().size());
-
-    assertTrue("returned has parent relationships with null parent",
-        returned.getParentRelationships().stream().allMatch(rel -> rel.getParent() != null));
 
     Set<Long> parentIdsAfter = returned.getParentRelationships().stream()
         .map(rel -> rel.getParent().getId())


miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -0,0 +1,126 @@
...
+        .contentType(MediaType.APPLICATION_JSON)
+        .content(makeJson(dto)))
+        .andExpect(status().isBadRequest());
+  }
+
+  @Test
+  public void testUpdate() throws Exception {
+    // create one first
+    SampleClassDto createDto = makeDto("Original", "OR");
+    SampleClass created = baseTestCreate(CONTROLLER_BASE, createDto, SampleClass.class, 201);
```
djcooke: Don't create one here. Add it to the test data if needed, or (ideally) one of the ones already there.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -0,0 +1,126 @@
...
+        .contentType(MediaType.APPLICATION_JSON)
+        .content(makeJson(dto)))
+        .andExpect(status().isNotFound());
+  }
+
+  @Test
+  public void testBulkDelete() throws Exception {
+    SampleClassDto dto = makeDto("To Delete", "TD");
+    SampleClass created = baseTestCreate(CONTROLLER_BASE, dto, SampleClass.class, 201);
+    int id = (int) created.getId();
```
djcooke: Don't create data as part of the test.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -0,0 +1,126 @@
...
+
+    // verify it's actually gone
+    assertNull(currentSession().get(SampleClass.class, id));
+  }
+
+  @Test
+  public void testBulkDeleteMultiple() throws Exception {
+    // create a couple to delete
+    SampleClass sc1 = baseTestCreate(CONTROLLER_BASE, makeDto("Class1", "C1"), SampleClass.class, 201);
+    SampleClass sc2 = baseTestCreate(CONTROLLER_BASE, makeDto("Class2", "C2"), SampleClass.class, 201);
```
djcooke: Don't create data as part of the test.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -0,0 +1,75 @@
...
+
+    getMockMvc().perform(post(CONTROLLER_BASE + "/bulk-delete")
+        .contentType(MediaType.APPLICATION_JSON)
+        .content(makeJson(ids)))
+        .andExpect(status().isNoContent());
+  }
+
+  @Test
+  @WithMockUser(username = "admin", roles = {"ADMIN"})
+  public void testDeleteNonExistent() throws Exception {
```
callunity: ~can use AbstractST.testBulkDelete here~ I'm wrong about this!

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -0,0 +1,75 @@
...
+
+import javax.ws.rs.core.MediaType;
+
+import org.junit.Test;
+import org.springframework.security.test.context.support.WithMockUser;
+
+import uk.ac.bbsrc.tgac.miso.core.data.SampleClass;
+import uk.ac.bbsrc.tgac.miso.dto.SampleClassDto;
+
+public class SampleClassRestControllerST extends AbstractST {
```
callunity, line 22: Many of the other REST controller ST classes have additional tests, like bulk create, bulk update, testDeleteFail. Can you please implement those as well? See here: https://github.com/miso-lims/miso-lims/blob/develop/miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/RunPurposeRestControllerST.java including convenience methods

djcooke: Sample classes don't have bulk create/edit. Only the endpoints defined in `SampleClassRestController` should be tested, and that's only single create, single update, and bulk delete

djcooke (commented):
There are now only tests for failed attempts to create and update..? Those are fine, but there need to be tests for a successful create and a successful update.

Edit: In-case my previous comments were unclear:
* It is ok to create data as part of the test when you are testing the creation of that data
* It is not ok to create data as part of the test when the creation of that data is not what you are testing

For example:
* If you are testing creating a sample class, your test should create the sample class and verify that it was created successfully and correctly.
* If you are testing an update method, the object you're updating should already exist in the test data - you should not do the initial creation during your test. Your test should update the existing item, and then verify that the update was successful and correct.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -44,32 +113,66 @@ public void testUpdateNonExistent() throws Exception {
...
+    if (res.getResponse().getStatus() != expected) {
+      fail("Expected HTTP " + expected + " but got "
+          + res.getResponse().getStatus()
+          + ". Response body:\n"
+          + res.getResponse().getContentAsString());
+    }
+  }
+
+
+  private SampleClass pickCloneableTemplate() throws Exception {
```
djcooke: It should be clear what data you're testing with. Select specific sample classes.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -44,32 +113,66 @@ public void testUpdateNonExistent() throws Exception {
...
+  }
+
+  private String pickUnusedOneCharSuffix(Set<String> used) {
+    for (char c = 'A'; c <= 'Z'; c++) {
+      String s = String.valueOf(c);
+      if (!used.contains(s))
+        return s;
+    }
+    return "Z";
   }
```
djcooke: Use specific values that you know are unused based on the test data

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -1,40 +1,109 @@
...
+
+    dto.setAlias("ST Created Sample Class");
+    dto.setSuffix(pickUnusedOneCharSuffix(getAllSuffixes()));
+
+    MvcResult res = getMockMvc().perform(post(CONTROLLER_BASE)
+        .contentType(MediaType.APPLICATION_JSON)
+        .content(makeJson(dto)))
+        .andReturn();
+
+    assertStatusOrFailWithBody(res, 201);
```
djcooke: Also assert that all of the fields are saved as expected

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -1,40 +1,109 @@
...
+    SampleClassDto dto = Dtos.asDto(template);
+
+    dto.setId(null);
+    dto.setChildRelationships(null);
+
+    dto.setAlias("Should Fail Identity Duplicate");
+    dto.setSuffix(pickUnusedOneCharSuffix(getAllSuffixes()));
+
+    dto.setSampleCategory("Identity");
+    dto.setSampleSubcategory("");
```
djcooke: Should be null rather than empty string

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -1,72 +1,65 @@
...
+    SampleClassDto dto = new SampleClassDto();
+    dto.setAlias("Test Identity Class");
+    dto.setSampleCategory("Identity");
+    dto.setSampleSubcategory(null);
+    dto.setSuffix(null);
+    dto.setArchived(false);
 
-    MvcResult res = getMockMvc().perform(post(CONTROLLER_BASE)
+    String json = makeJson(dto);
+    String response = getMockMvc().perform(post(CONTROLLER_BASE)
```
djcooke: Use `AbstractST#baseTestCreate`. There are other tests in here that you can use methods from the abstract class for too. Please go through each instance where you call `getMockMvc().perform` and check if there is a method in the abstract class that could replace this.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -0,0 +1,190 @@
...
+    Assert.assertNotNull(existing);
+    Assert.assertFalse(existing.getParentRelationships().isEmpty());
+
+    long parentRelationshipId = existing.getParentRelationships().iterator().next().getId();
+    Assert.assertTrue(parentRelationshipId > 0);
+
+    SampleClassDto dto = new SampleClassDto();
+    dto.setId((long) UPDATE_CLASS_ID);
+    dto.setAlias("Updated gDNA Aliquot Class");
+    dto.setSampleCategory("Aliquot"); // keep existing category
```
callunity: If you're not intending to update this field (and `archived` below), please use `existing.getX()` to indicate that these should stay the same

djcooke: Alternately, just use `Dtos#asDto` to convert `existing` and then change the one field you want. It's usually ideal not depend on other code, but for things like this it is less error prone.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -0,0 +1,190 @@
...
+
+    // Tissue classes must have a single Identity parent relationship
+    SampleValidRelationshipDto parentRel = new SampleValidRelationshipDto();
+    parentRel.setParentId(IDENTITY_CLASS_ID);
+    parentRel.setArchived(false);
+    dto.setParentRelationships(Collections.singletonList(parentRel));
+
+    SampleClassImpl created = baseTestCreate(CONTROLLER_BASE, dto, ENTITY_CLASS, 201);
+    Assert.assertNotNull(created);
+    Assert.assertEquals("Test Tissue Class", created.getAlias());
```
djcooke: Check all fields that you set.

djcooke: ?

miso-lims
djcooke, line 1: What is this?

Nandu-26: This is just the “happy path” check for the create endpoint.
The idea here is simply to confirm that a SampleClass can be created successfully and persisted, and that the primary user-visible field (alias) is saved correctly. I’m not trying to validate every field mapping in this test, just that creation works end-to-end.
For updates, I’ve made sure all non-target fields are explicitly preserved as discussed earlier.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -0,0 +1,137 @@
...
+    testCreateUnauthorized(CONTROLLER_BASE, dto, ENTITY_CLASS);
+  }
+
+  @Test
+  @WithMockUser(username = "admin", password = "admin", roles = {"INTERNAL", "ADMIN"})
+  public void testUpdateSuccess() throws Exception {
+    SampleClassImpl existing = (SampleClassImpl) currentSession().get(ENTITY_CLASS, (long) UPDATE_CLASS_ID);
+    assertNotNull("existing is null", existing);
+    assertNotNull("existing parent relationships is null", existing.getParentRelationships());
+    assertFalse("existing has no parent relationships", existing.getParentRelationships().isEmpty());
```
djcooke: You don't need to test the test data

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -53,11 +55,13 @@ public void testCreateSuccess() throws Exception {
...
     assertFalse("parent relationships is empty", created.getParentRelationships().isEmpty());
-    assertEquals("parent relationships size != 1", 1, created.getParentRelationships().size());
+
+    boolean foundExpectedParent = created.getParentRelationships().stream()
+        .anyMatch(rel -> rel.getParent() != null && rel.getParent().getId() == IDENTITY_CLASS_ID);
+    assertTrue("expected parent relationship not found", foundExpectedParent);
 
     SampleValidRelationship createdRel = created.getParentRelationships().iterator().next();
     assertNotNull(createdRel.getParent());
-    assertEquals("parent ID mismatch", IDENTITY_CLASS_ID, createdRel.getParent().getId());
```
djcooke: Why did you make this change? It was clearer how you had it before. You also removed the size check, so now you don't know if there are additional relationships that should not be there.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -78,15 +82,16 @@ public void testCreateUnauthorized() throws Exception {
...
-    assertFalse("existing has no parent relationships", existing.getParentRelationships().isEmpty());
+    assertNotNull(existing);
 
-    SampleValidRelationship existingRel = existing.getParentRelationships().iterator().next();
-    long parentId = existingRel.getParent().getId();
+    Set<Long> parentIdsBefore = existing.getParentRelationships() == null ? Collections.emptySet()
+        : existing.getParentRelationships().stream()
+            .filter(rel -> rel.getParent() != null)
+            .map(rel -> rel.getParent().getId())
+            .collect(Collectors.toSet());
```
djcooke: Why would `existing.getParentRelationships()` be null? Wouldn't that indicate a problem with your test data?

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -100,21 +105,23 @@ public void testUpdateSuccess() throws Exception {
...
 
-    boolean foundParent = returned.getParentRelationships().stream()
-        .anyMatch(rel -> rel.getParent().getId() == parentId);
-    assertTrue("expected parent relationship not found", foundParent);
+    if (!parentIdsBefore.isEmpty()) {
+      boolean containsAllParents = parentIdsBefore.stream()
+          .allMatch(id -> returned.getParentRelationships().stream()
+              .anyMatch(rel -> rel.getParent() != null && rel.getParent().getId() == id));
+      assertTrue("parent relationships changed", containsAllParents);
+    }
```
djcooke: * You know `parentIdsBefore` shouldn't be empty, and if it is, that's bad... But here, you're choosing to ignore it and skip testing if it's empty?
* If you're trying to test that the collection contains only what you expect it to, you need to check the size too.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
Nandu-26: You're correct, I've added back the size check to ensure there's exactly 1 parent relationship, not just "at least one  correct one".

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
Nandu-26: true! I've removed null check, if existing entity doesn't have parent relationships, the test should fail as that indicates bad test data.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -82,16 +79,15 @@ public void testCreateUnauthorized() throws Exception {
...
     SampleClassImpl existing = (SampleClassImpl) currentSession().get(ENTITY_CLASS, (long) UPDATE_CLASS_ID);
     assertNotNull(existing);
 
-    Set<Long> parentIdsBefore = existing.getParentRelationships() == null ? Collections.emptySet()
-        : existing.getParentRelationships().stream()
-            .filter(rel -> rel.getParent() != null)
-            .map(rel -> rel.getParent().getId())
-            .collect(Collectors.toSet());
+    Set<Long> parentIdsBefore = existing.getParentRelationships().stream()
+        .filter(rel -> rel.getParent() != null)
```
djcooke: Why would the relationship parent be null? That would be invalid. So here, you're choosing to exclude it from comparison if invalid?

Remove filter from both here and `parentIdsAfter` below.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/SampleClassRestControllerST.java
```diff
@@ -84,8 +84,13 @@ public void testUpdateSuccess() throws Exception {
     SampleClassImpl existing = (SampleClassImpl) currentSession().get(ENTITY_CLASS, (long) UPDATE_CLASS_ID);
     assertNotNull(existing);
 
+    assertNotNull("parent relationships is null", existing.getParentRelationships());
+    assertFalse("parent relationships is empty", existing.getParentRelationships().isEmpty());
+    assertTrue("existing has parent relationships with null parent",
+        existing.getParentRelationships().stream().allMatch(rel -> rel.getParent() != null));
```
djcooke: You don't need to validate the test data

PR #3172, opened 2025-12-01
GLT 4697 - Add Pools to Worksets

riyapatel36:
Add Pools to Workset

Jira ticket - [Add Pools](https://jira.oicr.on.ca/browse/GLT-4697)

### DIFF
--- a/sqlstore/src/it/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateListPoolViewDaoIT.java
+++ b/sqlstore/src/it/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateListPoolViewDaoIT.java
@@ -43,7 +43,7 @@ public class HibernateListPoolViewDaoIT extends AbstractDAOTest {
         for(WorksetPool wp: workset.getWorksetPools()){
             actualPoolIds.add(wp.getItem().getId());
         }
-        assertEquals(worksetId, expectedPoolIds.length, actualPoolIds.size());
+        assertEquals("Expected pool count for worksetId " + worksetId + " =  " + expectedPoolIds.length+ "and actual pool count is " + actualPoolIds.size() , expectedPoolIds.length, actualPoolIds.size());
         for(Long expectedId: expectedPoolIds){
             assertTrue("Expected poolID "+ expectedId + " for worksetID " + worksetId + " but it was not in " + actualPoolIds, actualPoolIds.contains(expectedId));
         }


miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/PoolDto.java
```diff
@@ -150,7 +151,11 @@ public void setIdentificationBarcode(String identificationBarcode) {
     this.identificationBarcode = identificationBarcode;
   }
 
-  public void setLastModified(String lastModified) {
+    public void setWorksetAddedTime(String worksetAddedTime) {
+        this.worksetAddedTime = worksetAddedTime;
+    }
```
djcooke: Should also have a getter for consistency

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/workset/WorksetPool.java
```diff
@@ -0,0 +1,89 @@
...
+            return LimsUtils.equals(this, obj,
+                WorksetPoolId::getWorkset,
+                WorksetPoolId::getItem);
+        }
+    }
+    
+    private static final long serialVersionUID = 1L;
+
+    @Id
+    @ManyToOne
```
callunity, line 61: Does this need the `targetEntity` to be specified?

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/PoolDto.java
```diff
@@ -105,6 +106,8 @@ public Boolean getQcPassed() {
     return qcPassed;
   }
 
+  public String getWorsetAddedTime() { return worksetAddedTime; }
```
callunity: `getWorsetAddedTime` -> `getWorksetAddedTime`

sqlstore/src/it/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateListPoolViewDaoIT.java
```diff
@@ -0,0 +1,49 @@
...
+        sut.restrictPaginationByWorksetId(builder, worksetId, msg -> fail(msg));
+
+        List<ListPoolView> results = builder.getResultList();
+        assertEquals(worksetId, expectedCount, results.size());
+    }
+
+    @Test
+    public void testRestrictPaginationByWorksetId(){
+        assertWorksetHasPools(1L, 2);
+        assertWorksetHasPools(2L, 2);
```
callunity: I see in `V9010__test_data.sql` how you add 1 pool1 to workset 1 and 1 pool to workset 2 - can you help me understand how is it that this tests which assert 2 pools per workset is passing?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/workset/WorksetPool.java
riyapatel36: Thanks for the review Heather. 
I believe it is required as Pool is an interface. We are using Pool interface here for instantiate -  `private Pool item;`. And the same way we are doing it with Sample, Library and Library Aliquotes. 
I hope that's what you were asking.

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/PoolDto.java
riyapatel36: Thanks for pointing it out. I will update that.

sqlstore/src/it/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateListPoolViewDaoIT.java
riyapatel36: Thanks for pointing it out Heather. I have made a mistake here. It should be PoolID not the expected count I guess. I will take a look into this.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/impl/workset/WorksetPool.java
callunity: Makes sense, thanks!

sqlstore/src/it/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateListPoolViewDaoIT.java
```diff
@@ -32,18 +34,27 @@ public void setup() throws IOException {
...
         sut.restrictPaginationByWorksetId(builder, worksetId, msg -> fail(msg));
 
         List<ListPoolView> results = builder.getResultList();
         assertEquals(worksetId, expectedCount, results.size());
+        List<Long> actualPoolIds = new ArrayList<>();
+        for(ListPoolView row: results){
+            actualPoolIds.add(row.getId());
+        }
+
+        for(long expectedId: expectedPoolIds){
```
callunity: Since you're now passing in the `expectedPoolIds`, you could drop the `expectedCount` arg (since it's equal to the length if `expectedPoolIds`)

callunity (approved):

sqlstore/src/it/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateListPoolViewDaoIT.java
riyapatel36: Updated! Thank you Heather.

sqlstore/src/it/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateListPoolViewDaoIT.java
```diff
@@ -32,18 +34,26 @@ public void setup() throws IOException {
         sut.setEntityManager(entityManager);
     }
 
-    public void assertWorksetHasPools(long worksetId, int expectedCount) {
+    public void assertWorksetHasPools(long worksetId,long... expectedPoolIds) {
         QueryBuilder<ListPoolView, ListPoolView> builder = new QueryBuilder<>(currentSession(), ListPoolView.class, ListPoolView.class);
         sut.restrictPaginationByWorksetId(builder, worksetId, msg -> fail(msg));
 
         List<ListPoolView> results = builder.getResultList();
-        assertEquals(worksetId, expectedCount, results.size());
```
djcooke: You should still check the result size to ensure that there are no extra unexpected pools. You can check against the length of `expectedPoolIds` instead of having an additional parameter though.

sqlstore/src/it/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateListPoolViewDaoIT.java
```diff
@@ -32,18 +34,26 @@ public void setup() throws IOException {
         sut.setEntityManager(entityManager);
     }
 
-    public void assertWorksetHasPools(long worksetId, int expectedCount) {
+    public void assertWorksetHasPools(long worksetId,long... expectedPoolIds) {
         QueryBuilder<ListPoolView, ListPoolView> builder = new QueryBuilder<>(currentSession(), ListPoolView.class, ListPoolView.class);
         sut.restrictPaginationByWorksetId(builder, worksetId, msg -> fail(msg));
 
         List<ListPoolView> results = builder.getResultList();
```
djcooke: Don't use `QueryBuilder` or DAOs to retrieve data for testing against, as you want to limit a test's dependencies. If `QueryBuilder` has a bug, we don't want that to affect the accuracy of the test, for example. You can fetch the workset directly using the Hibernate session, and then check the pools that it contains:

`currentSession().get(Workset.class, worksetId);`

sqlstore/src/it/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateListPoolViewDaoIT.java
```diff
@@ -34,26 +36,35 @@ public void setup() throws IOException {
...
+
+        Workset workset = (Workset) currentSession().get(Workset.class, worksetId);
+
+        List<Long> expectedPoolIds = new ArrayList();
+        for(WorksetPool pool: workset.getWorksetPools()){
+            expectedPoolIds.add(pool.getItem().getId());
+        }
+        assertEquals(worksetId, expectedPoolIds.size(), results.size());
```
djcooke: You should use `currentSession().get...` to get `results` because you should not be using `QueryBuilder` in your test. You still need the `expectedPoolIds` parameter. As is, you are just retrieving the same workset from the database in two different ways and testing it against itself, which will always match. You need to test against the expected state, which you need to provide.

riyapatel36: Thanks Dillan, I've updated test case so it no longer use QueryBuilder. The test now loads the Workset directly using `currentSession.get..` extracts pools via `getWorksetPools()`. And also compare that with the `expectedPoolIds` which I have passed through parameter.

sqlstore/src/it/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateListPoolViewDaoIT.java
```diff
@@ -36,35 +36,23 @@ public void setup() throws IOException {
...
-        assertEquals(worksetId, expectedPoolIds.size(), results.size());
-
         List<Long> actualPoolIds = new ArrayList<>();
-        for(ListPoolView row: results){
-            actualPoolIds.add(row.getId());
+        for(WorksetPool wp: workset.getWorksetPools()){
+            actualPoolIds.add(wp.getItem().getId());
         }
-
+        assertEquals(worksetId, expectedPoolIds.length, actualPoolIds.size());
```
djcooke: Why is `worksetId` in here?

riyapatel36: I wanted to check pool length for mentioned worksetId, should I add a clear message with WorksetID ?

djcooke: If you want to include the workset ID in a message, please do add a clear message. The workset ID alone is unclear.

riyapatel36: Thanks Dillan, I will update that.

PR #3180, opened 2025-12-19
GLT-4404-matrix barcode assignments

riyapatel36:
Jira ticket - https://jira.oicr.on.ca/browse/GLT-4404

Matrix Barcode Assignment: Automatically assign scanned barcodes. 

image

### DIFF
--- a/changes/add_barcodeScan.md
+++ b/changes/add_barcodeScan.md
@@ -1,0 +1,1 @@
+User can scan an entire box at once to automatically assign and update barcodes for multiple positions. After scanning, they can review a visual summary with warnings and confirm or cancel the changes. 


integration-tools/src/main/java/uk/ac/bbsrc/tgac/miso/integration/BoxScannerConfigurer.java
```diff
@@ -13,12 +13,14 @@
...
 import uk.ac.bbsrc.tgac.miso.integration.dp5mirage.DP5MirageScanner;
 import uk.ac.bbsrc.tgac.miso.integration.util.IntegrationException;
+import uk.ac.bbsrc.tgac.miso.integration.mock.MockBoxScanner;
 import uk.ac.bbsrc.tgac.miso.integration.visionmate.VisionMateScanner;
 
 @Configuration
 public class BoxScannerConfigurer {
   private static final String VISIONMATE_SCANNER = "visionmate";
   private static final String DP5MIRAGE_SCANNER = "dp5mirage";
+  private static final String MOCK_SCANNER = "mock";
```
djcooke, line 23: Please add a note to the developer's guide explaining how to enable this and what it does

integration-tools/src/main/java/uk/ac/bbsrc/tgac/miso/integration/mock/MockBoxScanner.java
```diff
@@ -0,0 +1,156 @@
...
+    private final String host;
+    private final int port;
+    private final HttpClient httpClient;
+    private int rows = 8;
+    private int cols = 12;
+
+    public MockBoxScanner(String host, int port) {
+        this.host = host;
+        this.port = port;
+        this.httpClient = HttpClient.newBuilder().version(HttpClient.Version.HTTP_2).build();
```
djcooke: None of these are used for anything, so remove both the fields and constructor parameters. Make sure to note in the developer's guide that the host and port will be ignored.

integration-tools/src/main/java/uk/ac/bbsrc/tgac/miso/integration/mock/MockBoxScanner.java
```diff
@@ -0,0 +1,156 @@
...
+        public MockScan(int rows, int cols){
+            this.rows = rows;
+            this.cols = cols;
+            generateBarcodes();
+        }
+
+        private void generateBarcodes() {
+            for(int r = 0; r<rows; r++) {
+                for (int c = 1; c<= cols; c++){
+                    String pos = String.format("%c%02d", (char) ('A' + r), c);
```
djcooke: Use `BoxUtils#getPositionString`. Same in the `getBarcodes` methods below. Note that this will require a starting column of 0 instead of 1

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -63,6 +63,38 @@ var Box = Box || {
   },
   // Start the scanning process
   initScan: function (scannerName) {
+
+    var dialog = jQuery("#dialogDialog");
+    jQuery("#dialogInfoAbove").html("<h1>Choose Scan Mode</h1>");
+    jQuery("#dialogInfoBelow").html("<p>Do you want to update item location or assign Barcode?</p>");
```
djcooke:
```suggestion
    jQuery("#dialogInfoBelow").html("<p>Do you want to update item locations or assign barcodes?</p>");
```

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -63,6 +63,38 @@ var Box = Box || {
...
+            Cancel: function(){
+                dialog.dialog("close");
+            }
+        }
+    });
+  },
+
+  assignMatrixBarcodes: function (scannerName) {
+    Box.startMatrixScan(scannerName);
+  },
```
djcooke: Unused

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -63,6 +63,38 @@ var Box = Box || {
...
+
+  startStandardScan: function(scannerName){
```
djcooke: Unclear what "standard" means. `startUpdateLocationsScan`?

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -71,6 +103,40 @@ var Box = Box || {
     Box.prepareScannerDialog.show();
   },
 
+  startMatrixScan: function(scannerName){
```
djcooke: `startAssignBarcodesScan`? Probably best to avoid the word "matrix". We do use it in some places, referring to data matrix barcodes, to distinguish between indices and scannable barcodes, but I'm not sure how effective that is. It shouldn't be necessary for this feature either way though.

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -71,6 +103,40 @@ var Box = Box || {
...
+      var onPrepareSuccess = function () {
+
+        jQuery("#dialogDialog").dialog("close");
+        jQuery("#dialogDialog").dialog("destroy");
+
+        Box.scanDialog.show({size: Box.boxJSON, data: Box.boxJSON});
```
djcooke: The other call to `Box.scanDialog.show` uses this for options:

```
{
  size: {
    rows: Box.boxJSON.rows,
    cols: Box.boxJSON.cols,
  },
  data: Box.boxJSON.items,
}
```

Using `Box.boxJSON` directly could cause unexpected behaviour - especially for `data`

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -144,13 +210,17 @@ Box.scan = {
...
+            onSuccess();
+        } else {
+            Box.scanDialog.show({
+              size: {
+                 rows: Box.boxJSON.rows,
+                 cols: Box.boxJSON.cols,
+              },
+              data: Box.boxJSON.items,
+           });
+        }
```
djcooke: Rather than defaulting to the old behaviour, the old behaviour should be passed as an `onSuccess` function

miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -764,6 +772,162 @@
...
+            resizable: false,
+            position: [ jQuery(window).width()/2 - Box.dialogWidth /2, 50 ],
+            buttons: [
+                {
+                    text:  "Confirm Assign",
+                    click: function() {
+                        var url = Urls.rest.boxes.matrixAssign(Box.boxJSON.id);
+
+                        console.log("URL: ", url);
+                        console.log("Result: ", results);
```
djcooke: remove

miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -764,6 +772,162 @@
...
+            size: {
+                rows: Box.boxJSON.rows,
+                cols: Box.boxJSON.cols
+            },
+            data: self.results.items || []
+        });
+
+        jQuery("#updateSelected, #removeSelected, #emptySelected")
+            .prop("disable", true)
+            .addClass("disable");
```
djcooke: Should both be "disabled"

miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -764,6 +772,162 @@
...
+
+        var rowChar = Box.utils.getRowLabel(row);
+        var colNum = parseInt(Box.utils.getColLabel(col), 10);
+        var paddedPos = rowChar + (colNum < 10 ? "0" + colNum: colNum);
+        var unpaddedPos = rowChar + colNum;
+
+        var boxables = self.results.items
+            ? self.results.items.filter( function (item) {
+                if(!item.coordinates) return false;
+                return item.coordinates === pos || item.coordinates === paddedPos || item.coordinates === unpaddedPos;
```
djcooke: Why might this be in 3 different formats?

miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -764,6 +772,162 @@
...
+
+        var hasError = self.results.errors && self.results.errors.some( function(err) {
+            if(!err.coordinates) return false;
+
+            var errPos = err.coordinates;
+            return errPos === pos || errPos ===paddedPos || errPos === unpaddedPos;
+        })
+
+        if(boxables.length > 0){
+            var title = boxables[0].description || boxables[0].alias || boxables[0].identificationBarcode || "Item";
```
djcooke: description shouldn't be used as an identifier, and alias is required, so you can just use the alias here

miso-web/src/main/webapp/scripts/box_ajax.js
riyapatel36: Yes I meant `startUpdateLocationsScan`, I have updated that. Thank you.

integration-tools/src/main/java/uk/ac/bbsrc/tgac/miso/integration/mock/MockBoxScanner.java
riyapatel36: Thanks for reminder Dillan. Forgot to implement this.

miso-web/src/main/webapp/scripts/box_ajax.js
riyapatel36: Thanks Dillan for suggestion. I will update this.

miso-web/src/main/webapp/scripts/box_ajax.js
riyapatel36: I am using `onSuccess` as an optional param. So that it won't make an issue for old scanning behaviour. should I make `onSuccess` compulsory to check ?

djcooke: Yes. There's no clear reason or benefit to this default behaviour. For example, if we add another feature that uses this function, it's very unlikely that the default behaviour would be used. It's cleaner to remove the default behaviour and require the `onSuccess` parameter instead.

riyapatel36: Yes.. I will update that Dillan. Thanks.

miso-web/src/main/webapp/scripts/box_visualization.js
riyapatel36: I forgot to remove that unnecessary formats. There was some experimental code. Thanks for pointing it out Dillan. I will remove that.

integration-tools/src/main/java/uk/ac/bbsrc/tgac/miso/integration/mock/MockBoxScanner.java
```diff
@@ -71,14 +63,13 @@ public String getBarcode(String position){
 
         @Override
         public String getBarcode(char row, int column) {
-            String pos = String.format("%c%02", row, column + 1 );
-            return getBarcode(pos);
+            int rowIndex = Character.toUpperCase(row)-'A';
+            return getBarcode(BoxUtils.getPositionString(rowIndex, column));
```
djcooke: There is a `BoxUtils#getPositionString` method that takes `char` and `int`, so you don't need to do that conversion.

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -63,14 +63,96 @@ var Box = Box || {
...
+        Box.scanDialog.show({
+          size: {
+            rows: Box.boxJSON.rows,
+            cols: Box.boxJSON.cols,
+          },
+          data: Box.boxJSON.items,
+        });
+
+        jQuery.ajax({
+                url: Urls.rest.boxes.matrixScan(Box.boxJSON.id),
```
djcooke: * The dialog close+destroy and `Box.scanDialog.show` are in both functions, so those parts could just stay in `prepareScanner`
* `Box.scanDialog.show` ends up calling `Box.scan.scanBox`, which I don't think is wanted for the assign barcodes function... Right? I think that should get moved to the prepareSuccess function for update location scan
* Please avoid the word "matrix" on its own, including in `Urls` and the controller. "Matrix barcode" is better if necessary, but just "barcode" should be fine in most cases since it's not going to be confused with indices here

integration-tools/src/main/java/uk/ac/bbsrc/tgac/miso/integration/mock/MockBoxScanner.java
riyapatel36: Thanks for pointing it out Dillan. I will update that.

miso-web/src/main/webapp/scripts/box_ajax.js
riyapatel36: Thanks Dillan, agreed. I'll refactor the code . and also I will rename url to avoid "matrix".

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -63,14 +63,81 @@ var Box = Box || {
...
+    jQuery("#dialogInfoBelow").html("<p>Do you want to update item locations or assign barcodes?</p>");
+    jQuery("#dialogVisual").html("");
+    dialog.dialog({
+        autoOpen: true,
+        title: "Scan Options",
+        modal: true,
+        width: 400,
+        height: 200,
+        buttons: {
+            "Update Location": function(){
```
djcooke:
```suggestion
            "Update Locations": function(){
```

docs/user_manual/boxes.md
```diff
@@ -225,6 +228,28 @@ sources of error are:
  * Specific to Ziath DP5 Mirage Rack Reader 
     * Check if Container UID on the DP5 application has a container named: 'mirage96sbs' 
 
+## Matrix Barcode Assigning
```
djcooke: Restructure and update titles something like below.

## Scanning a Box

If your MISO is configured to use a box scanner, it can be used to update the entire box in one operation... (continue explanation with the stuff that is common to both operations)

### Updating Item Locations

(explain this option)

### Assigning Barcodes

(explain this option. remove/replace the word "matrix" with clearer terms

miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -764,6 +772,155 @@
     return self;
   };
 
+  Box.BarcodeScanDialog = function(scannerName) {
+    var self = new BoxVisual();
+
+    self.show = function (results) {
+        self.results = results;
+
+        var summaryHTML = "<h1> Matrix Scan Results </h1><ul>";
```
djcooke: Remove "Matrix"

miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -764,6 +772,155 @@
...
+            });
+            Warnings += '</ul>';
+        }
+
+        jQuery("#dialogInfoAbove").html(summaryHTML);
+        jQuery("#dialogInfoBelow").html(legendHTML + Warnings);
+
+        jQuery("#dialogDialog").dialog({
+            autoOpen: false,
+            title: "Scan Results (Assign Mode)",
```
djcooke: Unclear what "Assign Mode" means

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -116,26 +112,15 @@ var Box = Box || {
   startAssignBarcodesScan: function(scannerName){
       var onPrepareSuccess = function () {
 
-        jQuery("#dialogDialog").dialog("close");
-        jQuery("#dialogDialog").dialog("destroy");
-
-        Box.scanDialog.show({
```
djcooke: Is this still working? You removed this from here, but didn't change anything in box_visualization's `prepareScanner` function to make it happen there... Also, it's still included in `startUpdateLocationsScan`'s `onPrepareSuccess` function, and the point of this change was to remove it from both to avoid repetition.

miso-web/src/main/webapp/scripts/box_visualization.js
riyapatel36: Should I just keep "Scan Results" or "Barcode Scan Results" ?

miso-web/src/main/webapp/scripts/box_ajax.js
riyapatel36: I understand the purpose behind removing that from both scanning function, But I couldn't find a way to implement this. I need `onPrepareSuccess` in both scanning functions `startUpdateLocationsScan` and `startAssignBarcodesScan`. For `startAssignBarcodesScan` I have some things to implement but in `startUpdateLocationsScan` I don't have much to put. So I couldn't figure out how to remove the code lines you said and put that in `onPrepareSuccess`. And this way it works fine.

miso-web/src/main/webapp/scripts/box_visualization.js
djcooke: I think "Scan Results" is fine

docs/user_manual/boxes.md
```diff
@@ -189,22 +189,22 @@ The page will refresh and the box will be empty.
...
+in the list, your box may not be scannable or MISO may not be configured to use this box scanner.
+
+After clicking on "Scan with <scanner-name\>", you will see two scan options 1) Update Locations 2) Assign Barcodes.
 
-After clicking on "Scan with <scanner-name\>", you will see two scan options 1) Update Location 2) Assign Barcodes.
-Click on Update Location for scanning box.
+### Updating Item Locations
 
-Wait until the Scan dialog box prompts you to place the box on the scanner. It will automatically scan the box.
+To update Item locations click on "Update Locations" from scan options. Wait until the Scan dialog box prompts you to place the box on the scanner. It will automatically scan the box.
```
djcooke, line 212: Before explaining how to do it, describe what "Updating Item Locations" means.

docs/user_manual/boxes.md
```diff
@@ -215,36 +215,30 @@ Once a successful scan is made, review the changes that would be made and click
 If there are any errors with the scan, the scan will need to be retried until there are no errors. Possible
 sources of error are:
```
djcooke: Move this to a "Troubleshooting" section below both scan functions since it applies to both

miso-web/src/main/webapp/scripts/0-urls.js
```diff
@@ -174,6 +174,8 @@ Urls = (function () {
     fillByPattern: middleIdUrlFunction(boxRestBase, "/positions/fill-by-pattern"),
     prepareScan: boxRestBase + "/prepare-scan",
     scan: middleIdUrlFunction(boxRestBase, "/scan"),
+    barcodeScan: middleIdUrlFunction(boxRestBase, "/barcode-scan"),
+    barcodeAssign: middleIdUrlFunction(boxRestBase, "/barcode-scan/assign"),
```
djcooke:
```suggestion
updateLocationsScan: middleIdUrlFunction(boxRestBase, "/update-locations-scan"),
    assignBarcodesScan: middleIdUrlFunction(boxRestBase, "/assign-barcodes-scan"),
    assignBarcodesSave: middleIdUrlFunction(boxRestBase, "/assign-barcodes-save"),
```

Make sure to update the mappings on the controller as well

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -63,14 +63,81 @@ var Box = Box || {
...
+                jQuery("#dialogDialog").dialog("close");
+                var error = JSON.parse(response.responseText);
+                var message = error.detail ? error.detail : error.message;
+                Utils.showOkDialog("Scan Failed", [message]);
+              });
+      };
+
+      Box.dialogWidth = Box.boxJSON.cols * 40 + 150;
+      Box.dialogHeight = Box.boxJSON.rows * 40 + 300;
+      Box.scanDialog = Box.ScanDialog(scannerName);
```
djcooke: `scanDialog` isn't used for the assign barcodes function, so shouldn't be initialized here.

miso-web/src/main/webapp/scripts/box_ajax.js
```diff
@@ -144,13 +211,17 @@ Box.scan = {
...
-        Box.scanDialog.show({
-          size: {
-            rows: Box.boxJSON.rows,
-            cols: Box.boxJSON.cols,
-          },
-          data: Box.boxJSON.items,
-        });
+        if(onSuccess){
+            onSuccess();
+        } else {
```
djcooke: There is always an `onSuccess` function. Remove the `if` and `else` block, and always call `onSuccess()`

miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -732,7 +732,15 @@
...
-      Box.scan.prepareScanner(scannerName, Box.boxJSON.rows, Box.boxJSON.cols);
+      Box.scan.prepareScanner(
+        scannerName,
+        Box.boxJSON.rows,
+        Box.boxJSON.cols,
+        function() {
+            if(typeof onSuccess === "function"){
+                onSuccess();
+            }
+        });
```
djcooke: Just pass along `onSuccess`. There's no need to wrap it in another function

miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -764,6 +772,155 @@
     return self;
   };
 
+  Box.BarcodeScanDialog = function(scannerName) {
```
djcooke:
```suggestion
  Box.AssignBarcodesScanDialog = function(scannerName) {
```

Make sure to update references. Please also rename the following things similarly:

- Box.barcodeScanDialog -> Box.assignBarcodesScanDialog
- Box.scanDialog -> Box.updateLocationsScanDialog
- Box.ScanDialog -> Box.UpdateLocationsScanDialog

miso-web/src/main/webapp/scripts/box_ajax.js
djcooke: I think the way you've fixed it is actually better than what I previously suggested because you were previously calling `Box.scanDialog.show` in a place where you shouldn't. I've suggested a couple minor fixes and some renaming that will make it clearer which functions and endpoints should be used for what, so it'll be easier to follow.

riyapatel36: Thanks Dillan for detailed review. So for this part I should keep it the way it is right now. and I will update the changes you suggested below.

miso-web/src/main/webapp/scripts/box_visualization.js
riyapatel36: By keeping this way Dillan it makes smoother transition in UI between prepare scanning and scan results. I removed and checked that the transition in UI is too quick and not smooth like before. I am not removing it for now. you can check and let me know if we still want to remove this.

djcooke: It shouldn't make a difference to the transition. Did you try using `onSuccess()` as the parameter? That would cause the function to run prematurely. It should be passed as `onSuccess` - just the reference without `()`.

riyapatel36: Thank you Dillan. It is working now.

PR #3181, opened 2025-12-22
GLT-4721-Attach files to Library Aliquots

riyapatel36:
Jira ticket - https://jira.oicr.on.ca/browse/GLT-4721
Attach Files to Library Aliquot


djcooke:
Change files are used to create the release notes during a release. More info here: https://github.com/miso-lims/miso-lims/blob/develop/changes/README.md

changes/add_attachFiles_libraryAliquot.md
djcooke, line 1: Attaching files to library aliquots is a new feature, so should be an `add_` type of change file rather than `change_` (for existing features)

PR #3197, opened 2026-01-29
[GLT-4367] added probe sets

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4367

- [x] Includes a change file

Note this is being merged into the other probes branch, which is still in PR

### DIFF
--- a/docs/user_manual/samples.md
+++ b/docs/user_manual/samples.md
@@ -190,29 +190,29 @@ Any number of quality control measures may be recorded for a sample. For more in
 
 ## Sample Probes
 
-Sample probes were mainly designed to represent 10X feature barcodes. They can be added to any
-tissue processing sample, and are available in detailed sample mode only. Probes are added directly
-to samples, but may also be saved as probe sets. Probe sets are saved collections of probes that
-can easily be applied to additional samples.
+Sample probes were mainly designed to represent 10X feature barcodes. They can be added to single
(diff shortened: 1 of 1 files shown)


docs/user_manual/samples.md
```diff
@@ -188,6 +188,41 @@ Any number of quality control measures may be recorded for a sample. For more in
...
+tissue processing sample, and are available in detailed sample mode only. Probes are added directly
+to samples, but may also be saved as probe sets. Probe sets are saved collections of probes that
+can easily be applied to additional samples.
+
+### Editing Probes in Bulk
+
+When working with tissue processing samples on the bulk Create/Edit/Propagate Samples pages, an
+"Edit Probes" button will appear above the table. This will allow you to select which samples you
+wish to edit the probes for, and whether you'd like to apply a probe set or define a new custom set
+of probes. When first adding probes, you can choose multiple samples to add the same probes to all.
```
callunity: `you can choose multiple samples to add the same probes to all` -> `you can select multiple samples and add the same probes to each sample` ?

djcooke: The original wording was awkward, but your suggestion makes it sound like adding to each sample is a separate action. Changed to this, which I think is clearer:

> When first adding probes, you can select multiple samples and the same probes will be added to each sample.

PR #3198, opened 2026-01-30
GLT-4734-Fixed-Problem-loading-EditArray-Page

riyapatel36:
Jira ticket - https://jira.oicr.on.ca/browse/GLT-4734
Big Fixed - Server error while adding array run

### DIFF
--- a/changes/Fix_editArrayLoading.md
+++ b/changes/Fix_editArrayLoading.md
@@ -1,1 +1,1 @@
-Problem while loading Edit array page if no arrays attached.
+Server Error while loading Edit Array Run page if no arrays attached.


changes/Fix_editArrayLoading.md
```diff
@@ -0,0 +1 @@
+Problem while loading Edit array page if no arrays attached.
```
djcooke:
```suggestion
Server Error while loading Edit Array Run page if no arrays attached.
```

- It's good to include the error name/message so that users can better identify what's being fixed if they've seen the error
- It's the Edit Array Run page - not Edit Array
- Title case for page titles

PR #3199, opened 2026-02-03
GLT-4733-Fixed-Update position sample issue in Array

riyapatel36:
Fixed : Update position on sample in Array to show clear message.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -590,7 +590,7 @@
         <plugin>
           <groupId>io.fabric8</groupId>
           <artifactId>docker-maven-plugin</artifactId>
-          <version>0.44.0</version>
+          <version>0.48.0</version>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>


miso-web/src/main/webapp/scripts/array_ajax.js
```diff
@@ -171,27 +171,32 @@
...
+      },
+      function (xhr, textStatus, errorThrown) {
+        var message = "Request Failed";
+        try {
+            var error = xhr.responseJSON || JSON.parse(xhr.responseText);
+            message = error.detail ? error.detail : error.message;
+        } catch (e) {
+            message = errorThrown || textStatus || message;
+        }
+        $("#warningMessages").html("Error adding item: " + message);
```
djcooke: This doesn't seem to be working - the message isn't displayed. The error message is already being shown in dialog before this though, so this isn't really needed.

djcooke: Doesn't work because `showAjaxErrorDialog` doesn't pass its parameters on to `errorCallback` when handling off to `showOkDialog`

PR #3200, opened 2026-02-04
GLT-4732-Add-Qcstatus-Array Run

riyapatel36:
Jira ticket - https://jira.oicr.on.ca/browse/GLT-4732

Added Qc Status in Array Run Information

### DIFF
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ArrayRun.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ArrayRun.java
@@ -69,6 +69,14 @@ public class ArrayRun implements Serializable, Aliasable, Attachable, ChangeLogg
       @JoinColumn(name = "attachmentId")})
   private List<FileAttachment> attachments;
 
+  private Boolean qcPassed;
+
+  @ManyToOne(targetEntity = UserImpl.class)
+  @JoinColumn(name = "qcUser")
+  private User qcUser;
+
+  private LocalDate qcDate;
+
   @Transient
   private List<FileAttachment> pendingAttachmentDeletions;
 
@@ -189,6 +197,30 @@ public class ArrayRun implements Serializable, Aliasable, Attachable, ChangeLogg
   @Override
   public void setPendingAttachmentDeletions(List<FileAttachment> pendingAttachmentDeletions) {
     this.pendingAttachmentDeletions = pendingAttachmentDeletions;
+  }
+
+  public Boolean getQcPassed() {
+      return qcPassed;
+  }
+
+  public void setQcPassed(Boolean qcPassed){
+      this.qcPassed = qcPassed;
+  }
+
+  public User getQcUser() {
+      return qcUser;
+  }
+
+  public void setQcUser(User qcUser){
+      this.qcUser = qcUser;
+  }
+
+  public LocalDate getQcDate(){
+      return qcDate;
+  }
+
+  public void setQcDate(LocalDate qcDate){
+      this.qcDate = qcDate;
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ArrayRun.java
```diff
@@ -69,6 +69,14 @@ public class ArrayRun implements Serializable, Aliasable, Attachable, ChangeLogg
       @JoinColumn(name = "attachmentId")})
   private List<FileAttachment> attachments;
 
+  private Boolean qcPassed;
+
+  @ManyToOne(targetEntity = UserImpl.class)
+  @JoinColumn(name = "qcUser")
+  private User qcUser;
+
```
Nandu-26, line 77: Can we set @ManyToOne(fetch = FetchType.LAZY) on qcUser for consistency with other User relationships and to avoid eager-loading in list views?

sqlstore/src/main/resources/db/migration/V8090__add_QcColumArrayRun.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE ArrayRun ADD COLUMN qcPassed BOOLEAN;
```
Nandu-26, line 1: A minute typo error- should be sqlstore/src/main/resources/db/migration/V8090__add_QcColumnArrayRun.sql

sqlstore/src/main/resources/db/migration/V8090__add_QcColumArrayRun.sql
```diff
@@ -0,0 +1,4 @@
...
+ALTER TABLE ArrayRun ADD COLUMN qcUser bigint;
```
Nandu-26, line 2: Should we add an index on ArrayRun.qcUser in the migration? it's a FK column and likely to be joined/filtered.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ArrayRun.java
riyapatel36: I have made it similar to what we have in `Run.java`.  @djcooke Do I need to update this ?

sqlstore/src/main/resources/db/migration/V8090__add_QcColumArrayRun.sql
riyapatel36: Similar to this one, I have tried to make it similar to `Run`. we don't explicitly add indexing in Run on `qcUser`. Mysql is creating indexes automatically for certain constraints. But I am not sure if we have to add Indexes here or not.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ArrayRun.java
djcooke: It's not an expensive join and the list page is paginated, so I don't think there's enough benefit to make it lazy

sqlstore/src/main/resources/db/migration/V8090__add_QcColumArrayRun.sql
djcooke: MySQL automatically indexes foreign key columns, so you should not

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ArrayRun.java
djcooke: Also, what user relationships are lazy-loaded? I just checked `creator`, `lastModifier`, and `qcUser` on `Run`, `SampleImpl`, and `LibraryImpl` and none of them are.

sqlstore/src/main/resources/db/migration/V8090__add_QcColumArrayRun.sql
riyapatel36: The name of the file will be changed into release migration. so is that okay if I leave this as it is ?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ArrayRun.java
Nandu-26: uh, my bad you're right I double checked existing models too and they are not lazy-loaded.

sqlstore/src/main/resources/db/migration/V8090__add_QcColumArrayRun.sql
djcooke: Yeah, the filename is temporary and doesn't really matter

PR #3204, opened 2026-02-20
Gp-5194 Barcode Assignment loading issue

riyapatel36:
Jira - https://jira.oicr.on.ca/browse/GP-5194

- Scannable box -> Options ->  Scan with DD Extraction Scanner ->  Select Assign Barcode -> Preparing scanner dialog -> Place box Dialog (Magnify glass) -> scan result -> Confirm , rescan, cancel 

- Adde Rescan option

### DIFF
--- a/changes/fix_barcode_assignment.md
+++ b/changes/fix_barcode_assignment.md
@@ -1,1 +1,1 @@
-Matrix Barcode Assignment UI glitches has been fixed.
+Incorrect scanner error shown before scanner properly scanned box for barcode assignment.
--- a/changes/fix_barcode_assignment.md
+++ b/changes/fix_barcode_assignment.md
@@ -1,0 +1,1 @@
+Matrix Barcode Assignment UI glitches has been fixed.
--- a/miso-web/src/main/webapp/scripts/box_visualization.js
+++ b/miso-web/src/main/webapp/scripts/box_visualization.js
@@ -945,6 +945,12 @@
                     }
                 },
                 {
+                    text: "Rescan",
+                    click: function () {
+                        Box.initScan(scannerName);
+                    }
+                },
+                {
                     text: "Cancel",
                     click: function() {
                         jQuery("#dialogDialog").dialog("close");
--- a/miso-web/src/main/webapp/scripts/box_visualization.js
+++ b/miso-web/src/main/webapp/scripts/box_visualization.js
@@ -565,6 +565,123 @@
     return self;
   };
 
+  Box.AssignBarcodesScanProgressDialog = function (scannerName) {
+    var self = new BoxVisual();
+
+    self.show = function (opts) {
+      self.size = opts.size;
+      self.data = opts.data;
+
+      self.getNewPosition = function () {
+        var h = Box.boxJSON.rows * 30 - 100;
+        var w = Box.dialogWidth - 400;
+        return [Math.floor(Math.random() * h) + 100, Math.floor(Math.random() * w) + 100];
+      };
+
+      self.animateMag = function () {
+        var newpos = self.getNewPosition();
+        var oldpos = jQuery("#magnify").offset();
+        var speed = self.getSpeed([oldpos.top, oldpos.left], newpos);
+        jQuery("#magnify").animate(
+          {
+            top: newpos[0],
+            left: newpos[1],
+          },
+          speed,
+          function () {
+            self.animateMag();
+          }
+        );
+      };
+
+      self.getSpeed = function (prev, next) {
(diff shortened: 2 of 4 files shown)


miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -925,6 +946,33 @@
         }
     };
 
+    self.error = function () {
```
callunity: This looks to be identical to the `self.error` function in `Box.prepareScannerDialog`. Is there a way to extract this function or make it more general?

miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -448,14 +448,29 @@
...
+       scanFn: function (name) {
+           Box.scan.scanBox(name);
+       }
+    });
+  };
+
+  Box.AssignBarcodesScanProgressDialog = function (scannerName) {
+    return Box.ScanProgressDialog(scannerName, {
+        scanFn: function (name) {
+            Box.scan.scanAssignBarcodes(name);
```
callunity: if the arg to `Box.scan.scanAssignBarcodes` will always be `scannerName`, then you can rewrite this to encapsulate it:
```
return Box.ScanProgressDialog(scannerName, {
        scanFn: function () {
            Box.scan.scanAssignBarcodes(scannerName);
        }
    });
```
and then `params.scanFn()` will already have the scannerName stored in the function.
(same applies to `Box.scan.scanBox` above)

miso-web/src/main/webapp/scripts/box_visualization.js
riyapatel36: I think I can remove this `self.error` function. I wanted `Box.AssignBarcodesScanDialog`and `Box.Scandiff`to be similar pattern.

miso-web/src/main/webapp/scripts/box_visualization.js
riyapatel36: Agree. Thanks heather I will update that.

miso-web/src/main/webapp/scripts/box_visualization.js
callunity: It makes sense to have them be a similar pattern, and I do wonder about removing it entirely. Did you look at when the `Box.prepareScannerDialog` `error()` function gets called, and confirmed that `Box.AssignBarcodesScanDialog().error()` can never get called?

miso-web/src/main/webapp/scripts/box_visualization.js
```diff
@@ -448,14 +448,29 @@
...
+
+  Box.AssignBarcodesScanProgressDialog = function (scannerName) {
+    return Box.ScanProgressDialog(scannerName, {
+        scanFn: function () {
+            Box.scan.scanAssignBarcodes(scannerName);
+        }
+    });
+  };
+
+  Box.ScanProgressDialog = function (scannerName, params) {
```
djcooke: The only thing used in `params` is `scanFn`, so `scanFn` could be the parameter instead

djcooke (commented):
Needs a change file

miso-web/src/main/webapp/scripts/box_visualization.js
riyapatel36: The shared `Box.ScanProgressDialog` already includes error handling for both scan types. The `Box.AssignBarcodesScanDialog()` and `Box.ScanDiff()` are only responsible for showing results after a successful scan. so I don't think we need duplicate error handling in the result dialogs.

callunity (approved):

changes/fix_barcode_assignment.md
```diff
@@ -0,0 +1 @@
+Matrix Barcode Assignment UI glitches has been fixed.
```
djcooke: * Not obvious this is talking about the scanner feature. People might worry there were issues displaying barcodes in other places, which could be a big problem if used to identify things
* This will appear under a "Fixed" heading, so "has been fixed" isn't necessary
* Avoid the word "matrix"

Something like "Display glitches when using the box scanner to assign barcodes" would be good, though describing the "glitch" is more helpful. I don't remember what was actually happening. The dialog wasn't showing up correctly?

riyapatel36: The issue was before getting the results, it was showing Error: could not find the scanner and then it shows the scan result. 

"Incorrect scanner error shown before scanner properly scanned box for barcode assignment" Is this looks good ?

djcooke (approved):
Approved once change message is updated

PR #3212, opened 2026-03-10
GLT-4722-Add-Qc-Status-Array-Run-Samples

riyapatel36:
Jira ticket -  https://jira.oicr.on.ca/browse/GLT-4722

Add Qc status to Array run samples


sqlstore/src/main/resources/db/migration/V8010__arrayrun_sample.sql
```diff
@@ -0,0 +1,18 @@
...
+    sampleId bigint NOT NULL,
+    statusId bigint,
+    qcNote varchar(255),
+    qcUser bigint,
+    qcDate date,
+    lastModifier bigint NOT NULL,
+    PRIMARY KEY (arrayRunId, position),
+    CONSTRAINT fk_arrayRunSample_arrayRun FOREIGN KEY (arrayRunId) REFERENCES ArrayRun (arrayRunId),
+    CONSTRAINT fk_arrayRunSample_array FOREIGN KEY (arrayId) REFERENCES Array (arrayId),
+    CONSTRAINT fk_arrayRunSample_sample FOREIGN KEY (sampleId) REFERENCES Sample (sampleId),
```
djcooke: What do you think about creating 2 composite foreign keys instead?

- ArrayRun (arrayRunId, arrayId)
- ArrayPosition (arrayId, position, sampleId)

Then the database will enforce all of the relationships so it's impossible to store an invalid ArrayRun_Sample.

sqlstore/src/main/resources/db/migration/V8020__insert_arrayrunsample.sql
```diff
@@ -0,0 +1,19 @@
...
+    NULL,
+    ar.lastModifier
+FROM ArrayRun ar
+JOIN ArrayPosition ap on ap.arrayId = ar.arrayId
+WHERE ar.arrayId IS NOT NULL
+    AND NOT EXISTS (
+        SELECT 1 FROM ArrayRun_Sample ars
+        WHERE ars.arrayRunId = ar.arrayRunId
+        AND ars.position = ap.position
+    );
```
djcooke: `WHERE` clause is unnecessary - `arrayId` can't be null, and you just created this table, so nothing exists in it.

This also doesn't have to be in a separate migration from the table creation, though it doesn't really matter. They'll get combined into one when we do the release either way.

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/ArrayRunSampleDto.java
```diff
@@ -0,0 +1,96 @@
+package uk.ac.bbsrc.tgac.miso.dto;
+
+public class ArrayRunSampleDto {
+
+    private Long id;
```
djcooke, line 6: Please add a comment similar to `RunPartitionAliquotDto` explaining the purpose of this and that it is not a database ID

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+        Array array = run.getArray();
+        List<ArrayRunSample> existing = arrayRunSampleDao.listByRunId(arrayRunId);
+        Map<String, Sample> samples = array.getSamples();
+        if(samples == null || samples.isEmpty()){
+            List<ArrayRunSample> filtered = existing.stream()
+                    .filter(item -> item.getArray() != null && item.getArray().getId() == array.getId())
+                    .collect(Collectors.toList());
+
+            return filtered;
```
djcooke: If the array has no samples, how is `existing` not null/empty? That would mean invalid data in the database, right? And then `filtered` will contain the `ArrayRunSample`s that are linked to the array, but they wouldn't be linked to any samples in that array (because they are none), so again invalid?

sqlstore/src/main/resources/db/migration/V8010__arrayrun_sample.sql
```diff
@@ -0,0 +1,18 @@
...
+    arrayRunId bigint NOT NULL,
+    position varchar(6) NOT NULL,
+    arrayId bigint NOT NULL,
+    sampleId bigint NOT NULL,
+    statusId bigint,
+    qcNote varchar(255),
+    qcUser bigint,
+    qcDate date,
+    lastModifier bigint NOT NULL,
+    PRIMARY KEY (arrayRunId, position),
```
djcooke: I think the primary key should also include the `arrayId` and `sampleId`. These are really part of the identity of the record and it's a bit awkward in code how they are treated differently.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -0,0 +1,141 @@
...
+        for(Map.Entry<String, Sample> entry : samples.entrySet()) {
+            ArrayRunSample item = byPosition.get(entry.getKey());
+            if(item ==null) {
+                item = new ArrayRunSample(run, entry.getKey());
+            }
+            item.setArray(array);
+            item.setSample(entry.getValue());
+            results.add(item);
+
+        }
```
djcooke: These should be constructed in the DAO like in the `get` method

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public ArrayRunSample get(ArrayRun run, String position) throws IOException {
+        ArrayRunSampleId id = new ArrayRunSampleId();
+        id.setArrayRun(run);
+        id.setPosition(position);
+        ArrayRunSample result = (ArrayRunSample) currentSession().get(ArrayRunSample.class, id);
+        if(result == null){
+            result = new ArrayRunSample(run, position);
```
djcooke: Make sure this is valid

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/ArrayRunSampleService.java
```diff
@@ -0,0 +1,18 @@
+package uk.ac.bbsrc.tgac.miso.core.service;
+
+import java.io.IOException;
+import java.util.List;
+
+import uk.ac.bbsrc.tgac.miso.core.data.ArrayRun;
+import uk.ac.bbsrc.tgac.miso.core.data.ArrayRunSample;
+
+public interface ArrayRunSampleService {
```
djcooke, line 11: There are no delete methods. How are ArrayRunSamples cleaned up when the following occur?

- the array run is deleted
- the array is removed from the array run
- the array is deleted
- the sample is removed from the array position
- the sample is deleted

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -0,0 +1,141 @@
...
+            save(sample);
+        }
+    }
+
+    @Override
+    public void save(ArrayRunSample arrayRunSample) throws IOException {
+        ArrayRun run = arrayRunService.get(arrayRunSample.getArrayRun().getId());
+        if(run == null){
+            throw new ValidationException(new ValidationError("arrayRunId", "Invallid Array Run"));
+        }
```
djcooke: *Invalid. Could use `loadChildEntity` here instead though

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -0,0 +1,141 @@
...
+        if(run.getArray() == null) {
+            throw new ValidationException(new ValidationError("arrayRunId", "Array Run has no array"));
```
djcooke: This is a problem, but you're validating it from the wrong angle. You should be validating what the client submitted - `arrayRunSample`. Ensure that:

- the array run in `arrayRunSample` exists
- the array in `arrayRunSample` is the same array that is actually attached to the array run
- the position in `arrayRunSample` exists in the array
- the sample in `arrayRunSample` is in that position in the array

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+        Sample arraySample;
+
+        try{
+            arraySample = run.getArray().getSample(arrayRunSample.getPosition());
+        } catch (IllegalArgumentException | IndexOutOfBoundsException e ){
+            throw new ValidationException(new ValidationError("position", "Invalid array position"));
+        }
```
djcooke, line 153: Don't rely on exceptions that you could prevent by doing a check beforehand

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -0,0 +1,141 @@
...
+        List<ValidationError> errors = new ArrayList<>();
+        validateQcUser(arrayRunSample.getQcStatus(), arrayRunSample.getQcUser(), errors);
+        if(!errors.isEmpty()) {
+            throw new ValidationException(errors);
+        }
+
+        managed.setArrayRun(run);
+        managed.setPosition(arrayRunSample.getPosition());
+        managed.setArray(arrayRunSample.getArray());
+        managed.setSample(arrayRunSample.getSample());
```
djcooke: It should never be necessary to update these because `managed` shouldn't exist with an invalid configuration, and a valid configuration should be loaded when it doesn't exist.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRunRestController.java
```diff
@@ -129,6 +135,27 @@ public DataTablesResponseDto<ArrayRunDto> dataTableByProject(@PathVariable("id")
...
+  @GetMapping(value = "/{arrayRunId}/samples", produces = "application/json")
+  public @ResponseBody List<ArrayRunSampleDto> listSamples(@PathVariable long arrayRunId) throws IOException {
+      RestUtils.retrieve("Array Run", arrayRunId, arrayRunService);
+      List<ArrayRunSampleDto> dtos = arrayRunSampleService.listByRunId(arrayRunId).stream()
+              .map(Dtos::asDto)
+              .collect(Collectors.toList());
+
+      for(int i=0; i< dtos.size(); i++) {
+          dtos.get(i).setId((long) i);
+      }
```
djcooke: Please add a comment to explain

miso-web/src/main/webapp/scripts/array_ajax.js
```diff
@@ -375,3 +375,81 @@
       });
   }
 })((window.SampleArray = window.SampleArray || {}), jQuery);
+
+var ArrayRunSamples = (function ($) {
```
djcooke: - Doesn't need its own namespace
- You're essentially replacing `SampleArray.updateSamplesTable`, but you didn't remove that function, and it's still being called in one place, which will cause issues
- Below is much different from the ways we normally handle these tables. You made the data available via the REST API instead of including it in the model for the Edit page request, which is a good choice. That can be used via the `createUrl` in `list_arrayrunsample.js`and the table can be added to the JSP using the `miso:list-section-ajax` tag

miso-web/src/main/webapp/scripts/list_arrayrunsample.js
```diff
@@ -0,0 +1,116 @@
...
+                mData: "sampleName",
+                mRender: function (data, type, full) {
+                    if(type === "display") {
+                        return Box.utils.hyperlinkifyBoxable(full.sampleName, full.sampleId, full.sampleName);
+                    }
+                    return data;
+                },
+            },
+            {
+                sTitle: "Sample Alias",
```
djcooke: I realize this is the same as the existing table, but it's a list of samples, so including "Sample" is redundant. Can rename these columns to just "Name" and "Alias."

Could also simplify by using `ListUtils.labelHyperlinkColumn` for both.

miso-web/src/main/webapp/scripts/list_arrayrunsample.js
```diff
@@ -0,0 +1,116 @@
+ListTarget.arrayrunsample = {
+    name: "Array Run Samples",
```
djcooke: formatting doesn't match the rest of MISO (2 space indents). Same with some other JS in this PR.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
riyapatel36: Yes..I’ll update this so it returns an empty list only when both sides are empty, and otherwise throws instead of filtering and returning invalid rows.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
riyapatel36: Could you clarify which exception paths I can replace with upfront checks here?

miso-web/src/main/webapp/scripts/array_ajax.js
riyapatel36: Agreed, I will update that accordingly.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/ArrayRunSampleService.java
riyapatel36: Yes.. Sorry Dillan, I will add delete method.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
djcooke: See `Array#isPositionValid`

sqlstore/src/main/resources/db/migration/V8010__arrayrun_sample.sql
riyapatel36: I agree that composite foreign keys would be the stronger design here. I will try to update the Java identity mapping for ArrayRunSample accordingly

sqlstore/src/main/resources/db/migration/V8010__arrayrun_sample.sql
```diff
@@ -1,18 +1,48 @@
...
+  PRIMARY KEY (arrayRunId, arrayId, position, sampleId),
+  CONSTRAINT fk_arrayRunSample_arrayRun FOREIGN KEY (arrayRunId, arrayId) REFERENCES ArrayRun (arrayRunId, arrayId),
+  CONSTRAINT fk_arrayRunSample_arrayPosition FOREIGN KEY (arrayId, position, sampleId)
+    REFERENCES ArrayPosition (arrayId, position, sampleId),
+  CONSTRAINT fk_arrayRunSample_status FOREIGN KEY (statusId) REFERENCES RunItemQcStatus (statusId),
+  CONSTRAINT fk_arrayRunSample_qcUser FOREIGN KEY (qcUser) REFERENCES User (userId),
+  CONSTRAINT fk_arrayRunSample_lastModifier FOREIGN KEY (lastModifier) REFERENCES User (userId)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+INSERT INTO ArrayRun_Sample (
```
djcooke: There's actually no need to create these records, is there?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ArrayRunSample.java
djcooke, line 20: Should include the array and sample to match the database primary key

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -26,116 +30,121 @@
...
-        List<ValidationError> errors = new ArrayList<>();
-        validateQcUser(arrayRunSample.getQcStatus(), arrayRunSample.getQcUser(), errors);
-        if(!errors.isEmpty()) {
-            throw new ValidationException(errors);
-        }
+    String position = arrayRunSample.getPosition();
+    Sample expectedSample = array.getSample(position);
+    ArrayRunSample managed = arrayRunSampleDao.get(run, position);
+    managed.setArray(array);
+    managed.setSample(expectedSample);
```
djcooke: Array and sample should never be updated for an existing record

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -0,0 +1,150 @@
...
+
+  @Override
+  public ArrayRunSample get(ArrayRun run, String position) throws IOException {
+    return arrayRunSampleDao.get(run, position);
+  }
+
+  @Override
+  public List<ArrayRunSample> listByRunId(long arrayRunId) throws IOException {
+    ArrayRun run = arrayRunService.get(arrayRunId);
+    if (run == null || run.getArray() == null) {
```
djcooke: Should probably be an `IllegalArgumentException` if the run doesn't exist - that should be verified at the controller level so you can return 400 or 404 error if appropriate

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunDao.java
```diff
@@ -37,6 +37,20 @@ public HibernateArrayRunDao() {
     super(ArrayRun.class);
   }
 
+  @Override
+  public long create(ArrayRun object) throws IOException {
+    long id = super.create(object);
+    currentSession().flush();
```
djcooke: Are these flushes necessary? Please explain why in a comment if so.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -0,0 +1,150 @@
...
+      return Collections.emptyList();
+    }
+
+    Array array = run.getArray();
+    List<ArrayRunSample> existing = arrayRunSampleDao.listByRunId(arrayRunId);
+    Map<String, Sample> samples = array.getSamples();
+    if (samples == null || samples.isEmpty()) {
+      if (!existing.isEmpty()) {
+        throw new IllegalStateException(
+            String.format("Array run %d has saved samples, but array %d has no samples", arrayRunId, array.getId()));
```
djcooke: I think this is impossible? You shouldn't be validating data as it's loaded from the database anyway.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
djcooke: Bump

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
djcooke: Bump. Should return null if the run, pos, array, sample combination is not valid (array and sample should be included in `id` after requested change to `ArrayRunSampleId`

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/ArrayRunSampleService.java
djcooke: Add service-level methods for the deletes so that other services (`DefaultArrayRunService`) don't need to have direct access to the `ArrayRunSampleDao`, as that can easily result in accidentally bypassing business logic.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -0,0 +1,150 @@
...
+
+  @Override
+  public void save(List<ArrayRunSample> arrayRunSamples) throws IOException {
+    for (ArrayRunSample sample : arrayRunSamples) {
+      save(sample);
+    }
+  }
+
+  @Override
+  public void save(ArrayRunSample arrayRunSample) throws IOException {
```
djcooke, line 81: Please separate the different parts of the save process into other methods consistent with other services in MISO (loading children, validation, updating fields)

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
djcooke: Missed:

* the array in arrayRunSample is the same array that is actually attached to the array run

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -0,0 +1,150 @@
...
+    if (array == null) {
+      errors.add(new ValidationError("arrayRunId", "Array Run has no array"));
+    } else {
+      String position = arrayRunSample.getPosition();
+      if (position == null || !array.isPositionValid(position)) {
+        errors.add(new ValidationError("position", "Invalid array position"));
+      } else {
+        Sample expectedSample = array.getSample(position);
+        if (expectedSample == null) {
+          errors.add(new ValidationError("position", "No Sample at this position"));
```
djcooke: "Sample" does not need to be capitalized. Same with "Run" in "Array Run" above.

miso-web/src/main/webapp/scripts/array_ajax.js
```diff
@@ -374,4 +374,63 @@
...
+  function removeRunSampleTableChrome(node) {
+    var previous = node.prev();
+    while (previous.length && (previous.hasClass("ui-toolbar") || previous.hasClass("parsley-error"))) {
+      var toRemove = previous;
+      previous = previous.prev();
+      toRemove.remove();
+    }
+  }
+
+  SampleArray.updateRunSamplesTable = function (array, arrayRunId) {
```
djcooke: Doesn't look like this is used anywhere, and you're reloading the page whenever the contents change, so there shouldn't be any need, right?

The old `SampleArray.updateSamplesTable` function should also be removed since it is no longer used.

sqlstore/src/main/resources/db/migration/V8010__arrayrun_sample.sql
riyapatel36: There will be an empty table so we won't get any samples to show in array run.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ArrayRunSample.java
riyapatel36: I originally updated the Java @IdClass to mirror the DB primary key, but based on your later clarification I’ve kept that enforcement at the database level only. I left the Hibernate identity as arrayRun + position.. array and sample are still treated as normal relationship fields in the service/save flow, and mirroring them into the Java identity was causing problems without adding anything beyond the DB constraints. Should I try again to implement this again ?

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunDao.java
riyapatel36: Thanks Dillan, I will add comments here.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
riyapatel36: Agree. I’ll separate the two cases so a missing run throws IllegalArgumentException, while an existing run with no array still returns an empty list.

miso-web/src/main/webapp/scripts/array_ajax.js
riyapatel36: `updateRunSamplesTable` is no longer used, so I’ll remove that. 
`updateSamplesTable` is still used on the Array page itself via updatePage so I’m keeping that one.

sqlstore/src/main/resources/db/migration/V8010__arrayrun_sample.sql
djcooke: Missing records are constructed by the DAO (and currently the service layer) when needed. Otherwise you would also have to be inserting these records whenever a new array is attached to an array run.

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/ArrayRunSample.java
djcooke: Our conversation was about the foreign key changes - not the primary key. The class is called ArrayRunSample; the sample should definitely be part of its identity (and the array should too).

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
djcooke: What I meant here is that the DAO should fill in the blanks - you shouldn't need to check in the service layer if they exist. `ArrayRunSampleDao#get` returns the `ArrayRunSample` whether the database record exists or not. `ArrayRunSampleDao#listByRunId` should work the same - it should return all of the appropriate `ArrayRunSample`s whether the database records exist or not.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
```diff
@@ -29,13 +31,15 @@ private Session currentSession() {
...
+  public ArrayRunSample get(ArrayRun run, Array array, String position, Sample sample) throws IOException {
     ArrayRunSampleId id = new ArrayRunSampleId();
     id.setArrayRun(run);
+    id.setArray(array);
     id.setPosition(position);
+    id.setSample(sample);
     ArrayRunSample result = currentSession().get(ArrayRunSample.class, id);
     if (result == null) {
-      result = new ArrayRunSample(run, position);
+      result = new ArrayRunSample(run, array, position, sample);
```
djcooke: I had commented on this previously, but it's kind of disconnected now, so leaving a new comment... Please make sure this combination is valid and return null if it's not.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -97,54 +111,64 @@ public void save(List<ArrayRunSample> arrayRunSamples) throws IOException {
...
+    validateChange(arrayRunSample);
+    Sample expectedSample = arrayRunSample.getArrayRun().getArray().getSample(arrayRunSample.getPosition());
+    ArrayRunSample managed = arrayRunSampleDao.get(arrayRunSample.getArrayRun(), arrayRunSample.getArrayRun().getArray(),
+        arrayRunSample.getPosition(), expectedSample);
+    User user = authorizationManager.getCurrentUser();
+    updateQcDetails(arrayRunSample, managed, ArrayRunSample::getQcStatus, ArrayRunSample::getQcUser,
+        ArrayRunSample::setQcUser, authorizationManager, ArrayRunSample::getQcDate, ArrayRunSample::setQcDate);
+
+    List<ValidationError> errors = new ArrayList<>();
+    validateQcUser(arrayRunSample.getQcStatus(), arrayRunSample.getQcUser(), errors);
```
djcooke: Why is this outside of `validateChange`?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -0,0 +1,174 @@
...
+  private ArrayRunService arrayRunService;
+
+  @Autowired
+  private RunItemQcStatusService runItemQcStatusService;
+
+  @Autowired
+  private AuthorizationManager authorizationManager;
+
+  @Override
+  public ArrayRunSample get(ArrayRun run, String position) throws IOException {
```
djcooke: Parameters should include array and sample

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunService.java
```diff
@@ -237,6 +252,20 @@ private void applyChanges(ArrayRun from, ArrayRun to) {
...
+    }
+    for (Map.Entry<String, Sample> entry : run.getArray().getSamples().entrySet()) {
+      ArrayRunSample sample = arrayRunSampleService.get(run, entry.getKey());
+      if (sample == null) {
+        sample = new ArrayRunSample(run, run.getArray(), entry.getKey(), entry.getValue());
+      }
+      sample.setLastModifier(authorizationManager.getCurrentUser());
+      arrayRunSampleService.save(sample);
+    }
+  }
```
djcooke: I think this isn't necessary - the ArrayRunSample database record doesn't need to exist unless its QC status is set. If this can be removed, you probably don't need the flushes in `HibernateArrayRunDao` either.

The `lastModifier` field is also not useful - it'll be the same as the `ArrayRun` so if you need it, you can just take it from there. You might need it for changelogs... The QC status changes should be logged.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
riyapatel36: It depends on `updateQcDetails` populating `qcUser` first, but I can restructure the flow.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
```diff
@@ -32,6 +32,21 @@ private Session currentSession() {
 
   @Override
   public ArrayRunSample get(ArrayRun run, Array array, String position, Sample sample) throws IOException {
+    if (run == null || array == null || sample == null || position == null) {
+      return null;
+    }
+    if (run.getArray() == null || run.getArray().getId() != array.getId()) {
```
djcooke: Pull the run from the database to use for these checks

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -112,22 +107,36 @@ public void save(List<ArrayRunSample> arrayRunSamples) throws IOException {
...
+  }
 
-    List<ValidationError> errors = new ArrayList<>();
-    validateQcUser(arrayRunSample.getQcStatus(), arrayRunSample.getQcUser(), errors);
-    if (!errors.isEmpty()) {
-      throw new ValidationException(errors);
+  private ArrayRunSample getManagedRecord(ArrayRunSample arrayRunSample) throws IOException {
+    ArrayRun run = arrayRunSample.getArrayRun();
+    if (run == null || run.getArray() == null) {
+      return new ArrayRunSample();
```
djcooke: If the run or array are null, that means the submitted ArrayRunSample is invalid. Throw a ValidationException rather than returning a new ArrayRunSample that is also invalid. Same for the other conditions checked below.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -158,17 +167,52 @@ private void validateChange(ArrayRunSample arrayRunSample) {
...
+    boolean noteChanged = !java.util.Objects.equals(oldQcNote, newQcNote);
+    if (!statusChanged && !noteChanged) {
+      return;
+    }
+
+    String sampleLabel = arrayRunSample.getSample() == null ? arrayRunSample.getPosition()
+        : arrayRunSample.getSample().getName();
+    String summary;
+    if (statusChanged && noteChanged) {
+      summary = String.format("QC updated for sample %s at %s: status %s -> %s; note %s -> %s", sampleLabel,
```
djcooke: QC user and date should also be logged. Better to do triggers via database trigger when possible. See `05_trigger_runpartition.sql` for how it's done with run-libraries.

djcooke (commented):
`listByRunId` in the service is still checking whether the ArrayRunSamples exist, which should be handled in the DAO.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
```diff
@@ -64,23 +63,17 @@ public List<ArrayRunSample> listByRunId(long arrayRunId) throws IOException {
...
-        .collect(Collectors.toMap(ArrayRunSample::getPosition, sample -> sample));
-
     List<ArrayRunSample> results = new ArrayList<>(samples.size());
     for (Map.Entry<String, Sample> entry : samples.entrySet()) {
-      ArrayRunSample item = byPosition.get(entry.getKey());
-      if (item == null) {
-        item = arrayRunSampleDao.get(run, array, entry.getKey(), entry.getValue());
+      ArrayRunSample item = arrayRunSampleDao.get(run, array, entry.getKey(), entry.getValue());
+      if (item != null) {
+        results.add(item);
```
djcooke: Use `ArrayRunSampleDao#listByRunId` to minimize the number of queries instead of fetching each one individually. `ArrayRunSampleDao#listByRunId` should construct any missing entries, just like `get`.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
```diff
@@ -48,13 +53,14 @@ public ArrayRunSample get(ArrayRun run, Array array, String position, Sample sam
     }
```
djcooke: You're validating that the parameters that were given actually match what's in the database, so don't trust the parameters. Use `persistedRun.getArray()` instead of `array`

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
```diff
@@ -48,13 +53,14 @@ public ArrayRunSample get(ArrayRun run, Array array, String position, Sample sam
...
-    id.setArrayRun(run);
+    id.setArrayRun(persistedRun);
     id.setArray(array);
     id.setPosition(position);
     id.setSample(sample);
+
     ArrayRunSample result = currentSession().get(ArrayRunSample.class, id);
     if (result == null) {
-      result = new ArrayRunSample(run, array, position, sample);
+      result = new ArrayRunSample(persistedRun, array, position, sample);
```
djcooke: Use all the persisted objects to construct `id` and `result`

sqlstore/src/main/resources/db/migration_afterMigrate/06_trigger_arrayrunsample.sql
```diff
@@ -0,0 +1,75 @@
...
+      COALESCE(CONCAT_WS(',',
+        makeChangeColumn('sample statusId', NULL, NEW.statusId),
+        makeChangeColumn('sample qcNote', NULL, NEW.qcNote),
+        makeChangeColumn('sample qcUser', NULL, NEW.qcUser),
+        makeChangeColumn('sample qcDate', NULL, NEW.qcDate)
+      ), ''),
+      COALESCE(NEW.qcUser, (SELECT lastModifier FROM ArrayRun WHERE arrayRunId = NEW.arrayRunId)),
+      CONCAT(
+        COALESCE((SELECT name FROM Sample WHERE sampleId = NEW.sampleId), NEW.position),
+        ' at ', NEW.position, ' ',
```
djcooke: Use the sample alias instead of name and don't coalesce. The sample alias will never be null. Same below.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
djcooke: You're logging changes via triggers now, so this is no longer needed.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayRunSampleService.java
riyapatel36: `listByRunId()` should load existing rows and also create any missing `ArrayRunSample`, so the service only makes that one DAO call?

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
```diff
@@ -40,38 +39,75 @@ public ArrayRunSample get(ArrayRun run, Array array, String position, Sample sam
...
+      return new ArrayList<>();
+    }
+
     QueryBuilder<ArrayRunSample, ArrayRunSample> builder =
         new QueryBuilder<>(currentSession(), ArrayRunSample.class, ArrayRunSample.class);
-    Join<ArrayRunSample, ArrayRun> runJoin = builder.getJoin(builder.getRoot(), ArrayRunSample_.arrayRun);
-    builder.addPredicate(builder.getCriteriaBuilder().equal(runJoin.get(ArrayRun_.id), arrayRunId));
-    return builder.getResultList();
+    builder.addPredicate(builder.getCriteriaBuilder().equal(
+        builder.getRoot().get("arrayRun").get("id"), arrayRunId));
```
djcooke: Use metamodel (`ArrayRun_`) properties rather than strings

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
```diff
@@ -40,38 +39,75 @@ public ArrayRunSample get(ArrayRun run, Array array, String position, Sample sam
...
+      Sample expectedSample = entry.getValue();
+
+      ArrayRunSample existingItem = existing.stream()
+          .filter(item -> item.getArrayRun() != null && item.getArrayRun().getId() == persistedRun.getId())
+          .filter(item -> item.getArray() != null && item.getArray().getId() == persistedArray.getId())
+          .filter(item -> item.getSample() != null && item.getSample().getId() == expectedSample.getId())
+          .filter(item -> position.equals(item.getPosition()))
+          .findFirst()
+          .orElse(null);
+
```
djcooke, line 99: You pulled this all from the database, so you can just filter by position and trust that the run, array, and sample will match.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultArrayService.java
```diff
@@ -228,6 +235,19 @@ private void recordRemoval(Array array, String sampleName, String position) thro
...
   }
 
+  private void cleanupArrayRunSamples(Array array) throws IOException {
+    for (ArrayRun run : arrayRunService.listByArrayId(array.getId())) {
+      for (ArrayRunSample arrayRunSample : arrayRunSampleService.listByRunId(run.getId())) {
+        Sample expectedSample = array.getSamples().get(arrayRunSample.getPosition());
+        if (expectedSample == null
+            || arrayRunSample.getSample() == null
+            || arrayRunSample.getSample().getId() != expectedSample.getId()) {
+          arrayRunSampleService.delete(arrayRunSample);
```
djcooke: Use `deleteByRunId`

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultSampleService.java
```diff
@@ -1284,6 +1290,15 @@ public ValidationResult validateDeletion(Sample object) throws IOException {
...
     }
+    final int arrays = arrayService.listBySampleId(object.getId()).size();
+    if (arrays > 0) {
+      result.addError(ValidationError.forDeletionUsage(object, arrays, Pluralizer.arrays(arrays)));
+    }
+    final int arrayRuns = arrayRunService.listBySampleId(object.getId()).size();
+    if (arrayRuns > 0) {
+      result.addError(
+          ValidationError.forDeletionUsage(object, arrayRuns, arrayRuns == 1 ? "array run" : "array runs"));
+    }
```
djcooke: Don't really need to check array runs since there would already be an error because of the array

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRunRestController.java
```diff
@@ -152,4 +215,39 @@ public void bulkDelete(@RequestBody(required = true) List<Long> ids) throws IOEx
     RestUtils.bulkDelete("Array Run", ids, arrayRunService);
   }
 
+  private static boolean containsIgnoreCase(String value, String search) {
+    return value != null && value.toLowerCase(Locale.ROOT).contains(search);
+  }
+
+  private static int parseIntOrDefault(String value, int defaultValue) {
```
djcooke: This is used for "iDisplayStart", "iDisplayLength", and "sEcho" parameters. Are there cases where they should be missing or invalid? Invalid should likely be treated as invalid (return 400 Bad Request), and defaults should only be used if there are cases where the parameters should be missing.

miso-web/src/main/webapp/scripts/list_arrayrunsample.js
```diff
@@ -0,0 +1,104 @@
...
+  },
+
+  onFirstLoad: function (data, config) {
+    $("#listingSamplesTable_wrapper").css("padding-bottom", "");
+    $("#listingSamplesTable_wrapper .datatable-scroll").css({
+      "margin-bottom": "",
+      "min-height": data && data.length ? "180px" : "",
+      "overflow-x": "auto",
+      "overflow-y": "visible",
+    });
```
djcooke: Why does this table need different styling than other tables in MISO?

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
```diff
@@ -0,0 +1,136 @@
...
+        .setParameter("runId", arrayRunSample.getArrayRun().getId())
+        .setParameter("arrayId", arrayRunSample.getArray().getId())
+        .setParameter("position", arrayRunSample.getPosition())
+        .setParameter("sampleId", arrayRunSample.getSample().getId())
+        .executeUpdate();
+  }
+
+  @Override
+  public void deleteByRunId(long arrayRunId) throws IOException {
+    currentSession().createMutationQuery("delete from ArrayRunSample where arrayRun.id = :runId")
```
djcooke: Avoid query strings like this. Can either adapt `QueryBuilder` to work for deletes too, or just retrieve the records to be deleted and call `currentSession().remove` with them. The second option is easier and performance will likely never be a problem.

sqlstore/src/main/resources/db/migration/V8010__arrayrun_sample.sql
```diff
@@ -0,0 +1,22 @@
+ALTER TABLE ArrayRun
+  ADD CONSTRAINT uk_arrayRun_id_array UNIQUE (arrayRunId, arrayId);
```
djcooke: I don't think we want to prevent using the same array on multiple array runs.

sqlstore/src/main/resources/db/migration/V8010__arrayrun_sample.sql
```diff
@@ -0,0 +1,22 @@
...
+
+ALTER TABLE ArrayPosition
+  ADD CONSTRAINT uk_arrayPosition_array_position_sample UNIQUE (arrayId, position, sampleId);
```
djcooke: You can't have multiple samples in the same position, so the unique constraint should be on array+position only

sqlstore/src/main/resources/db/migration_afterMigrate/05_trigger_arrayrunsample.sql
djcooke, line 1: The filename should start with `05_` like the other trigger scripts. The order of the triggers doesn't matter, but we want `06_reenable_warnings.sql` to run after all of them.

djcooke (commented):
There is an "unexpected error" if you try to remove a sample from the array after it has been QCed on an array run.

miso-web/src/main/webapp/scripts/list_arrayrunsample.js
riyapatel36: This table is different because the QC status column shows a hover tooltip, and that tooltip was getting clipped by the table container. The extra styling was only to give that tooltip room to display properly. I can try to switch it to match the run-libraries implementation so it stays consistent with the rest of MISO.

djcooke: That happens with other tables too and they seem to behave the same without the extra styles

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRunRestController.java
```diff
@@ -219,14 +220,14 @@ private static boolean containsIgnoreCase(String value, String search) {
...
+  private static int parseRequiredInt(String name, String value) {
     if (value == null) {
-      return defaultValue;
+      throw new IllegalArgumentException("Missing parameter: " + name);
     }
     try {
       return Integer.parseInt(value);
     } catch (NumberFormatException e) {
-      return defaultValue;
+      throw new IllegalArgumentException("Invalid parameter: " + name);
```
djcooke: Throw RestException with Bad Request status since this is a client error. IllegalArgument would result in a 500 Server Error response

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
```diff
@@ -118,19 +119,41 @@ public void save(ArrayRunSample arrayRunSample) throws IOException {
...
-        .executeUpdate();
+    ArrayRunSample managed = currentSession().get(ArrayRunSample.class, new ArrayRunSampleId(
+        arrayRunSample.getArrayRun(),
+        arrayRunSample.getArray(),
+        arrayRunSample.getPosition(),
+        arrayRunSample.getSample()));
+
+    if (managed != null) {
+      currentSession().remove(managed);
+      currentSession().flush();
```
djcooke, line 126: Is the flush necessary? If so, please add a comment explaining why. We don't flush after deleting `RunPartitionAliquot`s though, so I'm not sure why we'd need to here. Same in `deleteByRunId` below.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
```diff
@@ -118,19 +119,41 @@ public void save(ArrayRunSample arrayRunSample) throws IOException {
...
-    currentSession().createMutationQuery("delete from ArrayRunSample where arrayRun.id = :runId")
-        .setParameter("runId", arrayRunId)
-        .executeUpdate();
+    for (ArrayRunSample item : getStoredByRunId(arrayRunId)) {
+      currentSession().remove(item);
+    }
+    currentSession().flush();
+  }
+
+  private List<ArrayRunSample> getStoredByRunId(long arrayRunId) {
```
djcooke, line 139: You can use this in `listByRunId` too since the query is currently duplicated there.

sqlstore/src/main/resources/db/migration/V8010__arrayrun_sample.sql
djcooke: I'm not sure what I was thinking since that's not what this was doing... 😅 The constraint isn't necessary anyway though. `arrayRunId` is the primary key, so it already has to be unique.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
```diff
@@ -127,6 +122,7 @@ public void delete(ArrayRunSample arrayRunSample) throws IOException {
 
     if (managed != null) {
       currentSession().remove(managed);
+      // Flush so this delete is written before the same transaction updates the parent row.
```
djcooke, line 125: This is just a delete method, so there's not necessarily anything else happening in the same transaction, and it's not clear what a "parent row" is.

djcooke (commented):
Functionality looks good. Please add tests for `ArrayRunRestController` and `HibernateArrayRunSampleDao`, and documentation in the Array Runs section of the user manual.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateArrayRunSampleDao.java
riyapatel36: I meant this method is used in places where we may update ArrayRun or ArrayPosition in the same transaction, and that’s where we were seeing the FK error. I’ll update the comment to explain that more clearly.

miso-web/src/spring-test/java/uk/ac/bbsrc/tgac/miso/webapp/springtest/ArrayRunRestControllerST.java
```diff
@@ -117,4 +122,116 @@ public void testDeleteFail() throws Exception {
...
+    ArrayRunSampleId id1 = new ArrayRunSampleId(run, array, "R01C01", sample1);
+    assertNotNull(currentSession().get(ArrayRunSample.class, id1));
+
+    ArrayRun updatedArr = baseTestUpdate(CONTROLLER_BASE, arr, 2, controllerClass);
+
+    assertNull(updatedArr.getArray());
+    currentSession().clear();
+    assertNull(currentSession().get(ArrayRunSample.class, id1));
+
+    getMockMvc().perform(get(CONTROLLER_BASE + "/2/samples")
```
djcooke: Don't use a different endpoint to verify results. Pull the data from the database instead.

djcooke (approved):
Approved after addressing my other comment

callunity (approved):

PR #3218, opened 2026-03-20
Gp 5294 multi select position array

riyapatel36:
Jira ticket: https://jira.oicr.on.ca/browse/GP-5294

- [x] Includes a change file

### DIFF
--- a/changes/add_multi_position_array.md
+++ b/changes/add_multi_position_array.md
@@ -1,0 +1,1 @@
+Multiple positions can be selected and added samples in array.
--- a/changes/add_multi_position_array.md
+++ b/changes/add_multi_position_array.md
@@ -1,1 +1,1 @@
-Multiple positions can be selected and added samples in array.
+Multiple positions can be selected at once for adding/removing samples from the array on the Edit Array page.
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRestController.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRestController.java
@@ -97,12 +97,8 @@ public class ArrayRestController extends AbstractRestController {
   @DeleteMapping(value = "/{arrayId}/positions/{position}")
   public @ResponseBody ArrayDto removeSample(@PathVariable(name = "arrayId", required = true) long arrayId,
       @PathVariable(name = "position", required = true) String position) throws IOException {
-    Array array = arrayService.get(arrayId);
-    if (array == null) {
-      throw new RestException(ERROR_NOTFOUND, Status.NOT_FOUND);
-    } else if (!array.isPositionValid(position)) {
-      throw new RestException("Invalid array position", Status.BAD_REQUEST);
-    }
+    Array array = getArrayOrThrow(arrayId);
+    validatePositions(array, Collections.singletonList(position));
     if (array.getSample(position) == null) {
       // already empty - do nothing
       return Dtos.asDto(array);
--- a/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRestController.java
+++ b/miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRestController.java
@@ -2,7 +2,6 @@ package uk.ac.bbsrc.tgac.miso.webapp.controller.rest;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -92,23 +91,6 @@ public class ArrayRestController extends AbstractRestController {
       array.setSamples(existing.getSamples());
       return array;
     }), arrayService, Dtos::asDto);
-  }
-
-  @DeleteMapping(value = "/{arrayId}/positions/{position}")
-  public @ResponseBody ArrayDto removeSample(@PathVariable(name = "arrayId", required = true) long arrayId,
-      @PathVariable(name = "position", required = true) String position) throws IOException {
-    Array array = getArrayOrThrow(arrayId);
-    validatePositions(array, Collections.singletonList(position), Status.NOT_FOUND);
-    if (array.getSample(position) == null) {
-      // already empty - do nothing
-      return Dtos.asDto(array);
-    }
-
-    array.setSample(position, null);
-    arrayService.update(array);
-
-    Array saved = arrayService.get(arrayId);
-    return Dtos.asDto(saved);
   }
 
   @PutMapping(value = "/{arrayId}/positions/{position}")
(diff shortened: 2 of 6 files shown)


miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRestController.java
```diff
@@ -163,4 +221,24 @@ public void bulkDelete(@RequestBody(required = true) List<Long> ids) throws IOEx
     RestUtils.bulkDelete("Array", ids, arrayService);
   }
 
+  private Array getArrayOrThrow(long arrayId) throws IOException {
```
callunity, line 202: Nice utility method! Would it make sense to refactor `removeSample` to also use `getArrayOrThrow` and `validatePositions` ?

riyapatel36: Agree Heather, I will update that. Thanks

callunity (approved):

miso-web/src/main/webapp/WEB-INF/pages/editArray.jsp
```diff
@@ -48,44 +48,71 @@
   <br/>
   <h1>Samples</h1>
   <div id="arraySamplesSection">
+    <div class="note">
+      Hold down Ctrl (Windows, Linux) or Command (Mac) to select multiple positions.
+      Click a column header to select the entire column.
```
djcooke: Row headers too. The only array model we have defined only has 1 column, so this doesn't seem useful, but arrays may have multiple columns.

changes/add_multi_position_array.md
```diff
@@ -0,0 +1 @@
+Multiple positions can be selected and added samples in array.
```
djcooke:
```suggestion
Multiple positions can be selected at once for adding/removing samples from the array on the Edit Array page.
```

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRestController.java
```diff
@@ -163,4 +217,24 @@ public void bulkDelete(@RequestBody(required = true) List<Long> ids) throws IOEx
...
+
+  private void validatePositions(Array array, List<String> positions) {
+    if (positions == null || positions.isEmpty()) {
+      throw new RestException("No array positions selected", Status.BAD_REQUEST);
+    }
+    List<String> invalidPositions = positions.stream()
+        .filter(position -> !array.isPositionValid(position))
+        .collect(Collectors.toList());
+    if (!invalidPositions.isEmpty()) {
+      throw new RestException("Invalid array position", Status.BAD_REQUEST);
```
djcooke: Status should be "not found" when the position is part of the URL; otherwise, "bad request" is correct

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRestController.java
```diff
@@ -116,22 +115,77 @@ public DataTablesResponseDto<ArrayDto> dataTable(HttpServletRequest request) thr
...
-    } else if (!array.isPositionValid(position)) {
-      throw new RestException("Invalid array position", Status.BAD_REQUEST);
-    }
+    return addSample(arrayId, sampleId, Collections.singletonList(position));
+  }
+
+  @PutMapping(value = "/{arrayId}/bulk-update")
+  public @ResponseBody ArrayDto addSample(@PathVariable(name = "arrayId", required = true) long arrayId,
+      @RequestParam(name = "sampleId", required = true) long sampleId,
+      @RequestBody(required = true) List<String> positions) throws IOException {
```
djcooke: I think the previously existing endpoint was better. There's no need to set a single sample to multiple positions as this allows. It is only being used to set a single sample to a single position, which the previous endpoint covered.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRestController.java
```diff
@@ -116,22 +115,77 @@ public DataTablesResponseDto<ArrayDto> dataTable(HttpServletRequest request) thr
...
+    public String getSearchString() {
+      return searchString;
+    }
+
+    public void setSearchString(String searchString) {
+      this.searchString = searchString;
+    }
+  }
+
+  @PostMapping(value = "/{arrayId}/bulk-update-search")
```
djcooke: `PUT` to `/{arrayId}/positions` would be a better REST URL

miso-web/src/main/webapp/scripts/array_ajax.js
```diff
@@ -109,22 +97,34 @@
...
-      url: Urls.rest.arrays.position(arrayJson.id, selectedPosition),
-      type: "DELETE",
-      dataType: "json",
-    })
-      .done(function (data) {
-        clearSampleSearchResults();
-        SampleArray.setArrayJson(data);
+    var removeNext = function (index) {
+      $.ajax({
+        url: Urls.rest.arrays.position(arrayJson.id, selectedPositions[index]),
```
djcooke: Add an endpoint to bulk remove (or use the bulk update endpoint with no search strings since that already works for removing samples) rather than making multiple requests in serial

miso-web/src/main/webapp/WEB-INF/pages/editArray.jsp
riyapatel36: I might have to make changes in code as well, This current version only support column headers I believe, I will update for multiple column support. Thanks Dillan

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/rest/ArrayRestController.java
```diff
@@ -225,15 +225,15 @@ private Array getArrayOrThrow(long arrayId) throws IOException {
     return array;
   }
 
-  private void validatePositions(Array array, List<String> positions) {
+  private void validatePositions(Array array, List<String> positions, Status invalidStatus) {
```
djcooke: This is only used in one place and it's for a single position, so the list is unnecessary.

`bulkUpdatePositions` does similar validation for multiple positions, but I think the way it's handled there is better and more specific to that endpoint.

miso-web/src/main/webapp/scripts/array_ajax.js
```diff
@@ -103,28 +103,46 @@
     }
     showSamplesLoading(true);
 
-    var removeNext = function (index) {
+    if (selectedPositions.length === 1) {
```
djcooke: You could simplify this by using the bulk logic even if there's only one position selected. Can also remove the single delete endpoint if it's not needed.

djcooke (approved):

PR #3219, opened 2026-03-20
[GLT-4769] add library aliquot QCs to Pinery

djcooke:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4769

- [x] Includes a change file

### DIFF
--- a/changes/change_pinery_libraryAliquotQc.md
+++ b/changes/change_pinery_libraryAliquotQc.md
@@ -1,0 +1,1 @@
+Library aliquots in Pinery will now include as sample attributes the same QC types that are included from samples or libraries
--- a/pinery-miso/src/main/resources/ca/on/oicr/pinery/lims/miso/queryAllSampleQcs.sql
+++ b/pinery-miso/src/main/resources/ca/on/oicr/pinery/lims/miso/queryAllSampleQcs.sql
@@ -26,5 +26,20 @@ WHERE qc.qcId IN (
   GROUP BY qc.library_libraryId, qc.type
 )
 
+UNION ALL
+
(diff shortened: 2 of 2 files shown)


changes/change_pinery_libraryAliquotQc.md
```diff
@@ -0,0 +1 @@
+Library aliquots in Pinery will now include as sample attributes the same QC types that are included from samples or libraries
```
callunity, line 1: I'm not sure I understand this fully: "the same QC types that are included from samples or libraries". It looks to me like it's the actual QC that's being selected in the SQL?
Is it accurate to say "Library aliquots in Pinery now include QCs in their sample attributes, similar to the QCs that are already present on samples or libraries" ? or "Library aliquots in Pinery now include QCs which have the same QC types as are already present on samples or libraries"?

djcooke: The SQL selects all QCs, and then specific ones are included as sample attributes based on the QC type (See [QcConverter](https://github.com/miso-lims/miso-lims/blob/develop/pinery-miso/src/main/java/ca/on/oicr/pinery/lims/miso/converters/QcConverter.java)). Realistically, this is going to mean little to anyone since the QCs that are included are specific to our configuration and our purposes. I don't want to give the impression that "QCs are available in Pinery" as something other than sample attributes, or that all QCs are included.

callunity: Great, thanks for clarifying.

callunity (approved):

riyapatel36 (approved):

PR #3223, opened 2026-04-06
#3221: add configuration to hide bulk table columns

### LINKED ISSUE #3221: Add configurable visibility for bulk table columns
Several bulk table columns (used across library receipt, sample receipt, and other bulk pages) are not needed by all MISO instances.
  Currently there is no way to hide them without modifying source code.

  The existing pattern for library-specific columns (
`miso.display.library.bulk.libraryalias, miso.display.library.bulk.description,
  miso.display.library.bulk.volume`) works well, but there are no equivalent options for shared columns like Discarded, QC Status,
  Requisition, etc.

  Proposal

  Add `miso.display.bulk.*` configuration properties that allow hiding the following columns across all bulk pages:

```
  - miso.display.bulk.receiptConfirmed
  - miso.display.bulk.receiptQcPassed
  - miso.display.bulk.receiptQcNote
  - miso.display.bulk.requisitionAlias
  - miso.display.bulk.requisition
  - miso.display.bulk.assay
  - miso.display.bulk.matrixBarcode
  - miso.display.bulk.discarded
  - miso.display.bulk.detailedQcStatus
  - miso.display.bulk.qcNote
```
  All default to true (visible), so this is a no-op for existing installations. Instances that don't use these fields can set them to false
   in miso.properties.

  Implementation approach

  - Config properties via `@Value` in `ConstantsController`, exposed through the existing `Constants` JS object
  - include checks added to column definitions in `bulk.js`
  - `updateField` made tolerant of missing columns (skips silently) to handle cross-column `onChange` dependencies when a target column is
  hidden
  - When `detailedQcStatus` is hidden, defaults to "Ready" instead of requiring user selection

mohammedkhalfan:
Closes https://github.com/miso-lims/miso-lims/issues/3221                                                                                
                                              
  ## Summary

  Adds `miso.properties` flags to control visibility of column groups on bulk edit/create pages:                                           
  - `miso.display.bulk.receiptQc` — Receipt Confirmed, Receipt QC Passed, Receipt QC Note
  - `miso.display.bulk.requisition` — Requisition Alias, Requisition, Assay                                                                
  - `miso.display.bulk.qcStatus` — QC Status, QC Note                                                                                      
  - `miso.display.bulk.matrixBarcode` — Matrix Barcode
  - `miso.display.bulk.discarded` — Discarded                                                                                              
  - `miso.display.bulk.defaultDetailedQcStatus` — configurable default QC status when qcStatus is hidden
                                                                                                                                           
  All flags default to `true` (visible), so existing installations are unaffected.
                                                                                                                                           
  ## Technical notes                          

  All `getColumnIndex` call sites made null-tolerant so that cross-column `onChange` handlers (e.g., Date of Receipt → Receipt Confirmed)  
  gracefully skip hidden columns instead of throwing errors.

### DIFF
--- a/changes/add_bulk_column_visibility_config.md
+++ b/changes/add_bulk_column_visibility_config.md
@@ -1,0 +1,9 @@
+Configuration options to hide columns on bulk edit/create pages. The following `miso.properties`
+settings can be set to `false` to hide their respective column groups (all default to `true`):
+* `miso.display.bulk.receiptQc` - Receipt Confirmed, Receipt QC Passed, Receipt QC Note
+* `miso.display.bulk.requisition` - Requisition Alias, Requisition, Assay
+* `miso.display.bulk.qcStatus` - QC Status, QC Note
+* `miso.display.bulk.matrixBarcode` - Matrix Barcode
+* `miso.display.bulk.discarded` - Discarded
+* `miso.display.bulk.defaultDetailedQcStatus` - when `qcStatus` is hidden, sets the default QC
+  status value (should match a configured QC status description)
--- a/miso-web/src/main/resources/miso.properties
+++ b/miso-web/src/main/resources/miso.properties
@@ -90,6 +90,14 @@ miso.display.library.bulk.libraryalias:true
 miso.display.library.bulk.description:false
 miso.display.library.bulk.volume:false
 
+## bulk table column visibility settings (all default to true/visible)
+# miso.display.bulk.receiptQc:true
+# miso.display.bulk.requisition:true
+# miso.display.bulk.qcStatus:true
+# miso.display.bulk.matrixBarcode:true
+# miso.display.bulk.discarded:true
+# miso.display.bulk.defaultDetailedQcStatus:
+
 ## Strict Index Conflict rules for adding Library Aliquots to Pools.
 ## When enabled, Library Aliquots with duplicate or near-duplicate indices cannot be added to Pools.
 miso.pools.strictIndexChecking:false
--- a/miso-web/src/main/resources/miso.properties
+++ b/miso-web/src/main/resources/miso.properties
@@ -96,7 +96,7 @@ miso.display.library.bulk.volume:false
 # miso.display.bulk.qcStatus:true
 # miso.display.bulk.matrixBarcode:true
 # miso.display.bulk.discarded:true
-# miso.display.bulk.defaultDetailedQcStatus:
+# miso.defaults.bulk.detailedQcStatus:
 
 ## Strict Index Conflict rules for adding Library Aliquots to Pools.
 ## When enabled, Library Aliquots with duplicate or near-duplicate indices cannot be added to Pools.
--- a/miso-web/src/main/webapp/scripts/bulk.js
+++ b/miso-web/src/main/webapp/scripts/bulk.js
@@ -258,6 +258,26 @@ BulkUtils = (function ($) {
       });
     },
 
+    applyDefaultDetailedQcStatus: function (data) {
+      if (Constants.showQcStatus !== false || !Constants.defaultDetailedQcStatus) {
+        return;
+      }
+      var status = Constants.detailedQcStatuses.find(function (s) {
+        return s.description === Constants.defaultDetailedQcStatus;
+      });
+      if (!status) {
+        return;
+      }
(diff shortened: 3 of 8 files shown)


changes/add_bulk_column_visibility_config.md
```diff
@@ -0,0 +1,9 @@
+Configuration options to hide columns on bulk edit/create pages. The following `miso.properties`
+settings can be set to `false` to hide their respective column groups (all default to `true`):
+* `miso.display.bulk.receiptQc` - Receipt Confirmed, Receipt QC Passed, Receipt QC Note
+* `miso.display.bulk.requisition` - Requisition Alias, Requisition, Assay
+* `miso.display.bulk.qcStatus` - QC Status, QC Note
+* `miso.display.bulk.matrixBarcode` - Matrix Barcode
+* `miso.display.bulk.discarded` - Discarded
+* `miso.display.bulk.defaultDetailedQcStatus` - when `qcStatus` is hidden, sets the default QC
+  status value (should match a configured QC status description)
```
djcooke, line 9: `miso.defaults.bulk.detailedQcStatus` would be more consistent with other properties, and this could work whether or not the column is hidden

miso-web/src/main/webapp/scripts/bulk.js
```diff
@@ -794,6 +799,7 @@ BulkUtils = (function ($) {
               ).description;
             },
             required: true,
+            initial: Constants.showQcStatus === false ? Constants.defaultDetailedQcStatus : undefined,
```
djcooke: As previously suggested, this could be useful even with the column visible.

If the column is hidden, this won't do what you want though, as the `initial` value won't be applied if the column isn't included. I don't think we want to change that, so the best way to handle it is probably in the `confirmSave` function of each relevant bulk target (sample, library, library aliquot).

miso-web/src/main/resources/miso.properties
```diff
@@ -90,6 +90,14 @@ miso.display.library.bulk.libraryalias:true
 miso.display.library.bulk.description:false
 miso.display.library.bulk.volume:false
 
+## bulk table column visibility settings (all default to true/visible)
+# miso.display.bulk.receiptQc:true
+# miso.display.bulk.requisition:true
+# miso.display.bulk.qcStatus:true
+# miso.display.bulk.matrixBarcode:true
+# miso.display.bulk.discarded:true
+# miso.display.bulk.defaultDetailedQcStatus:
```
djcooke: Please also document these in `docs/admin/site-configuration.md`

mohammedkhalfan:
Thanks for the detailed feedback! I'll work on these changes:
                                                                                                                            
  1. Rename to miso.defaults.bulk.detailedQcStatus for consistency
  2. Move the default application to confirmSave so it works whether the column is visible or hidden                                       
  3. Add documentation in docs/admin/site-configuration.md                                                                            
  
Appreciate the guidance. I'm still getting familiar with the codebase internals, so it's very helpful. Will update the PR once the changes are ready.

djcooke:
> 2. Move the default application to confirmSave so it works whether the column is visible or hidden

I think it needs to be in both places to work for both situations. If the column is shown, the default value should be selected automatically when the table is loaded, but the user may change it, including a change to "Not Ready" (null). `initial` can handle all of that. `confirmSave` should only set the value if the column is hidden.

miso-web/src/main/webapp/scripts/bulk.js
```diff
@@ -258,6 +258,26 @@ BulkUtils = (function ($) {
       });
     },
 
+    applyDefaultDetailedQcStatus: function (data) {
+      if (Constants.showQcStatus !== false || !Constants.defaultDetailedQcStatus) {
+        return;
+      }
```
djcooke: I think this should also be skipped when editing. Otherwise, someone could manually change the QC status to "Not Ready" (null) on the single Edit Sample page, and that would be undone if the sample was later edited in bulk, and with no indication of that happening.

The config object for each of these pages includes a "pageMode" property that will be "create"/"propagate"/"edit", so you can use that to check whether the page is an edit page. The `config` object is passed into `confirmSave`.

djcooke (approved):

riyapatel36 (approved):

PR #3234, opened 2026-05-06
GLT-4786-Contacts-to-requisition

riyapatel36:
Jira ticket : https://jira.oicr.on.ca/browse/GLT-4786

- [x] Includes a change file

### DIFF
--- a/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRequisitionService.java
+++ b/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRequisitionService.java
@@ -26,13 +26,17 @@ import uk.ac.bbsrc.tgac.miso.core.data.Library;
 import uk.ac.bbsrc.tgac.miso.core.data.Requisitionable;
 import uk.ac.bbsrc.tgac.miso.core.data.Sample;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.Assay;
+import uk.ac.bbsrc.tgac.miso.core.data.impl.Contact;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.Requisition;
+import uk.ac.bbsrc.tgac.miso.core.data.impl.RequisitionContact;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.RequisitionPause;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.RequisitionSupplementalLibrary;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.RequisitionSupplementalSample;
 import uk.ac.bbsrc.tgac.miso.core.security.AuthorizationManager;
 import uk.ac.bbsrc.tgac.miso.core.service.AssayService;
 import uk.ac.bbsrc.tgac.miso.core.service.ChangeLogService;
+import uk.ac.bbsrc.tgac.miso.core.service.ContactRoleService;
+import uk.ac.bbsrc.tgac.miso.core.service.ContactService;
 import uk.ac.bbsrc.tgac.miso.core.service.LibraryService;
 import uk.ac.bbsrc.tgac.miso.core.service.RequisitionService;
 import uk.ac.bbsrc.tgac.miso.core.service.SampleService;
@@ -69,6 +73,10 @@ public class DefaultRequisitionService extends AbstractSaveService<Requisition> 
   private TransactionTemplate transactionTemplate;
   @Autowired
   private HibernateUtilDao hibernateUtilDao;
+  @Autowired
+  private ContactService contactService;
+  @Autowired
+  private ContactRoleService contactRoleService;
 
   @Override
   public DeletionStore getDeletionStore() {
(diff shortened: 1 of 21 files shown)


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRequisitionService.java
```diff
@@ -132,6 +169,39 @@ protected void collectValidationErrors(Requisition object, Requisition beforeCha
       }
     }
     validatePauses(object.getPauses(), errors);
+    validateContacts(object.getContacts(), errors);
+  }
+
+  private void validateContacts(List<RequisitionContact> contacts, List<ValidationError> errors) {
```
avarsava, line 175: Could this nested loop be avoided by using a Set of RequisitionContacts instead?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRequisitionService.java
```diff
@@ -170,6 +240,25 @@ protected void applyChanges(Requisition to, Requisition from) throws IOException
...
+    applyContactChanges(to, from);
+  }
+
+  private void applyContactChanges(Requisition to, Requisition from) {
+    to.getContacts().removeIf(
+        toItem -> from.getContacts().stream().noneMatch(fromItem -> contactsMatch(toItem, fromItem)));
+    from.getContacts().forEach(fromItem -> {
+      if (to.getContacts().stream().noneMatch(toItem -> contactsMatch(fromItem, toItem))) {
+        to.getContacts().add(fromItem);
+      }
```
avarsava, line 255: Correct me if I'm misreading this, but this looks like it's saying 'remove contacts from `to` if they're not in `from`'s contacts, add contacts to `to` if `from` has them and `to` doesn't'. Isn't that effectively `to.setContacts(new ArrayList<RequisitionContact>(from.getContacts())` ?

riyapatel36: Yes, logically it’s the same result. I kept it this way to avoid replacing the hibernate managed collection and to ensure newly added contacts have their requisition set.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRequisitionService.java
riyapatel36: Yes, thanks Alexis. I can change this to use a set of contact-role keys to avoid the nested loop.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateContactRoleDao.java
```diff
@@ -21,7 +22,8 @@ public HibernateContactRoleDao() {
 
   @Override
   public long getUsage(ContactRole contactRole) throws IOException {
-    return getUsageBy(ProjectContact.class, "contactRole", contactRole);
+    return getUsageBy(ProjectContact.class, "contactRole", contactRole)
+        + getUsageBy(RequisitionContact.class, "contactRole", contactRole);
```
djcooke: This needs to be separated into two separate methods because when a user attempts to delete a contact role, they should see an error message telling them why it can't be deleted - it is linked to a number of project contacts and/or a number of requisition contacts. That will need updated in `DefaultContactRoleService#validateDeletion`.

`DefaultContactService` should also have a `validateDeletion` that checks for usage by project contacts and requisition contacts. Can you add that too please?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRequisitionService.java
djcooke: If a `Set` is used to prevent duplication, then the duplicate will be ignored instead of being caught by validation. This could hide a mistake in data entry.

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateContactRoleDao.java
riyapatel36: Sure Dillan, I will add that. Thanks

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRequisitionService.java
riyapatel36: I’ll keep the validation behaviour but use a Set only to track seen contact-role keys so duplicates are still reported.

avarsava (approved):

sqlstore/src/main/java/uk/ac/bbsrc/tgac/miso/persistence/impl/HibernateContactDao.java
```diff
@@ -36,4 +39,14 @@ public Contact getByEmail(String email) throws IOException {
     return getBy(Contact_.EMAIL, email);
   }
 
+  @Override
+  public long getProjectUsage(Contact contact) throws IOException {
+      return getUsageBy(ProjectContact.class, "contact", contact);
```
djcooke: Use metamodel properties (e.g. from `ProjectContact_`) rather than Strings for field names.

Please also add tests for the new DAO methods.

djcooke (approved):

Issue #3232, opened 2026-04-27
Performance issue when uploading samples in Bulk (>100)


lucasperception:
For many of our projects we rely on bulk imports of samples 10 < n < 1000. While testing these, the following issue was raised:

**Issue**
When importing large sample sheets (n>100) miso-lims freezes temporarily. Starting at n>300 miso-lims freezes and crashes the browser tab.

**How to reproduce**
Import a sample sheet with more than 100 samples on a miso-lims instance with detailed sample mode enabled.

**Probable cause**
A quick look in the performance profiler shows that most recorded frames are related to rendering within handsontable.
<img width="1838" height="1269" alt="Image" src="https://github.com/user-attachments/assets/8014ceaf-1e75-4d41-9e57-f1137ee20ce5" />
Trying to identify what causes the significant amount of frames related to rendering leads to the on change handlers within the bulk target definition for samples. `bulk_sample.js` where at least `n * 4 calls` to `bulkjs.updateField()` are made triggering a render each. As a result too many handsontable renders are triggered, filling up the JS event loop.

**Potential Fixes**
- Handsontable starting with version 9 supports [batch operations](https://handsontable.com/docs/9.0/batch-operations/#overview), batching multiple renders into one. Unfortunately, miso‑lims uses version 6.2.2. Could an upgrade be viable?
- Reduce the number of updateField() calls.
- Bypass client‑side validations altogether by providing an API to upload large sample sheets.

Thanks a lot for your help, and looking forward to hearing your take on this.

PR #3235, opened 2026-05-07
#3232 Queued identityId renders to increase performance on bulk sample import

lucasperception:
- bulk.js: add skipRender flag to updateField defaulting to previous behavior
- bulk.js: expose renderFields to render changes in bulk
- bulk_sample.js: when resolving the identity lookup callback update the field without rendering
- bulk_sample.js: added a queue and timeout to render fields at once

Jira ticket or GitHub issue:

- [ ] Includes a change file

### DIFF
--- a/changes/fix_bulk_create_samples_performance.md
+++ b/changes/fix_bulk_create_samples_performance.md
@@ -1,0 +1,1 @@
+Improved performance when creating samples in bulk.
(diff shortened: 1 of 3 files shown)


djcooke (commented):
This works well and noticeably improves performance. I think more should be contained within `bulk.js` though. The parameter could be called `bulkRender` instead of `skipRender`, and when true, `updateField` could handle the queue and timer.

That way it's easily reusable, and not possible to make the mistake of setting `skipRender` and then never calling `renderFields`.

djcooke:
Looks good! Just need to add a couple pieces of documentation:

1. There's documentation for the API at the top of `bulk.js` where the new parameter in `updateField` should be noted.
2. A change file should be added to mention improved performance of import on the bulk Create Samples page. We usually classify performance improvements as "fix" type changes.

riyapatel36 (approved):

djcooke (approved):

PR #3238, opened 2026-05-14
runscanner auth

riyapatel36:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GP-5407

### DIFF
--- a/docs/admin/site-configuration.md
+++ b/docs/admin/site-configuration.md
@@ -309,6 +309,14 @@ a space to separate.
 
 Interval in milliseconds to poll Run Scanner(s). Defaults to `300000` (5 minutes) if unspecified;
 
+### `miso.runscanner.httpBasicAuthUsername`
+
+Username to use for HTTP Basic authentication when connecting to Run Scanner.
+
+### `miso.runscanner.httpBasicAuthPassword`
+
+Password to use for HTTP Basic authentication when connecting to Run Scanner.
+
 ### `miso.boxscanner.servers`
 
 Config for Visionmate or DP5Mirage bulk barcode scanners. Enter in format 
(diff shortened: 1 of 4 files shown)


miso-web/src/main/resources/miso.properties
```diff
@@ -63,7 +63,12 @@ miso.taxonLookup.enabled:false
 # A white-space delimited list of run scanner instances to poll
 #miso.runscanner.urls:http://your.runscann.er:8080/
 # The delay between queries to the server, in milliseconds
+
+#miso.runscanner.urls:http://localhost:8080/runscanner
```
djcooke: Remove - this property is already included above, and the comment above belongs with the property below

miso-web/src/main/resources/miso.properties
```diff
@@ -63,7 +63,12 @@ miso.taxonLookup.enabled:false
...
 miso.runscanner.interval:300000
+#miso.runscanner.httpbasicauth:some_username
+#miso.runscanner.basicpass:some_password
```
djcooke: Use `miso.runscanner.httpBasicAuthUsername` and `...httpBasicAuthPassword` to match the similar properties for JIRA integration.

These new properties should also be described in the Site Configuration section of the admin guide.

docs/admin/site-configuration.md
```diff
@@ -309,6 +309,14 @@ a space to separate.
 
 Interval in milliseconds to poll Run Scanner(s). Defaults to `300000` (5 minutes) if unspecified;
 
+### `miso.runscanner.httpBasicAuthUsername`
+
+Username to use for HTTP Basic authentication when connecting to Run Scanner.
```
djcooke: It should be noted that this is optional and only necessary when authentication is configured at the proxy level since Run Scanner doesn't have any built-in authentication

djcooke (commented):
Needs a change file

djcooke (approved):

callunity (approved):

PR #3240, opened 2026-05-21
allow miso to read ULTIMA runs from runscanner

grapigeau:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GP-5431

- [ x] Includes a change file


core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/UltimaRun.java
```diff
@@ -23,4 +23,12 @@ public String getDeleteType() {
...
   }
 
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj)
+      return true;
+    if (!super.equals(obj))
+      return false;
+    return true;
+  }
```
djcooke: This isn't doing anything the parent class wasn't already doing

djcooke (approved):

avarsava (approved):

PR #3251, opened 2026-06-23
#3250: add configurable required fields for pools

### LINKED ISSUE #3250: Add configuration to require Pool size field
We need Pool's Size (bp) field to be required at our site, but it's optional in core MISO. Currently there's no way to enforce this without modifying source code.

Following discussion in #3237, we landed on an enum-based approach.

**Design**

A `RequiredPoolField` enum defines the fields that can be configured as required. Each enum value owns its own validation logic. Config is a comma-separated list of enum names using UI field names:

miso.required.pool:SIZE

Defaults to empty (no-op for existing installations).

**Backend**

Each enum value implements `collectValidationErrors`(`Pool pool`, `List<ValidationError> errors`). The service iterates the configured list during save:

```
for (RequiredPoolField field : requiredPoolFields) {
  field.collectValidationErrors(pool, errors);
}
```

Spring binds the property to `List<RequiredPoolField>`, so invalid names fail at startup. Validation runs on every save path including indirect edits. If an admin requires a previously-optional field, they're responsible for backfilling existing records first.

**Frontend**

`ConstantsController` exposes the configured field names via `Constants.requiredPoolFields`. Both the bulk page and single-edit form read this and set `required: true` on matching fields. On bulk pages, fields hidden via `miso.display.bulk.*` are skipped.

**Extensibility**

The same pattern (new enum + config property) can be used for other object types if the need comes up.

Supersedes #3237.

mohammedkhalfan:
## Summary
  - Added `miso.required.pool` property to make specified pool fields required
  - Currently supports `SIZE` (insert size / DNA size)
  - Validation enforced on all save paths: single edit form, bulk tables, and indirect edits (transfers, boxes)
  - Invalid field names cause a startup error
  - Front-end marks configured fields as required on both the pool edit form and bulk pool table
  - [x] Includes a change file

  Closes #3250

### DIFF
--- a/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultPoolService.java
+++ b/miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultPoolService.java
@@ -8,7 +8,9 @@ import java.math.BigDecimal;
 import java.time.LocalDate;
 import java.time.ZoneId;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.LinkedList;
@@ -18,6 +20,8 @@ import java.util.TreeSet;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+
+import jakarta.annotation.PostConstruct;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
@@ -29,6 +33,7 @@ import com.eaglegenomics.simlims.core.Note;
 
 import uk.ac.bbsrc.tgac.miso.core.data.Box;
 import uk.ac.bbsrc.tgac.miso.core.data.Pool;
+import uk.ac.bbsrc.tgac.miso.core.data.type.RequiredPoolField;
 import uk.ac.bbsrc.tgac.miso.core.data.SequencingOrder;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.OrderLibraryAliquot;
 import uk.ac.bbsrc.tgac.miso.core.data.impl.PoolOrder;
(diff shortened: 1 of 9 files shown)


miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultPoolService.java
```diff
@@ -69,6 +74,11 @@ public class DefaultPoolService implements PoolService {
   @Value("${miso.autoGenerateIdentificationBarcodes}")
   private Boolean autoGenerateIdBarcodes;
 
+  @Value("${miso.required.pool:#{null}}")
+  private String requiredPoolFieldsConfig;
+
+  private List<RequiredPoolField> requiredPoolFields;
```
djcooke: You can put the `Value` annotation on the `List<RequiredPoolField>` and Spring will handle the conversion

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/controller/ConstantsController.java
```diff
@@ -250,6 +251,8 @@ public class ConstantsController {
   private boolean showDiscarded;
   @Value("${miso.defaults.bulk.detailedQcStatus:#{null}}")
   private String defaultDetailedQcStatus;
+  @Value("${miso.required.pool:#{null}}")
+  private String requiredPoolFieldsConfig;
```
djcooke: Simplify by declaring as `List<RequiredPoolField>` here too

miso-web/src/main/webapp/scripts/form_pool.js
```diff
@@ -100,7 +100,16 @@ FormTarget.pool = (function ($) {
...
+            (function () {
+              var field = FormUtils.makeDnaSizeField();
+              if (
+                Constants.requiredPoolFields &&
+                Constants.requiredPoolFields.indexOf("dnaSize") !== -1
+              ) {
+                field.required = true;
+              }
+              return field;
+            })(),
```
djcooke, line 112: Should probably be implemented in a more generic way in `form.js`, similar to `bulk.js`. It's ok like this for now, but should be refactored when/if more options are added.

callunity (approved):

riyapatel36 (approved):

PR #3252, opened 2026-06-30
Glt 4710 populate sop data from runscanner

riyapatel36:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GLT-4710

- [x] Includes a change file

### DIFF
--- a/changes/add_consumableValueRun.md
+++ b/changes/add_consumableValueRun.md
@@ -1,1 +1,1 @@
-Sop field values can be automatically filled to MISO run.
+Run SOP field values can be automatically filled based on data from Run Scanner
--- a/changes/add_consumableValueRun.md
+++ b/changes/add_consumableValueRun.md
@@ -1,0 +1,1 @@
+Sop field values can be automatically filled to MISO run.
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/RunService.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/RunService.java
@@ -3,6 +3,7 @@ package uk.ac.bbsrc.tgac.miso.core.service;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 
 import uk.ac.bbsrc.tgac.miso.core.data.Run;
 import uk.ac.bbsrc.tgac.miso.core.exception.MisoNamingException;
@@ -44,11 +45,13 @@ public interface RunService extends DeleterService<Run>, SaveService<Run>, Pagin
 
   /**
    * Save a scanned run to the database or update the run if it exists.
-   * 
+   *
    * @param run the update from notification server
+   * @param consumableLotNumbersByType lot numbers reported by the notification server, keyed by consumable type
    * @return true if the run is new, false if it already existed
    * @throws MisoNamingException
    */
-  boolean processNotification(Run run) throws IOException, MisoNamingException;
+  boolean processNotification(Run run, Map<String, String> consumableLotNumbersByType)
+      throws IOException, MisoNamingException;
 
 }
--- a/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/RunService.java
+++ b/core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/RunService.java
@@ -47,11 +47,12 @@ public interface RunService extends DeleterService<Run>, SaveService<Run>, Pagin
    * Save a scanned run to the database or update the run if it exists.
    *
    * @param run the update from notification server
-   * @param consumableLotNumbersByType lot numbers reported by the notification server, keyed by consumable type
+   * @param consumableDataByType lot numbers reported by the notification server, keyed by consumable
+   *        type
    * @return true if the run is new, false if it already existed
    * @throws MisoNamingException
    */
-  boolean processNotification(Run run, Map<String, String> consumableLotNumbersByType)
+  boolean processNotification(Run run, Map<String, String> consumableDataByType)
       throws IOException, MisoNamingException;
 
 }
--- a/miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
+++ b/miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
@@ -38,7 +38,6 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 
-import ca.on.oicr.gsi.runscanner.dto.Consumable;
 import ca.on.oicr.gsi.runscanner.dto.IlluminaNotificationDto;
 import ca.on.oicr.gsi.runscanner.dto.NotificationDto;
 import ca.on.oicr.gsi.runscanner.dto.OxfordNanoporeNotificationDto;
@@ -2143,8 +2142,8 @@ public class Dtos {
     setDateString(dto::setDataReviewDate, from.getDataReviewDate());
     setId(dto::setSopId, from.getSop());
     dto.setSopFieldValues(from.getSopFieldValues().stream()
-                    .filter(value -> value.getSopField() != null)
-                    .filter(value -> !isStringBlankOrNull(value.getValue()))
+        .filter(value -> value.getSopField() != null)
+        .filter(value -> !isStringBlankOrNull(value.getValue()))
         .collect(Collectors.toMap(
             value -> value.getSopField().getId(),
             RunSopFieldValue::getValue)));
@@ -3006,17 +3005,6 @@ public class Dtos {
       to.setWorkflowType(IlluminaWorkflowType.get(from.getWorkflowType()));
     }
     to.setDataManglingPolicy(getDataManglingPolicy(from.getIndexSequencing()));
-    setRunScannerConsumables(from, to);
-  }
-
-  private static void setRunScannerConsumables(@Nonnull IlluminaNotificationDto from, @Nonnull Run to) {
-    if (from.getConsumables() == null) {
-      return;
-    }
-    to.setRunScannerConsumables(from.getConsumables().stream()
-        .filter(consumable -> !isStringEmptyOrNull(consumable.getType()))
-        .filter(consumable -> !isStringEmptyOrNull(consumable.getLotNumber()))
-        .collect(Collectors.toMap(Consumable::getType, Consumable::getLotNumber, (existing, replacement) -> existing)));
   }
 
   private static InstrumentDataManglingPolicy getDataManglingPolicy(IndexSequencing indexSequencing) {
--- a/miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
+++ b/miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
@@ -2141,12 +2141,9 @@ public class Dtos {
     setString(dto::setDataReviewer, maybeGetProperty(from.getDataReviewer(), User::getFullName));
     setDateString(dto::setDataReviewDate, from.getDataReviewDate());
     setId(dto::setSopId, from.getSop());
-    dto.setSopFieldValues(from.getSopFieldValues().stream()
-        .filter(value -> value.getSopField() != null)
-        .filter(value -> !isStringBlankOrNull(value.getValue()))
-        .collect(Collectors.toMap(
-            value -> value.getSopField().getId(),
-            RunSopFieldValue::getValue)));
+    Map<Long, String> sopFieldValues = new HashMap<>();
+    from.getSopFieldValues().forEach(value -> sopFieldValues.put(value.getSopField().getId(), value.getValue()));
+    dto.setSopFieldValues(sopFieldValues);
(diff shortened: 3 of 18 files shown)


miso-web/src/main/webapp/scripts/form_instrumentmodel.js
```diff
@@ -97,6 +98,22 @@ FormTarget.instrumentmodel = (function ($) {
...
+        data: "defaultRunSopId",
+        type: "read-only",
+        include: object.instrumentType === "SEQUENCER",
+        getDisplayValue: function (model) {
+          if (!model.defaultRunSopId) {
+            return "n/a";
+          }
+          var sop = Utils.array.findUniqueOrThrow(
+            Utils.array.idPredicate(model.defaultRunSopId),
+            config.sops || []
```
callunity, line 112: What do you think about adding a miso.properties setting to toggle whether this field should be visible in the bulk tables? If a particular MISO install isn't using SOPs, then they probably don't need to see this column.

djcooke: The "SOP" field is already not shown on runs and other items if no applicable SOPs have been created. It would make sense to do the same here. No need for added config

changes/add_consumableValueRun.md
```diff
@@ -0,0 +1 @@
+Sop field values can be automatically filled to MISO run.
```
djcooke:
```suggestion
Run SOP field values can be automatically filled based on data from Run Scanner
```

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
```diff
@@ -2142,6 +2142,8 @@ private static RunDto asDto(@Nonnull Run from, boolean includeContainers, boolea
     setDateString(dto::setDataReviewDate, from.getDataReviewDate());
     setId(dto::setSopId, from.getSop());
     dto.setSopFieldValues(from.getSopFieldValues().stream()
+        .filter(value -> value.getSopField() != null)
+        .filter(value -> !isStringBlankOrNull(value.getValue()))
```
djcooke: Why would the field value's SOP field ever be null?
Why not include ones where the value is null?

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultInstrumentModelService.java
```diff
@@ -124,7 +130,17 @@ protected void collectValidationErrors(InstrumentModel object, InstrumentModel b
...
         errors.add(new ValidationError("dataManglingPolicy", "Should be 'normal' for non-sequencer models"));
       }
+      if (object.getDefaultRunSop() != null) {
+        errors.add(new ValidationError("defaultRunSopId", "Should be empty for non-sequencer models"));
+      }
     }
+    if (object.getDefaultRunSop() != null && object.getDefaultRunSop().getCategory() != SopCategory.RUN) {
+      errors.add(new ValidationError("defaultRunSopId", "Only run SOPs may be selected"));
+    }
+    if(object.getDefaultRunSop() != null && object.getDefaultRunSop().isArchived() && (beforeChange == null || beforeChange.getDefaultRunSop() == null || beforeChange.getDefaultRunSop().getId() != object.getDefaultRunSop().getId())) {
```
djcooke: Could group these both inside an `if (object.getDefaultRunSop() != null)` block instead of checking twice

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -879,6 +880,12 @@ public boolean processNotification(Run source) throws IOException, MisoNamingExc
     isMutated |= updateSequencingKitFromNotification(target, source.getSequencingKit());
     isMutated |= updateDataManglingPolicyFromNotification(target, source.getDataManglingPolicy(), user);
 
+
+    if (isNew && source.getSop() != null) {
+      target.setSop(source.getSop());
+    }
```
djcooke: This can be moved up into the `if (runFromDb == null)` block, and the null check on `source.getSop()` isn't necessary because it's fine to set the SOP to null.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -916,6 +923,55 @@ public boolean processNotification(Run source) throws IOException, MisoNamingExc
...
+        .collect(Collectors.toMap(
+            value -> value.getSopField().getId(),
+            value -> value,
+            (existing, replacement) -> existing));
+
+    boolean changed = false;
+    for (RunSopFieldValue sourceValue : sourceValues) {
+
+      if (sourceValue.getSopField() == null || isStringBlankOrNull(sourceValue.getValue())) {
+        continue;
```
djcooke: If either is true, it would be due to a bug in `RunScannerClient`, right? If that's the only reason, don't check these here.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/service/RunScannerClient.java
djcooke, line 1: You've reformatted this entire file to have 4-space tabs instead of 2

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/service/RunScannerClient.java
```diff
@@ -136,265 +136,355 @@ public class RunScannerClient {
...
+          for (NotificationDto dto : results) {
+              try (AutoCloseable timer = saveTime.start()) {
+                  Run notificationRun = Dtos.to(dto);
+                  setSequencer(notificationRun, dto.getSequencerName());
+                  setRunSequencingParameters(notificationRun, dto);
+                  setContainers(notificationRun, dto);
+                  Run existingRun = runService.getRunByAlias(notificationRun.getAlias());
+                  setDefaultRunSop(notificationRun, existingRun);
+                  setSopFieldValueFromConsumables(notificationRun, dto, existingRun);
+                  boolean isNew = runService.processNotification(notificationRun);
```
djcooke: You don't need to look up the existing run or find the appropriate SOP here. Keep all of that logic in `DefaultRunService`. Here you should just provide all the values that come from Run Scanner. Maybe that's as SopFieldValues, or maybe it's just a key/value map.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/service/RunScannerClient.java
```diff
@@ -136,265 +136,355 @@ public class RunScannerClient {
...
+      InstrumentModel model = notificationRun.getSequencer().getInstrumentModel();
+      if (model == null || model.getDefaultRunSop() == null || model.getDefaultRunSop().isArchived()) {
+          return null;
+      }
+      return model.getDefaultRunSop();
+  }
+
+  private void setSopFieldValueFromConsumables(Run notificationRun, NotificationDto dto, Run existingRun) {
+      if (!(dto instanceof IlluminaNotificationDto)) {
+          return;
```
djcooke: Ultima has consumables now too

miso-dto/src/main/java/uk/ac/bbsrc/tgac/miso/dto/Dtos.java
riyapatel36: Removed both filters. Thanks Dillan

miso-web/src/main/webapp/scripts/form_instrumentmodel.js
callunity: So, add a check for if `config.sops` includes any Run SOPs in the `include` field check?

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/service/RunService.java
```diff
@@ -44,11 +45,13 @@ public interface RunService extends DeleterService<Run>, SaveService<Run>, Pagin
...
    * Save a scanned run to the database or update the run if it exists.
-   * 
+   *
    * @param run the update from notification server
+   * @param consumableLotNumbersByType lot numbers reported by the notification server, keyed by consumable type
    * @return true if the run is new, false if it already existed
    * @throws MisoNamingException
    */
-  boolean processNotification(Run run) throws IOException, MisoNamingException;
+  boolean processNotification(Run run, Map<String, String> consumableLotNumbersByType)
```
djcooke: These are not necessarily all lot numbers. Maybe just `consumableDataByType`?

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/service/RunScannerClient.java
```diff
@@ -152,8 +156,8 @@ private void processResults(List<NotificationDto> results) {
           setSequencer(notificationRun, dto.getSequencerName());
           setRunSequencingParameters(notificationRun, dto);
           setContainers(notificationRun, dto);
-
-          boolean isNew = runService.processNotification(notificationRun);
+          Map<String, String> consumableLotNumbersByType = getConsumableLotNumbersByType(dto);
+          boolean isNew = runService.processNotification(notificationRun, consumableLotNumbersByType);
```
djcooke: Naming here too (not neccessarily lot numbers).

I'm seeing now that Run Scanner does call them lot numbers, but that is not always correct... 🫠

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -916,6 +922,71 @@ public boolean processNotification(Run source) throws IOException, MisoNamingExc
     return isNew;
   }
 
+  private boolean updateSopFieldValueFromConsumables(Run target, Map<String, String> consumableLotNumbersByType) {
```
djcooke:
```suggestion
  private boolean updateSopFieldValuesFromConsumables(Run target, Map<String, String> consumableLotNumbersByType) {
```

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -916,6 +922,71 @@ public boolean processNotification(Run source) throws IOException, MisoNamingExc
...
+
+    return changed;
+  }
+
+  private static String normalizeConsumableFieldName(String name) {
+    if (name == null) {
+      return "";
+    }
+    String normalized = name.toLowerCase(Locale.ROOT).replaceAll("[^a-z0-9]", "");
+    return normalized.replaceFirst("(lotnumber|lotno|lotnum|lot)$", "");
```
djcooke, line 978: Please add some info on this to the user manual under Type Data -> SOP Fields. Should mention that

1. If Run Scanner is being used, field values may be populated based on consumables reported by Run Scanner
2. For that to work, the SOP field name in MISO must match the consumable type from Run Scanner
3. One of these ("lot number", "lot no", etc.) may be added for clarity (case insensitive, spaces and special characters ignored). e.g. MISO SOP field "Reagent LOT#" will be considered a match for Run Scanner consumable type "Reagent"

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -916,6 +922,71 @@ public boolean processNotification(Run source) throws IOException, MisoNamingExc
...
+          existingValue.setRun(target);
+          existingValue.setSopField(sopField);
+          target.getSopFieldValues().add(existingValue);
+          existingValueByFieldId.put(sopField.getId(), existingValue);
+        }
+
+        if (!Objects.equals(existingValue.getValue(), consumable.getValue())) {
+          existingValue.setValue(consumable.getValue());
+          changed = true;
+        }
```
djcooke: Only set the value if it is currently blank. We never want to overwrite an existing value.

miso-web/src/main/java/uk/ac/bbsrc/tgac/miso/webapp/service/RunScannerClient.java
riyapatel36: Thanks Dillan, will update that

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -916,6 +922,69 @@ public boolean processNotification(Run source) throws IOException, MisoNamingExc
...
+        continue;
+      }
+
+      for (SopField sopField : matchingFields) {
+        if (!sopField.isValidValue(consumable.getValue())) {
+          continue;
+        }
+
+        RunSopFieldValue existingValue = existingValueByFieldId.get(sopField.getId());
+        if (existingValue != null && !isStringBlankOrNull(existingValue.getValue())) {
```
djcooke, line 953: A comment of "never overwrite existing values" here would be nice to clarify

djcooke (approved):

callunity (approved):

PR #3253, opened 2026-07-02
Added: Pinery-MISO mapping in related software

riyapatel36:
Jira ticket or GitHub issue : https://jira.oicr.on.ca/browse/GLT-4812

- [ ] Includes a change file


docs/user_manual/related_software.md
```diff
@@ -20,3 +20,46 @@ Documentation for Pinery can be found in its [GitHub repository](https://github.
...
 [MISO's GitHub repository](https://github.com/miso-lims/miso-lims/tree/develop/pinery-miso).
 
+### Pinery - MISO mapping
+
+|   | Pinery                                                                                                                                     | Pinery e.g.                                                                                                                                            | MISO DB                                                                                                               | MISO UI location                                                                                         |
+|---|:-------------------------------------------------------------------------------------------------------------------------------------------|:-------------------------------------------------------------------------------------------------------------------------------------------------------|:----------------------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------------------------------------|
```
avarsava: Not sure if you're following a style from other miso documentation, but imo markdown tables are easier to read with less whitespace. Won't block on this though

docs/user_manual/related_software.md
```diff
@@ -20,3 +20,46 @@ Documentation for Pinery can be found in its [GitHub repository](https://github.
 the MISO implementation of Pinery can be found in
 [MISO's GitHub repository](https://github.com/miso-lims/miso-lims/tree/develop/pinery-miso).
 
+### Pinery - MISO mapping
+
+|   | Pinery                                                                                                                                     | Pinery e.g.                                                                                                                                            | MISO DB                                                                                                               | MISO UI location                                                                                         |
```
avarsava: 'Example value' might be clearer

docs/user_manual/related_software.md
```diff
@@ -20,3 +20,46 @@ Documentation for Pinery can be found in its [GitHub repository](https://github.
...
+| 5  | sampleName                                                                                                                                 | "GLCS_0002_Ly_R_nn_1-3_LB11-01"                                                                                                                        | Sample.alias / Library.alias /LibraryAliquot.name                                                                     | Samples/Libraries/Library Aliquots > Edit page > Alias or Name                                           |
+| 6  | sampleAttributes. barcode_kit                                                                                                              | [ "ILMN UDI Set A v3" ]                                                                                                                                | LibraryIndexFamily.name                                                                                               | Libraries > Edit Library > index/index family                                                            |
```
avarsava: Just confirming that some of these should be in `[]`?

riyapatel36: Yes

docs/user_manual/related_software.md
riyapatel36: Thanks Alexis, will update that.

avarsava (approved):

docs/user_manual/related_software.md
riyapatel36: Thanks Alexis, I didn't see any other tables in markdown

docs/user_manual/related_software.md
```diff
@@ -20,3 +20,46 @@ Documentation for Pinery can be found in its [GitHub repository](https://github.
...
+| 32 | iusTag                                                                                                                                     | "CCACCAGGCA-ATTCCATAAG"                                                                                                                                | LibraryIndex.sequence                                                                                                 | Libraries > Edit Library > Index 1 / Index 2                                                             |
+| 33 | skip                                                                                                                                       | false                                                                                                                                                  | PartitionQCType.analysisSkipped                                                                                       | Edit Run > Partitions/Lanes table > partition QC / analysis skipped status                               |
```
callunity: What do you think about setting the fourth column to: `Edit Run > Partitions/Lanes table > analysis skipped value associated with the partition/lane's QC Status`

docs/user_manual/related_software.md
```diff
@@ -20,3 +20,46 @@ Documentation for Pinery can be found in its [GitHub repository](https://github.
...
+| 34 | sampleProvenanceId                                                                                                                         | "9200_7_LDI148800"                                                                                                                                     | Run.runId + _Partition.partitionNumber + LibraryAliquot.name                                                          | Not directly set in UI                                                                                   |
```
callunity: I don't think it's quite accurate to say it's `Not directly set in UI` as UI changes like changing a library in a pool or changing the lane that a pool is associated with will change the sampleProvenanceId . What do you think about something like:
`Can be modified by changing a library's presence in a pool, and/or a pool's association with a sequencer run partition/lane`

docs/user_manual/related_software.md
```diff
@@ -20,3 +20,46 @@ Documentation for Pinery can be found in its [GitHub repository](https://github.
...
+| 35 | version                                                                                                                                    | "b742c895d65af242b50036de930a7499c33682fcefb2fa55917adc6792899ae6"                                                                                     | SHA-256 checksum of the object's fields and values, useful for detecting object modifications                         | Not set in UI                                                                                            |
+| 36 | lastModified                                                                                                                               | "2026-05-19T14:16:00Z"                                                                                                                                 | Sample.lastModified / Library.lastModified /  LibraryAliquot.lastUpdated / Run.lastModified                           | Edit Sample/Library/Library Aliquot/Run metadata, System managed                                         |
```
callunity: `System managed` -> `system-managed` (and below)

docs/user_manual/related_software.md
```diff
@@ -20,3 +20,46 @@ Documentation for Pinery can be found in its [GitHub repository](https://github.
...
+| 37 | createdDate                                                                                                                                | "2026-05-10T00:00:00Z"                                                                                                                                 | Sample.created / Library.created / LibraryAliquot.created / Run.created                                               | Edit Sample/Library/Library Aliquot/Run metadata, System managed                                         |
+| 38 | Batch ID attribute:for libraries built from Library.creationDate, Library.creator, Library.sopId, Library.kitDescriptorId, Library.kitLot  | [ "2026-04-30_u160_s29_k280-21017065 , 21037394" ]                                                                                                     | batchIds                                                                                                              | Libraries > Edit Library > Batch ID                                                                      |
```
callunity: Second field should be `batchIds`
For the fourth field, I think this should be where the composition description goes. What do you think about something like:
```
A composite key built from several fields. 
For libraries: Library.creationDate, Library.creator, Library.sopId, Library.kitDescriptorId, Library.kitLot
```

docs/user_manual/related_software.md
riyapatel36: But can we see `sampleProvenanceId` in MISO UI ? I wrote not directly set in UI because it's a combination of things but not directly set in UI.

docs/user_manual/related_software.md
riyapatel36: I will update that Thanks Heather.

docs/user_manual/related_software.md
riyapatel36: I don't understand what do you mean by second field to set `batchIds`. 
4th column is for the location in MISO UI. I think it's not appropriate for MISO UI. I can update that to `Libraries > Edit Library > Batch > Batch ID  `

callunity: Ah, I wonder if you're missing the first column in your count? First column is the row number, second column is Pinery value (`batchIds`), etc.

riyapatel36: oh got it.. Thanks Heather

docs/user_manual/related_software.md
callunity: Maybe this could be clarified by changing the value to `Not displayed in MISO UI` ? I still think some additional explanatory text would help though, as it's an essential identifier and right now it's just described by database columns. What do you think about this?
```
Not displayed in MISO; this value is derived from the run ID, lane/partition number, and library aliquot name
```

riyapatel36: I will update to this. Thanks Heather

docs/user_manual/related_software.md
```diff
@@ -20,3 +20,46 @@ Documentation for Pinery can be found in its [GitHub repository](https://github.
...
+| 37 | createdDate                                       | "2026-05-10T00:00:00Z"                                                                                                                                 | Sample.created / Library.created / LibraryAliquot.created / Run.created                                                                                | Edit Sample/Library/Library Aliquot/Run metadata, system-managed                                                  |
+| 38 | batchIds                                          | [ "2026-04-30_u160_s29_k280-21017065 , 21037394" ]                                                                                                     | Batch ID attribute:for libraries built from Library.creationDate, Library.creator, Library.sopId, Library.kitDescriptorId, Library.kitLot              | Libraries > Edit Library > Batch > Batch ID                                                                       |
```
callunity: Fourth column suggestion:
```
For libraries: built from Library.creationDate, Library.creator, Library.sopId, Library.kitDescriptorId, Library.kitLot
```
Additionally, I see your example (column 3) has a comma-separated value. I actually went looking at the data to confirm whether it should be two strings in the list, because it looks kinda incorrect (which is an issue with our internal data, not with your work) and it does invite confusion for new users. What do you think about truncating the string to `"2026-04-30_u160_s29_k280-21017065"` to reduce confusion?

riyapatel36: I think that makes sense as current change might create confusion to our users.Thanks Heather Updating this now

callunity (approved):

PR #3259, opened 2026-07-20
GP-5747 workstation sop field type

riyapatel36:
Jira ticket or GitHub issue: https://jira.oicr.on.ca/browse/GP-5747

- [x] Includes a change file


callunity (approved):

changes/add_workstation_sopField.md
```diff
@@ -0,0 +1 @@
+New sopField type workstation has been added.
```
djcooke:
```suggestion
New SOP field type for recording workstations used
```

djcooke: The user manual (Type Data -> Standard Operating Procedures) should also be updated since it currently lists text and number as the only types

core/src/main/java/uk/ac/bbsrc/tgac/miso/core/data/SopField.java
```diff
@@ -106,6 +106,7 @@ public boolean isValidValue(String value) {
         return BigDecimalValidator.getInstance().validate(value) != null;
 
       case TEXT:
+      case WORKSTATION:
       default:
         return true;
     }
```
djcooke: This isn't a great place for the validation since there's no database access. You could ensure that it's an integer...

It would be better to move the validation into the service layer though. Then you can ensure that the workstation actually exists too.

Also in the service layer, the `validateDeletion` method in `DefaultWorkstationService` has to check for SOP fields using the workstation. This is the only way to protect against deleting the values since we can't have a foreign key constraint in the database.

changes/add_workstation_sopField.md
riyapatel36: can I add Instrument also in user manual  ? As we are going to merge both of this PR together probably

djcooke: Yeah, that's fine

docs/user_manual/type_data.md
```diff
@@ -287,8 +287,11 @@ name and the new version. This way, you continue to track which version of an SO
 ### SOP Fields
 
 SOP fields are used to capture additional SOP-specific information, such as reagent LOT numbers.
-Each field has a name, a data type (text or number), and optional units. When the SOP is added to an
-item, that item gains all of the fields defined by the SOP.
+Each field has a name, a data type (text, number, workstation, or instrument), and optional units.
+A workstation field is entered by selecting one of the workstations defined in MISO, rather than
+typing a value. An instrument field also specifies an instrument model when it is created; the
+field is entered by selecting one of the instruments of that model defined in MISO. When the SOP
+is added to an item, that item gains all of the fields defined by the SOP.
```
djcooke: > Each field has a name, a data type (text, number, workstation, or instrument), and optional units. When the SOP is added to an item, that item gains all of the fields defined by the SOP. A workstation field is entered by selecting one of the workstations defined in MISO, rather than typing a value. An instrument field also specifies an instrument model when it is created; the field is entered by selecting one of the instruments of that model defined in MISO.

Easier to follow in this order (just move the last sentence to earlier)

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -492,14 +496,21 @@ private void validateSopFieldValues(Run run, List<ValidationError> errors) {
         errors.add(new ValidationError(property, "Maximum length: 255 characters"));
       }
 
-      if (!sopField.isValidValue(value.getValue())) {
+      boolean valid = sopField.isValidValue(value.getValue())
+          && (sopField.getFieldType() != SopField.FieldType.WORKSTATION || isValidWorkstation(value.getValue()));
```
djcooke: Better to move all of the validation into the service and get rid of `SopField#isValidValue` since that method already can't really do its job and will get worse with instrument and other potential types.

miso-service/src/main/java/uk/ac/bbsrc/tgac/miso/service/impl/DefaultRunService.java
```diff
@@ -496,16 +497,35 @@ private void validateSopFieldValues(Run run, List<ValidationError> errors) throw
...
+    switch (sopField.getFieldType()) {
+      case NUMBER:
+        return BigDecimalValidator.getInstance().validate(value) != null;
+
+      case WORKSTATION:
+        return isValidWorkstation(value);
+
+      case TEXT:
+      default:
+        return true;
```
djcooke, line 522: Each possible type should be implemented here, so hitting the default case is a programming error and should throw an exception indicating the unexpected/unhandled value

djcooke (approved):

